PR #17, opened 2019-08-22
Kubernetes improvements

redondos:
- use EBS volumes
- specify resource requests/limits


redondos:
@skerschb does this go inline with your plans for this application? I believe the EBS volume part is important since it will improve this app's performance.

skerschb:
@redondos yes I think so, we are just waiting for a moment to test it.

PR #24, opened 2019-10-03
init for getting repos

danielborowski:


worker/utils/mongo.js
```diff
@@ -28,6 +31,41 @@ module.exports = {
     return null;
   },
 
+  getMetaCollection() {
+    if (client) {
+      // hardcoded for now because if I use env variables
+      // I'll need to create a meta collection in both databases: `pool` and `pool_test`
+      return client.db('pool').collection('meta');
```
skerschb: Go ahead and create the collection in the staging database.

worker/utils/mongo.js
```diff
@@ -28,6 +31,41 @@ module.exports = {
...
+  },
+
+  async getRepoPublishedBranches(repoObject) {
+    const pubBranchesFile = `https://raw.githubusercontent.com/${repoObject.repoOwner}/${repoObject.repoName}/meta/published-branches.yaml`;
+    const returnObject = {};
+    return new Promise(function (resolve, reject) {
+      request(pubBranchesFile, function(error, response, body) {
+        if (!error && response.statusCode == 200) {
+          try {
+            const yamlParsed = yaml.safeLoad(body);
```
skerschb: Can we put a null check on the body?

worker/utils/mongo.js
```diff
@@ -28,6 +31,41 @@ module.exports = {
...
+            returnObject['status'] = 'success';
```
skerschb: What is success in this case? Validation may be required.

worker/utils/mongo.js
```diff
@@ -1,4 +1,6 @@
 const { MongoClient } = require('mongodb');
+const request = require('request');
```
skerschb: mongo.js is a utility class for accessing the mongoDB database -- what architecture concept might we use to insulate the caller (in this case githubpushJob.js) from the implementation details? Hint there is another class in this directory that would be a nice home for calls to encapsulated database calls.

worker/utils/mongo.js
danielborowski: done

PR #26, opened 2019-10-11
DOCSP-7001: init

danielborowski:


skerschb:
Per our conversation about cleaning up the design a bit, can we split this into three components? Maybe:

1.  The githubpushJob.js which serves as a sort of "shell" for executing the build from within an instantiated GitHubJob.js class -- and handles of all of the validation, sanitizing and logging?
2. The GitHubJob.js class which would have all of the github-specific build code (cloning, patching, building, and for production builds -- encrypting and decrypting messages, creating other build messages for the world and repo builders)
3. An S3Publish class which has the publish functionality for both production and staging S3

Let me know what you think!

worker/jobTypes/githubJob.js
```diff
@@ -0,0 +1,114 @@
...
+      const command = `
+        cd repos/${workerUtils.getRepoDirName(currentJob)}; 
+        git checkout ${currentJob.payload.branchName}; 
+        git pull origin ${currentJob.payload.branchName};
+      `;
+
+      await exec(command);
+   
+      const commandbuild = `
+        . /venv/bin/activate;
```
skerschb: Should we use && here to short circuit after a fail on one of the commands?

worker/jobTypes/githubJob.js
```diff
@@ -0,0 +1,114 @@
...
+        throw errResult;
+      }
+    }
+    return new Promise(function(resolve, reject) {
+      resolve(true);
+    });
+  }
+
+  async buildRepo(logger) {
+    const currentJob = this.currentJob;
```
skerschb: one further abstraction would be to have buildRepo call cleanup and clone and pull that out of the githubPushJob.js to make it even cleaner.

worker/jobTypes/githubJob.js
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      // master branch cannot run through staging build
+      if (currentJob.payload.branchName === 'master') {
+        logger.save(`${'(BUILD)'.padEnd(15)} failed, master branch not supported`);
+        throw new Error('master branches not supported');
+      }
+
+      const exec = workerUtils.getExecPromise();
```
skerschb: pull the getBasePath function into githubJob since that path is very specific to this job.

worker/jobTypes/githubJob.js
```diff
@@ -0,0 +1,114 @@
...
+        logger.save(`${'(BUILD)'.padEnd(15)}failed with code: ${errResult.code}`);
+        logger.save(`${'(BUILD)'.padEnd(15)}stdErr: ${errResult.stderr}`);
+        throw errResult;
+      }
+    }
+  }
+
+}
+
+module.exports = {
```
skerschb: It would be awesome if we could limit the visibility of the functions -- only expose build. Is it possible?

danielborowski: No :(

worker/jobTypes/S3Publish.js
```diff
@@ -0,0 +1,52 @@
+const fs = require('fs-extra');
+const workerUtils = require('../utils/utils');
+
+class S3PublishClass {
+
```
skerschb: Do we want two objects in the same core that have exactly the same definition? S3 might work more statically -- we pass the GitHubJob into the function.

worker/jobTypes/S3Publish.js
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  async pushToStage(logger) {
+    const currentJob = this.currentJob;
+    logger.save(`${'(stage)'.padEnd(15)}Pushing to staging`);
+    try {
+      const exec = workerUtils.getExecPromise();
+      const command = `
+        . /venv/bin/activate && 
+        cd repos/${workerUtils.getRepoDirName(currentJob)} &&
```
skerschb: This getRepoDirName function should probably move into the GitHubJob class, then reference it here. I don't like that repos is hardcoded all over the place. So I'm wondering if we should put that in the GitHubJob class as well and only reference it there.

worker/jobTypes/S3Publish.js
```diff
@@ -0,0 +1,52 @@
...
+        cd repos/${workerUtils.getRepoDirName(currentJob)} && 
+        make stage;
+      `;
+      const { stdout, stderr } = await exec(command);
+      let stdoutMod = '';
+      // get only last part of message which includes # of files changes + s3 link
+      if (stdout.indexOf('Summary') !== -1) {
+        stdoutMod = stdout.substr(stdout.indexOf('Summary'));
+      } 
+      console.log(stdoutMod);
```
skerschb: console logs in our autobuilders don't give us much.

worker/jobTypes/S3Publish.js
```diff
@@ -0,0 +1,52 @@
+const fs = require('fs-extra');
+const workerUtils = require('../utils/utils');
+
+class S3PublishClass {
+
+  constructor(currentJob) {
+    this.currentJob = currentJob;
+  }
+
```
skerschb: I would put a placeholder in for the productionPush even though we aren't ready to code it yet.

worker/jobTypes/githubJob.js
```diff
@@ -0,0 +1,134 @@
...
+        cd repos/${workerUtils.getRepoDirName(currentJob)}; 
+        git checkout ${currentJob.payload.branchName}; 
+        git pull origin ${currentJob.payload.branchName};
+      `;
+
+      await exec(command);
+   
+      const commandbuild = `
+        . /venv/bin/activate && 
+        cd repos/${workerUtils.getRepoDirName(currentJob)} &&
```
skerschb: We talked about changing this to only check what type of build and then run the targets directly.

PR #28, opened 2019-10-21
Encryption/Decryption Utility Functions

madelinezec:
https://jira.mongodb.org/browse/DOCSP-7344

### DIFF
--- a/worker/tests/unit/utils.test.js
+++ b/worker/tests/unit/utils.test.js
@@ -61,17 +61,19 @@ describe('Mongo Tests', () => {
   it('encryptJob()', () => {
     workerUtils.retrievePassword = jest.fn().mockReturnValue("password");
     workerUtils.generateSalt = jest.fn().mockReturnValue("A1+Czhen/oTh7k4UmbdOVQ==");
-    const digest = workerUtils.encryptJob("test string1", "test string2");
-    digest.then(function(value) {
-      expect(value).toBe("df22f6a6c4e704a1fadd4af251d16ba78a6b04adbb75126f758416185da2310bab6915a4b1bb524c67c4943e0462617778aff7a5ed03e7b039768857437bfab6");
-    });
+    const promise1 = workerUtils.encryptJob("test string1", "test string2");
+    const promise2 = workerUtils.encryptJob("test string1", "test string2");
+   
+    Promise.all([promise1, promise2]).then(function(values) {
+      expect(values[0] === values[1]).toBeTruthy();
+    })
+
   });
 
-  //test decrypt job
-  it('decryptJob()', () => {
-    workerUtils.retrievePassword = jest.fn().mockReturnValue("password");
-    workerUtils.generateSalt = jest.fn().mockReturnValue("A1+Czhen/oTh7k4UmbdOVQ==");
-    const success = workerUtils.decryptJob("df22f6a6c4e704a1fadd4af251d16ba78a6b04adbb75126f758416185da2310bab6915a4b1bb524c67c4943e0462617778aff7a5ed03e7b039768857437bfab6", "test string1", "test string2");
+  //test validate job
+  it('validateJob()', () => {
+    const digest = Buffer.from("df22f6a6c4e704a1fadd4af251d16ba78a6b04adbb75126f758416185da2310bab6915a4b1bb524c67c4943e0462617778aff7a5ed03e7b039768857437bfab6", 'utf8');
+    const success = workerUtils.validateJob(digest, "test string1", "test string2");
     success.then(function(value) {
      expect(success).toBeTruthy();
     });
--- a/worker/tests/unit/utils.test.js
+++ b/worker/tests/unit/utils.test.js
@@ -57,7 +57,7 @@ describe('Mongo Tests', () => {
     expect(setTimeout).toHaveBeenCalledTimes(1);
   });
 
-
+  //test encrypt job
   it('encryptJob()', () => {
     workerUtils.retrievePassword = jest.fn().mockReturnValue("password");
     workerUtils.generateSalt = jest.fn().mockReturnValue("A1+Czhen/oTh7k4UmbdOVQ==");
--- a/worker/tests/unit/utils.test.js
+++ b/worker/tests/unit/utils.test.js
@@ -61,13 +61,13 @@ describe('Mongo Tests', () => {
   it('encryptJob()', () => {
     const digest = workerUtils.encryptJob("test string1", "test string2");
     digest.then(function(value) {
-      expect(value).toBe("55397fd48c066c9747ce6b08bf75a6d8be92310a576d6e4b7dcf283e03a3bc6c");
+      expect(value).toBe("232d1df9b58c1d7d6c611a9b10eb752687927931bfeb6aea3eae430e056bde7e");
     });
   });
 
   //test decrypt job
   it('decryptJob()', () => {
-    const success = workerUtils.decryptJob("55397fd48c066c9747ce6b08bf75a6d8be92310a576d6e4b7dcf283e03a3bc6c", "test string1", "test string2");
+    const success = workerUtils.decryptJob("232d1df9b58c1d7d6c611a9b10eb752687927931bfeb6aea3eae430e056bde7e", "test string1", "test string2");
     success.then(function(value) {
      expect(success).toBeTruthy();
     });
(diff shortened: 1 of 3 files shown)


worker/utils/utils.js
```diff
@@ -7,6 +7,15 @@ const yaml = require('js-yaml');
 const { promisify } = require('util');
 const exec = promisify(require('child_process').exec);
 const mongo = require('./mongo');
+const util= require('util');
+var crypto = require('crypto');
+
+try {
+  crypto = require('crypto');
+} catch (err) {
+  console.log(err)
```
skerschb: Would it be better to allow the tests to fail and prevent deployment that way?

worker/utils/utils.js
```diff
@@ -49,6 +58,23 @@ module.exports = {
     })
   },
 
+  async encryptJob(string1, string2){
+    const message = string1 + string2;
+    const hmac = crypto.createHmac('sha256', message);
```
skerschb: We need to use a password to create the Hmac. Then update the Hmac with the message. Example: https://nodejs.org/api/crypto.html#crypto_crypto_createhmac_algorithm_key_options.

The best way to manage the password is to use an environment variable which we will later set in kanopy using helm. For now you can assume that the tests will run with access to an environment variable named 'crypto_secret' or whatever you'd like to call it. An example of using environment variables in the worker pool: https://github.com/mongodb/docs-worker-pool/blob/master/worker/jobTypes/githubPushJob.js#L21

worker/utils/utils.js
madelinezec: I don't believe we need this test because crypto is included in the build of nodejs we're using. I forgot to delete this code in my original pull request

madelinezec:
incomplete - will update when we decide on logic for handling when password is not set

madelinezec:
Ready for a second pass!

worker/utils/utils.js
```diff
@@ -7,6 +7,9 @@ const yaml = require('js-yaml');
 const { promisify } = require('util');
 const exec = promisify(require('child_process').exec);
 const mongo = require('./mongo');
+const DB_NAME = process.env.DB_NAME ? process.env.DB_NAME : 'pool';
+var crypto = require('crypto');
```
skerschb: const DB_NAME is created here but not referenced, you can remove it. Instead of var, it's fine to use const for the crypto import. Using elint can catch some of these for you. https://eslint.org/docs/user-guide/getting-started. I use the air B&B style guide, and just have it locally installed for now (run the commands in the tutorial from worker dir).

worker/utils/utils.js
```diff
@@ -49,6 +52,29 @@ module.exports = {
...
+    digest = hmac.digest('hex')
+    return digest;
+  },
+
+  async decryptJob(digest, string1, string2){
+    const password = this.retrievePassword();
+    const message = string1 + string2;
+    const hmac = crypto.createHmac('sha256',password);
+    const hash = hmac.digest('hex')
+    if(hash == digest){
```
skerschb: you could just do return (hash == digest) which will give you a boolean.

madelinezec:
Pass back to you!

madelinezec:
Replaced SHA256 algo with scrypt algo per @i80and suggestion

worker/utils/utils.js
```diff
@@ -49,6 +51,27 @@ module.exports = {
     })
   },
 
+  async encryptJob(string1, string2){
```
i80and: `encrypt` is inaccurate nomenclature: encryption is reversible with the key, but the purpose of password hashing is that this procesws should be irreversible.

I would go with `pwhash` or something like that.

worker/utils/utils.js
```diff
@@ -49,6 +51,27 @@ module.exports = {
...
+    const secret = this.retrievePassword() + string1 + string2
+    const salt = this.generateSalt();
+    const digest = crypto.scryptSync(secret, salt, 64);
+    return digest.toString('hex');
+  },
+
+  async decryptJob(digest, string1, string2){
```
i80and: You're not decrypting: you're checking a password. Additionally, this can be defined in terms of `pwhashJob`, right?

worker/utils/utils.js
```diff
@@ -49,6 +51,27 @@ module.exports = {
...
+
+    const secret = this.retrievePassword() + string1 + string2;
+    const salt = this.generateSalt();
+    const hash = crypto.scryptSync(secret, salt, 64,);
+    return (hash == digest);
```
i80and: This probably doesn't matter in this specific case in *practical* terms, but as a general rule, any kind of comparison in cryptographic code should be constant-time using `crypto.timingSafeEqual()`.

(c.f. the OpenBSD implementation of password checking: https://github.com/openbsd/src/blob/b66614995ab119f75167daaa7755b34001836821/lib/libc/crypt/bcrypt.c#L229 )

worker/utils/utils.js
```diff
@@ -49,6 +51,27 @@ module.exports = {
...
+    const salt = this.generateSalt();
+    const digest = crypto.scryptSync(secret, salt, 64);
+    return digest.toString('hex');
+  },
+
+  async decryptJob(digest, string1, string2){
+
+    const secret = this.retrievePassword() + string1 + string2;
+    const salt = this.generateSalt();
+    const hash = crypto.scryptSync(secret, salt, 64,);
```
i80and: Extra comma at the end?

worker/utils/utils.js
i80and: The digest is converted to a hex string, but hash isn't?

worker/utils/utils.js
i80and: What are string1 and string2, and why is it important that they be concatenated with the password in this function?

worker/tests/unit/utils.test.js
```diff
@@ -57,6 +57,26 @@ describe('Mongo Tests', () => {
...
+    const digest = workerUtils.encryptJob("test string1", "test string2");
+    digest.then(function(value) {
+      expect(value).toBe("df22f6a6c4e704a1fadd4af251d16ba78a6b04adbb75126f758416185da2310bab6915a4b1bb524c67c4943e0462617778aff7a5ed03e7b039768857437bfab6");
+    });
+  });
+
+  //test decrypt job
+  it('decryptJob()', () => {
+    workerUtils.retrievePassword = jest.fn().mockReturnValue("password");
+    workerUtils.generateSalt = jest.fn().mockReturnValue("A1+Czhen/oTh7k4UmbdOVQ==");
```
skerschb: same here

worker/utils/utils.js
```diff
@@ -49,6 +51,27 @@ module.exports = {
...
 
+  async encryptJob(string1, string2){
+    const secret = this.retrievePassword() + string1 + string2
+    const salt = this.generateSalt();
+    const digest = crypto.scryptSync(secret, salt, 64);
+    return digest.toString('hex');
+  },
+
+  async decryptJob(digest, string1, string2){
+
```
skerschb: Since we aren't decrypting, you could call this "validateJob", and just call encryptJob within. Something like 
crypto.timingSafeEqual(digest, encryptJob(string1, string2);

More about timing attack protection -- there are other vulnerabilities to this approach that I'd be more worried about but Andrew reminded me it's good practice. https://nodejs.org/api/crypto.html#crypto_crypto_timingsafeequal_a_b

madelinezec:
Ready for third pass!

worker/tests/unit/utils.test.js
```diff
@@ -57,6 +57,26 @@ describe('Mongo Tests', () => {
     expect(setTimeout).toHaveBeenCalledTimes(1);
   });
 
+  //test encrypt job
+  it('encryptJob()', () => {
+    workerUtils.retrievePassword = jest.fn().mockReturnValue("password");
+    workerUtils.generateSalt = jest.fn().mockReturnValue("A1+Czhen/oTh7k4UmbdOVQ==");
```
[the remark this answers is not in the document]
skerschb: If we need to mock the generateSalt method, the encryption method won't work as-is. The reason is that in the wild we will not have the generated salt unless we add it to the job document along with the encrypted key...we can do that it's just not exposed in your implementation.

worker/utils/utils.js
```diff
@@ -49,6 +51,27 @@ module.exports = {
     })
   },
 
+  async encryptJob(string1, string2){
+    const secret = this.retrievePassword() + string1 + string2
```
skerschb: If you parameterize salt the caller can generate it and then store it and use it to re-encrypt for the validation.

PR #41, opened 2019-11-21
DOCSP-7636

varsha804:
I've added the Fastly API token as an environment variable (details in `README.md`)!


worker/jobTypes/fastlyJob.js
```diff
@@ -0,0 +1,54 @@
...
+class FastlyJobClass {
+    // pass in a job payload to setup class
+    constructor(currentJob) {
+        this.currentJob = currentJob;
+    }
+
+    // connects to dochub database and upsert {source: target} mappings
+    // to the fastly edge dictionary
+    connectAndUpsert(MongoClient, fastly) {
+        MongoClient.connect(mongo.url, function(err, client) {
```
skerschb: We could put the MongoClient connect stuff in the mongo utils class directly. In the event we decide to change the backend db...as odd as that would be...we could just swap out the implementation in the utility.

varsha804: Would it be better to instead add database and collection parameters to `connectAndUpsert()`? `mongo.js` seems to make use of environment variables in most its functions. I suppose we could also add environment variables for the `dochub` database and its `keys` collection, but that may make the `docker run` call too bulky.

skerschb: Sorry I missed this! If you have time on Monday...I'd rather pull all of the mongodb dependencies out of the Fastly component. The reason is that since we are following an adapter design strategy, having two integration points bound to each in one class breaks the abstraction. It's easier to maintain distributed applications if their bindings are limited in scope. Gives us better plugability.

worker/jobTypes/fastlyJob.js
```diff
@@ -0,0 +1,55 @@
...
+            assert.equal(null, err);
+        
+            const db = client.db(database);
+        
+            var cursor = db.collection(collection).find({});
+        
+            function iterateFunc(doc) {
+              const page = "https://dochub.mongodb.org/core/" + doc.name;
+              var request = require('request');
```
skerschb: We want to check to make sure that the target page is not 404ing. Before the dochub redirect is set up it likely will 404!

worker/tests/unit/publishDochubJob.unit.test.js
```diff
@@ -0,0 +1,127 @@
+const job = require('../../jobTypes/publishDochubJob');
+const workerUtils = require('../../utils/utils');
+
```
skerschb: very nice test suite!

worker/jobTypes/fastlyJob.js
varsha804: Oh whoops! I think I accidentally deleted that code 👀

worker/jobTypes/fastlyJob.js
varsha804: Ah, makes sense!

PR #59, opened 2020-01-03
Varsha804 docsp 7636

skerschb:


worker/jobTypes/S3Publish.js
```diff
@@ -13,26 +13,35 @@ class S3PublishClass {
       const exec = workerUtils.getExecPromise();
       const command = this.GitHubJob.deployCommands.join(' && ');
       const { stdout, stderr } = await exec(command);
```
madelinezec: does this not need to be inside a try/catch?

PR #60, opened 2020-01-06
Docsp 8124

skerschb:


Dockerfile.xlarge
```diff
@@ -33,10 +33,11 @@ WORKDIR /home/docsworker-xlarge
...
 RUN python3 -m pip uninstall -y snooty
 RUN python3 -m pip install --upgrade pip flit
-RUN git clone https://github.com/mongodb/snooty-parser.git snooty-parser && \
+RUN git clone https://github.com/mongodb/snooty-parser.git && \
 	cd snooty-parser && \
-	git fetch --all && \
-	git reset --hard origin/master && \
+	git fetch --tags && \
+	latestTag=$(git describe --tags `git rev-list --tags --max-count=1`) && \
+	git checkout $latestTag && \
```
i80and: In theory this should be safe, but it's generally a best practice to wrap shell variables in double-quotes; e.g. `git checkout "$latestTag"`

madelinezec:
built and ran autobuilder, successfully completed job

Dockerfile.xlarge
skerschb: changed, lemme know if ok.

danielborowski:
lgtm

PR #64, opened 2020-01-13
DOCSP 7930

danielborowski:


worker/jobTypes/githubPushJob.js
```diff
@@ -129,7 +132,15 @@ async function runGithubPush(currentJob) {
   }
 
   if (isMaster) {
-    // TODO: push to prod
+    // make sure user is entitled to deploy this repo
```
madelinezec: We only care about master logic for publishing builds right? Therefore if it is master, and in the case we dont want to stage master), we should reject it. But the entitlements checking should be reserved for productionDeployJob I believe?

PR #74, opened 2020-01-27
Docs 7567

madelinezec:


scripts/stagingUtils.js
```diff
@@ -0,0 +1,361 @@
...
+  ) {
+    const payload = {
+      jobType: "githubPush",
+      source: "github",
+      action: "push",
+      repoName: repoNameArg,
+      branchName: upstreamBranchName,
+      isFork: true,
+      private: false,
+      isXlarge: false,
```
skerschb: these should build on the xlarge builders

scripts/Makefile
```diff
@@ -0,0 +1,32 @@
...
+MAKEFLAGS += --silent
+
+# This allows us to accept extra arguments (by doing nothing when we get a job that doesn't match, 
+# rather than throwing an error).
+%: 
+    @: 
+
+# $(MAKECMDGOALS) is the list of "targets" spelled out on the command line
+stage: 
+	git clone --quiet https://github.com/madelinezec/test-submodules.git build_scripts
```
skerschb: We need to move this to a proper repository -- either back to workerpool or into a separate repo, although I'm not sure why we would want it to be separate? It has a fairly tight coupling with the docs-workerpool-code. Let's discuss.

skerschb:
I can't find the Makefile here or in the other repo. You could put it here under a directory called "build". If you'd like to add me to the other repo as collaborator, that will give me the ability to pull the PR down.

PR #92, opened 2020-02-06
Docsp 8796

madelinezec:


worker/jobTypes/githubJob.js
```diff
@@ -132,23 +132,51 @@ class GitHubJobClass {
...
           `git checkout ${currentJob.payload.branchName}`,
-          `git pull origin ${currentJob.payload.branchName}`,
-        ]);
+          `git branch ${currentJob.payload.branchName} --contains ${currentJob.payload.newHead}`
+        ];
+
+        const result = await exec(commitCheckCommands.join('&&'));
+        const returnObject = {};
+        
+        if (!result.stdout.includes(`* ${currentJob.payload.branchName}`)) {
```
skerschb: This means of checking is too brittle, may vary by version of git, by specifics of the relationship to the repository -- i.e., if a commit is in the repo but not that branch, etc. I think we can capture standard error here and whenever we have an error report back to the logger the text of the error -- this will make the slack output more elegant and will capture the full range of errors we might get. The githubPushJob has examples of how to capture standard error from an exec.

worker/jobTypes/githubJob.js
```diff
@@ -132,23 +132,51 @@ class GitHubJobClass {
...
+      } else {
+        pullRepoCommands.push(
+          ...[
+            `git checkout ${currentJob.payload.branchName}`,
+            `git pull origin ${currentJob.payload.branchName}`
+          ]
+        );
       }
 
       await exec(pullRepoCommands.join(' && '));
```
skerschb: let's check for standard error here to. In general, if we are running exec we need to check output for SE and report back/fail properly.

skerschb:
This PR isn't working for me. This the test I ran:

1. running a deploy on a commit that is not on the branch -- failed (correct behavior)
2. running a deploy from a suggested branch -- success (correct behavior)
3. running a deploy from a commit that is on the branch -- job hangs on push to stage (workerpool failed to pick up new messages even on deletion of message until I restarted the server)
4. restarted server
5. ran a deploy from a commit that is on the branch -- job hangs on push to stage
6. push another job, based solely on branch list in dropdown on deploy -- job is not picked up
7. delete job created in #5 above, job still isn't picked up
8. restart workerpool, job created in #6 is picked up and completes

It appears that when the commit provided is correct for the branch, the job completes the build but fails to push to "prod" (which is actually staging). The failure is such that the workerpool goes into a zombie state.

Please see if you can reproduce this behavior.

PR #96, opened 2020-02-12
add makefile for node

skerschb:
https://jira.mongodb.org/browse/DOCSP-8878

This PR adds the node production bucket to the makefile and adds targets to build and deploy to production with next gen (woo hoo!).

At this point, it's probably suboptimal to test this manually. So the CR here should probably consist of looking at the targets (make-next-gen-html and deploy-next-gen) to make sure they look like they could potentially work.

Since this prod bucket is new and not attached to any sort of CDN service or DNS record at the moment, we can test this with the staging autobuilder without impact to our existing docs site.

### DIFF
--- a/makefiles/Makefile.docs-node
+++ b/makefiles/Makefile.docs-node
@@ -5,7 +5,7 @@ STAGING_URL="https://docs-mongodborg-staging.corp.mongodb.com"
 STAGING_BUCKET=docs-mongodb-org-staging
 
 PRODUCTION_URL="https://docs.mongodb.com"
-PRODUCTION_BUCKET=docs-node
+PRODUCTION_BUCKET=docs-node-prod
 
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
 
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-node
```diff
@@ -23,10 +23,8 @@ help: ## Show this help message
 	@echo 'Variables'
 	@printf "  \033[36m%-18s\033[0m %s\n" 'ARGS' 'Arguments to pass to mut-publish'
 
-publish: remote-includes ## Builds this branch's publishable HTML and other artifacts under build/public
-	if [ ${GIT_BRANCH} = master ]; then rm -rf build/master build/public; fi
-	giza make publish
-	if [ ${GIT_BRANCH} = master ]; then mut-redirects config/redirects -o build/public/.htaccess; fi
+next-gen-publish: next-gen-html
```
i80and: If you add a comment with two leading hashes, e.g. `next-gen-publish: next-gen-html ## ...`, then this rule will be listed in `make help`

PR #100, opened 2020-02-13
Copy snooty front-end but exclude .git file

madelinezec:


makefiles/Makefile.docs-node
```diff
@@ -29,7 +29,7 @@ next-gen-publish: next-gen-html
 next-gen-html:
 	# snooty parse and then build-front-end
 	@-echo ${SNOOTY_DB_PWD} | snooty build "${REPO_DIR}" "mongodb+srv://${SNOOTY_DB_USR}:@cluster0-ylwlz.mongodb.net/snooty?retryWrites=true" --commit "${COMMIT_HASH}" || exit 0;
-	cp -r ${REPO_DIR}/../../snooty ${REPO_DIR};
+	rsync -az --exclude '.git' ${REPO_DIR}/../../snooty ${REPO_DIR};
```
i80and: Variables like `${REPO_DIR}` should always be wrapped in double-quotes in shell: e.g. `"${REPO_DIR}"`

Otherwise, spaces in the variables will badly mess things up.

If in doubt, always run things through https://www.shellcheck.net !

skerschb: Thanks for the tip on  shellcheck.net!

PR #101, opened 2020-02-14
Docsp 8976

madelinezec:
IN PROGRESS

### DIFF
--- a/worker/jobTypes/githubJob.js
+++ b/worker/jobTypes/githubJob.js
@@ -230,10 +230,26 @@ class GitHubJobClass {
         // check if need to build next-gen
         if (this.buildNextGen() ) {
             commandsToBuild[commandsToBuild.length - 1] = 'make next-gen-html';
-           }
+        }
+        
+         //overwrite gatsby config file
+        if (currentJob.payload.jobType === 'productionDeploy'){
+            const gatsbyConfigFileContents = await this.downloadGatsbyConfig();
+            if (makefileContents && makefileContents.status === 'success') {
+                await fs.writeFileSync(
+                    `repos/${this.getRepoDirName()}/snooty/gatsby-config.js`,
+                    makefileContents.content, {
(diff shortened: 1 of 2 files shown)


madelinezec:
I left our orig dockerfile lines commented out in case after we test prod, we want to go back to only using snooty release

madelinezec:
IN PROGRESS

worker/jobTypes/githubJob.js
```diff
@@ -213,8 +230,29 @@ class GitHubJobClass {
...
+                await fs.writeFileSync(
+                    `${snootyDir}/gatsby-config.js`,
+                    gatsbyConfigFileContents.content, {
+                        encoding: 'utf8',
+                        flag: 'w'
+                    }
+                );
+            } else {
+                console.log(
+                    'ERROR: gatsby config file does not exist in /configfiles directory on meta branch.'
```
skerschb: Won't this just get overwritten when snooty is rsync'ed over during make next-gen-html?

https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/Makefile.docs-node#L32

We could copy the file local to the repo but not within snooty dir, and then split out the next-gen-html target to a publish target that has the step that copies this config to repoDir/snooty?

Also we should make a failure to download this file a fatal error -- return control to the prodDeploy job and fail on the job.

worker/jobTypes/githubJob.js
```diff
@@ -213,8 +230,29 @@ class GitHubJobClass {
         // check if need to build next-gen
         if (this.buildNextGen()) {
             commandsToBuild[commandsToBuild.length - 1] = 'make next-gen-html';
+           }
+         //overwrite gatsby config file
+        if (currentJob.payload.jobType === 'productionDeploy'){
```
skerschb: This conditional needs a check for next gen as well, perhaps wrap in the previous.

PR #127, opened 2020-02-26
stitch fix for yaml module error

madelinezec:
temporary fix so we are still on track for release this week


madelinezec:
stitch side was fixed so this pr is obsolete

PR #139, opened 2020-03-06
Docsplat 868

madelinezec:


skerschb:
Let's hold off on this merge for now. I would rather see these tests living in the snooty-scripts repo.

madelinezec:
> Let's hold off on this merge for now. I would rather see these tests living in the snooty-scripts repo.

should I put the ticket in the backlog or keep in code review on jira?

skerschb:
We should remove this file from this repo as it's breaking tests. As a rule, we should not have unit tests for code that doesn't live in the repo in which the tests are run. I will remove it from master for now.

PR #141, opened 2020-03-25
fastly purge urls

danielborowski:


worker/utils/fastlyJob.js
```diff
@@ -12,6 +13,50 @@ class FastlyJobClass {
     }
   }
 
+  // takes in an array of urls and purges cache for each
+  async purgeCache(urlArray) {
+    if (!Array.isArray(urlArray)) {
+      throw new Error('Parameter `urlArray` needs to be an array of urls');
```
skerschb: nice type validation.

worker/utils/fastlyJob.js
```diff
@@ -12,6 +13,50 @@ class FastlyJobClass {
...
+    }
+
+    let urlCounter = urlArray.length;
+    let purgeMessages = [];
+
+    const headers = {
+      'fastly-key': environment.getFastlyToken(),
+      'accept': 'application/json',
+      'Fastly-Soft-Purge': '1',
```
skerschb: can you either create a var here with a name that suggests the purpose of this number, or put a comment here?

danielborowski: will do

skerschb:
I just noticed this PR is against master. We can't stage this build unless you merge to Stage. Please fix and reissue the PR.

worker/tests/unit/fastlyJob.test.js
```diff
@@ -24,8 +24,20 @@ const map = {
   'target': 'target'
 };
 
+// test purge urls
+const opsmanagerUrls = [
+  'https://docs.opsmanager.mongodb.com/current/installation/',
+  'https://docs.opsmanager.mongodb.com/current/core/requirements/',
+  'https://docs.mongodb.com/drivers/cxx/'
```
skerschb, line 31: I added a drivers url to see if this works for all properties. I think it would be good to add a bunch of other urls for valid pages that are unlikely to change across the different properties. We also need to add a test (or a few tests) to test failures -- urls that don't exist, for example. This is to make sure we recover from these scenarios properly.

worker/utils/fastlyJob.js
```diff
@@ -12,6 +12,62 @@ class FastlyJobClass {
...
+
+    return new Promise((resolve, reject) => {
+      for (let i = 0; i < urlArray.length; i++) {
+        // perform request to purge
+        request({
+          method: 'PURGE',
+          url: urlArray[i],
+          headers: headers,
+        }, function(err, response, body) {
+          // url was not valid to purge
```
skerschb, line 40: I thought we can't determine when a url is valid?

worker/utils/fastlyJob.js
```diff
@@ -12,6 +12,62 @@ class FastlyJobClass {
...
+          if (!response) {
+            utils.logInMongo(that.currentJob, `Error: service for this url does not exist in fastly for purging ${urlArray[i]}`);
+            purgeMessages.push({
+              'status': 'failure',
+              'message': `url ${urlArray[i]} does not exist in fastly`
```
skerschb, line 45: Is this accurate as an error message?

worker/utils/fastlyJob.js
danielborowski: discussed over DM

PR #188, opened 2020-05-05
adding patchID

skerschb:


sophstad:
As discussed on Slack, `--patch` is not a required flag, so we don't need to include `--patch=None` when running the parser. We will only use `--patch` flags when a patch ID has been generated.

PR #111, opened 2020-02-24
Docs 8452 - Regression Tests

madelinezec:
Would love your CR when you have a moment. Hoping to get this merged tomorrow so I can test my comparison functions in stitch

PR #195, opened 2020-05-14
Add Regression Test Stitch Functions

madelinezec:
Creating this PR so that regression test stitch functions can be reviewed before we [merge this PR](https://github.com/madelinezec/docs-worker-pool/tree/DOCS-8452/scripts/build).


makefiles/Makefile.devhub-content
```diff
@@ -27,14 +27,13 @@ help: ## Show this help message
 
 update-snooty:
 	cd ../../snooty-devhub; \
-	git checkout master; \
-	git pull; \
+	git fetch --all; \
```
skerschb, line 30: let's remove the makefiles from this PR -- the way the merge was resolved will break our devhub-content builds.

StitchApp/RegressionTestAppFunctions/compareChildJobs/source.js
```diff
@@ -0,0 +1,63 @@
...
+    const testStatus = testJob.status;
+    const prodStatus = prodJob.status;
+    const testLogs = testJob.logs; 
+    const prodLogs = prodJob.logs;
+    //compare time it took in seconds, only get an endTime if job completed 
+    //also compare results of curl-ing the link in Summary, which is only generated if job completes
+    if (testStatus === "completed"){
+      testTime = (testJob.endTime - testJob.startTime)/1000;
+      //lastMessage.indexOf('Summary')
+      testLink = testLogs[testLogs.length - 1].split('Hosted at:')[0];
```
skerschb, line 27: We need to adjust our measure of fail and success and rely solely on status (this will be possible once we parameterize the parser to emit a fail status of diagnostics). That way we can leverage each component of snooty to do its job in reporting pass and fail state rather than coding brittle measures. For now, let's rely on status to at least capture catastrophic failures and once the diagnostics can be set to influence the status, we'll have better coverage.

StitchApp/RegressionTestAppFunctions/compareChildJobs/source.js
```diff
@@ -0,0 +1,63 @@
...
+      // The response body is encoded as raw BSON.Binary. Parse it to JSON.
+          const ejson_body = EJSON.parse(response.body.text());
+          testLinkCode = ejson_body;
+          console.log("this is testlink ", testLinkCode, testTime, testLink);
+        })
+    }
+    
+    if (prodStatus === "completed"){
+       prodTime = (prodJob.endTime - testJob.startTime)/1000;
+       prodLink = prodLogs[prodLogs.length - 1].split('Hosted at:')[0];
```
skerschb, line 39: same here

StitchApp/RegressionTestAppFunctions/sendSummaryOnSlack/source.js
```diff
@@ -0,0 +1,63 @@
...
+  const scheme = "https";
+  const host = "slack.com";
+  const postMessage = "api/chat.postMessage";
+  const lookupUser = "api/users.lookupByEmail"; 
+  const token = context.values.get("slack_token");
+  
+  // Get the Slack Service
+  const slack = context.services.get("slackHTTPService");
+  
+  const admins = context.functions.execute("getAdmins");
```
skerschb, line 13: switch to function that calls db.

makefiles/Makefile.devhub-content
madelinezec: Ah my apologies, was so busy exporting the app from stitch that I forgot to stash these changes

madelinezec:
This PR contains the restructured stitch app. These functions live in the `workerpool_regression_test` app on stitch. 

**To test this PR**: 

1. add a record containing your slack email to the staging database's admins collection. This will ensure you get the slack summary once all the jobs are finished. Because the child processes are regular githubPush jobs, you will still get the slack updates associated with those jobs as well.
2. boot up the version of the workerpool contained in [this PR](https://github.com/mongodb/docs-worker-pool/pull/111)
3.  Enqueue a regression test by pushing an empty commit to to [snooty-regression-test repo](https://github.com/10gen/snooty-regression-tests/). 
**Please Read These Important Notes**
You cannot redeliver a payload because all regression child processes are associated with a newHead commit hash. And the program only expects as many child processes as there are repos to test. So if you redeliver the payload, that will create more child processes associated with that same newHead commitHash, breaking the app (it doesn't break down - but it "breaks" the intended behavior)

We are not currently enqueuing jobs to the production server. Currently, for testing purposes, we just enqueue one set of jobs to the staging server - you can verify this by checking at `StitchApp/RegressionTest/githubService/regressionTestEnv.js`. We recreate the behavior of comparing two sets of jobs by just querying the same database twice and compare each. jobs to itself - check out `StitchApp/RegressionTest/RegressionTestAppFunctions/retrieveRegressionChildJobs/source.js` and `StitchApp/RegressionTest/RegressionTestAppFunctions/compareChildJobs/source.js` to see how this works. Once this PR is approved, we will change the secrets' values in the stitch app to give it access to prod db and also uncomment the line in regressionTestEnv.js so that we insert in prod as well as in staging.

StitchApp/RegressionTest/RegressionTestAppFunctions/compareChildJobs/source.js
```diff
@@ -0,0 +1,25 @@
+exports = function(commitHash, testJobs, prodJobs){
```
skerschb, line 1: Realm functions are a bit disembodied because they are executed in sometimes mysterious ways that aren't obvious when examining the source -- a good best practice moving forward would be to leave a comment in the code to describe the function -- what calls it and what it does. I'm confused what the locally scoped testTime, etc. does here, are we missing something?

StitchApp/RegressionTest/RegressionTestAppFunctions/insertJobinProdDB/source.js
```diff
@@ -0,0 +1,41 @@
+exports = function(payloadArg, jobTitle, jobUserName, jobUserEmail){
```
skerschb, line 1: We don't want to be writing to the prod and dev databases from the regression tests app for a number of reasons:

1. It doesn't test our job creation infrastructure -- which is an important part of our build system and which does change from time to time/require regression testing.
2. It opens up a potential vulnerability

Can we move the execution of the job initialization to be triggered by a webhook that is dedicated to regression tests and runs from within the environment?

StitchApp/RegressionTest/RegressionTestAppFunctions/insertJobinTestDB/source.js
```diff
@@ -0,0 +1,38 @@
+exports = function(payloadArg, jobUserName, jobUserEmail){
```
skerschb, line 1: same here.

StitchApp/RegressionTest/RegressionTestAppFunctions/retrieveRegressionChildJobs/source.js
```diff
@@ -0,0 +1,62 @@
+exports = async function(changeEvent) {
```
skerschb, line 1: comment on how this is executed.

StitchApp/RegressionTest/RegressionTestAppFunctions/retrieveRegressionChildJobs/source.js
```diff
@@ -0,0 +1,62 @@
...
+          prodDbComplete = true;
+          items.forEach(function(childJob) {
+            regression_collection.insertOne(childJob);
+            prodJobs[childJob.payload.repoName] = childJob;
+          })
+        }
+      })
+  .catch(err => console.error(`Failed to find documents: ${err}`));
+  if(stageDbComplete === true && prodDbComplete === true){
+    console.log("hi!")
```
skerschb, line 57: probably want to remove this.

StitchApp/RegressionTest/RegressionTestAppFunctions/compareChildJobs/source.js
madelinezec: ah no just artifacts from when we were going to compare the time it took, num of failures, etc etc. ill delete

madelinezec:
The new structure is as follows:

A push to the [snooty-regression-tests](https://github.com/10gen/snooty-regression-tests/) repo hits the webhook in staging and prod apps. 

The associated function enqueues test jobs to their respective databases. 

A trigger in the RegressionTests stitch app is fired whenever a child process finishes. It calls a function that determines if all child processes are completed. Once all child processes are completed, another function is called to compare the results between the two environments and send the results in a slack bot message.

PR #196, opened 2020-05-19
DOP-64 and DOP-829

madelinezec:
This PR serves as a reminder to do integration tests for DOP-64 and DOP-829 before merging into master.


madelinezec:
found a bug in testing

PR #197, opened 2020-05-20
Update Makefile.docs-bi-connector for catching parser failure

madelinezec:
This makefile logic allows us to throw an error ONLY when parser throws error code 1 in the course of running `snooty build...`. 

We exit 0 on all other errors.

### DIFF
--- a/makefiles/Makefile.docs-bi-connector
+++ b/makefiles/Makefile.docs-bi-connector
@@ -29,7 +29,12 @@ html: ## Builds this branch's HTML under build/<branch>/html
 next-gen-html:
 	# snooty parse and then build-front-end
 	curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-bi-connector.yaml > ${REPO_DIR}/published-branches.yaml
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-bi-connector
```diff
@@ -29,7 +29,12 @@ html: ## Builds this branch's HTML under build/<branch>/html
 next-gen-html:
 	# snooty parse and then build-front-end
 	curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-bi-connector.yaml > ${REPO_DIR}/published-branches.yaml
-	@-echo ${SNOOTY_DB_PWD} | snooty build "${REPO_DIR}" "mongodb+srv://${SNOOTY_DB_USR}:@cluster0-ylwlz.mongodb.net/snooty?retryWrites=true" --commit "${COMMIT_HASH}" || exit 0;
+	@echo ${SNOOTY_DB_PWD} | snooty build "${REPO_DIR}" "mongodb+srv://${SNOOTY_DB_USR}:@cluster0-ylwlz.mongodb.net/snooty?retryWrites=true" --commit "${COMMIT_HASH}";\
```
skerschb, line 32: We've typically condensed makefile scripting instead of having it span multiple lines and that prevents the targets from bloating beyond what a terminal screen will show at once. Will this work condensed into one line? If so let's stick to that convention for now.

madelinezec: I could unfortunately not get it to work in one line! I believe conditionals in make are something that only work with `\` and new lines. It seems to be very sticky!

PR #210, opened 2020-05-28
Stage

skerschb:


worker/jobTypes/githubJob.js
```diff
@@ -314,11 +296,14 @@ class GitHubJobClass {
                 });
             });
         } catch (error) {
+          if (error.code === 1) {
```
madelinezec, line 299: hmmm if we know this code is broken - do we want to merge it to master? without makefile changes, no errors will be generated. but technically this stanza is unnecessary in this PR

skerschb: Merged DOP-64 so this should be resolved.

worker/jobTypes/S3Publish.js
```diff
@@ -69,11 +72,39 @@ class S3PublishClass {
...
-      if (stdout.indexOf('Summary') !== -1) {
-        stdoutMod = stdout.substr(stdout.indexOf('Summary'));
+      let stdoutMod = stdout;
+
+      // check if json was returned from mut
+      try {
+        const stdoutJSON = JSON.parse(stdout);
+        const urls = stdoutJSON.urls;
+        // pass in urls to fastly function to purge cache
+        this.fastly.purgeCache(urls).then(function(data) {
```
i80and, line 82: This function is async -- could we await here instead of promise-chaining?

worker/jobTypes/S3Publish.js
```diff
@@ -69,11 +72,39 @@ class S3PublishClass {
...
+            }
+            // if over certain length, send as a single slack message and reset the array
+            if (batchedUrls.length > 20 || i >= (urls.length - 1)) {
+              logger.sendSlackMsg(`${batchedUrls.join('\n')}`);
+              batchedUrls = [];
+            }
+          }
+        });
+      } catch(e) {
+        // if not JSON, then it's a normal string output from mut
```
i80and, line 101: It looks like this is only meant to handle exceptions from the `JSON.parse()` call. Would it make sense to limit the scope of the try...catch to just that statement?

(Due to JS being weird you'd have to `let stdoutJSON = null` outside the try)

worker/jobTypes/S3Publish.js
danielborowski: Explained offline -- we do not want to actually stop execution of this function to wait for the purge to finish

PR #211, opened 2020-05-29
new docs makefile

skerschb:
These makefile changes enable patch id to be incorporated into builds of branches that have uncommitted/unpushed patches applied.

It enables the use of the commitless staging script with the server docs repository.

This makefile was tested using docs-stage.

### DIFF
--- a/makefiles/Makefile.docs
+++ b/makefiles/Makefile.docs
@@ -102,9 +102,9 @@ deploy-search-index: ## Update the search index for this branch
 	fi
 
 patchapply:
-	echo "GATSBY_SITE=${PROJECT}" > .env.production; \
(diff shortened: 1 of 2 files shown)


makefiles/Makefile.docs
```diff
@@ -92,9 +101,16 @@ deploy-search-index: ## Update the search index for this branch
 		mut-index upload build/public/${GIT_BRANCH} -o manual-${GIT_BRANCH}.json -u ${PRODUCTION_URL}/${GIT_BRANCH} -s; \
 	fi
 
-next-gen-html: examples
+patchapply:
+	echo "GATSBY_SITE=${PROJECT}" > .env.production; \
```
i80and, line 105: These don't need to be joined into one expression

PR #228, opened 2020-06-04
Split deploy

danielborowski:


skerschb:
I did attempt to test this PR against docs-node, but it appears the --json flag has been reverted. Please post on this PR how I may test both with and without --json. If makefile changes are required, please issue PRs for those makefiles as well so they can be merged first. Also, as a reminder: each PR should have the referring ticket, and the test plan in the PR description.

PR #244, opened 2020-07-22
Proposed Makefile Changes for DOP-1238

madelinezec:
[Relevant ticket](https://jira.mongodb.org/browse/DOP-1238)
[Spec](https://docs.google.com/document/d/1CvHbhA1lQEx4aEg109UVgkw2h_vCvC7wwve2mXEtRvU/edit?usp=sharing)

### DIFF
--- a/makefiles/Makefile.docs-node
+++ b/makefiles/Makefile.docs-node
@@ -63,7 +63,7 @@ next-gen-stage: ## Host online for review
 		echo "we have args!!!  $(filter-out $@,$(MAKECMDGOALS))"; \
 	else \
 		echo "we dont have args!!!!"; \
-		mut-publish public ${STAGING_BUCKET} --prefix="${COMMIT_HASH}/${PREFIX}" --stage ${ARGS}; \
+		mut-publish public ${STAGING_BUCKET} --prefix="${COMMIT_HASH}/${PROJECT}" --stage ${ARGS}; \
 		echo "Hosted at ${STAGING_URL}/${COMMIT_HASH}/${PREFIX}/${USER}/${GIT_BRANCH}/"; \
 	fi
 %:
--- a/makefiles/Makefile.docs-node
+++ b/makefiles/Makefile.docs-node
@@ -43,6 +43,7 @@ next-gen-html:
 	fi
 	rsync -az --exclude '.git' ${REPO_DIR}/../../snooty ${REPO_DIR};
 	cp "${REPO_DIR}/.env.production" "${REPO_DIR}/snooty/.env.production"
+  echo ${REPO_DIR}
 	cd snooty; \
 #	echo "GATSBY_SITE=${PROJECT}" > .env.production; \
 	npm run build; \
--- a/makefiles/Makefile.docs-node
+++ b/makefiles/Makefile.docs-node
@@ -25,9 +25,9 @@ help: ## Show this help message
 	
 next-gen-publish: 
 	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-node.yaml > ${REPO_DIR}/published-branches.yaml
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-node
```diff
@@ -59,9 +49,12 @@ next-gen-html:
 	echo "GATSBY_PARSER_USER=${USER}" >> .env.production; \
 	echo "GATSBY_PARSER_BRANCH=${GIT_BRANCH}" >> .env.production; \
 	echo "COMMIT_HASH=${COMMIT_HASH}" >> .env.production; \
+	if [ $(filter-out $@,$(MAKECMDGOALS)) = publish]; then \
+		echo "PATH_PREFIX=${PREFIX}/${GIT_BRANCH}" >> .env.production; \
```
skerschb, line 53: This would put master into the prefix for the site, is that what we want for this repository?

makefiles/Makefile.docs-node
```diff
@@ -23,28 +22,19 @@ help: ## Show this help message
 	@echo 'Variables'
 	@printf "  \033[36m%-18s\033[0m %s\n" 'ARGS' 'Arguments to pass to mut-publish'
 
-next-gen-html-publish:
-	# snooty parse and then build-front-end
+download-published-branches:
```
skerschb, line 25: mentioned in the spec that we might want to give this a generic target name.

makefiles/Makefile.docs-node
```diff
@@ -23,28 +22,19 @@ help: ## Show this help message
...
-	
+
+deploy-search-index: ## Update the search index for this branch
+	@echo "Building search index"
+	mut-index upload build/public -o ${PROJECT}-${GIT_BRANCH}.json -u ${PRODUCTION_URL}/${PROJECT} -g -s --exclude build/public/sdk/iOS
+
+%:
+	@:
+
 next-gen-html:
```
sophstad, line 38: Doesn't this target have a dependency on the new `download-published-branches` target?

makefiles/Makefile.docs-node
madelinezec: yes! i checked with @sophstad

makefiles/Makefile.docs-node
madelinezec: we would actually run each target separately, see:https://github.com/madelinezec/docs-worker-pool/blob/DOP-1238/worker/jobTypes/githubJob.js#L275-L276

sophstad: Interesting, what is the reasoning for that?

makefiles/Makefile.docs-node
sophstad: Sue is right actually, we don't want to include the version because node is an unversioned repository. This is where we need to include the logic we discussed over Slack about determining whether or not the repo is versioned:
> If, in the published-branches.yaml file, if the length of version.active > 1, then it is versioned
If the repo is _not_ versioned, then the branch should be omitted.

makefiles/Makefile.docs-node
madelinezec: It can be very difficult to call one makefile target from another target

madelinezec: And in this case, you can't have next-gen-publish: download-published-branches next-gen-html ${parameter}, whereas calling each individually allows us to pass a commandline arg to next-gen-html

makefiles/Makefile.docs-node
madelinezec: You are right - I forgot that this docs-node is not versioned! 

Sophie in dms with Sue I've been discussing another option:
Remove all the `.env.production` writing logic from next-gen-html and instead have the server write a `.env.production` file that next-gen-html then copies into snooty. What do you think?

sophstad: Sounds great — much easier to maintain consistency across properties. Good idea!

PR #245, opened 2020-07-22
Proposed Server Side Changes for DOP-1238

madelinezec:
This ticket provides the functionality required by [DOP-1238](https://jira.mongodb.org/browse/DOP-1238) and [DOP-1253](https://jira.mongodb.org/browse/DOP-1253)

Server team requested that `stagel commit` jobs would rewrite the same url for each staging job. This is achieved here:
this [job](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?jobId=5f335b542f9a5cfb5b91b0a1) and [this job](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?jobId=5f335c182f9a5cfb5b91d3ee) both write to this url: https://docs-mongodbcom-staging.corp.mongodb.com/compass/madelinezec/aug11/docsworker-xlarge/master/, where aug11 is the name of the branch. 

To test `stagel commit`, you will need to follow the [instructions here](https://wiki.corp.mongodb.com/pages/viewpage.action?pageId=111777809) with the caveat that you need to adjust one of the commands of the stagel target:
`@ cd build_scripts && git checkout DOP-1253 && npm list mongodb || npm install mongodb`
This will include the type of patch build you are doing in the job payload. There is a PR open for the server script [here](https://github.com/mongodb/snooty-scripts/pull/8)

To test deploy, you will need to add get-build-dependencies and change next-gen-deploy:
download-published-branches:
```
	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/${repoName}.yaml > ${REPO_DIR}/published-branches.yaml

next-gen-deploy:
	@yes | mut-publish public ${PRODUCTION_BUCKET} --prefix="$(filter-out $@,$(MAKECMDGOALS))" --deploy --deployed-url-prefix=https://docs.mongodb.com --json --all-subdirectories ${ARGS};
	echo "Hosted at ${PRODUCTION_URL}/$(filter-out $@,$(MAKECMDGOALS))"; 
%:
	@:
```

I fake deployed [docs-node](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?jobId=5f33625c2f9a5cfb5b92cedc) (non-versioned repo) and [docs-bi-connector](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?jobId=5f336c532f9a5cfb5b945a96) (versioned), which generated these URLs respectively: 
https://docs.mongodb.com/drivers/node/
https://docs.mongodb.com/bi-connector/v2.12

Relevant Makefile changes: https://github.com/mongodb/docs-worker-pool/pull/285


worker/jobTypes/githubJob.js
```diff
@@ -40,6 +42,56 @@ class GitHubJobClass {
...
+
+      if(isProdDeployJob){
+        //download published branches file to check if repo is versioned 
+        const repoObject = {
+          repoOwner: this.currentJob.payload.repoOwner, repoName: this.currentJob.payload.repoName,
+        };
+        const repoContent = await workerUtils.getRepoPublishedBranches(repoObject)
+        //versioned repo
+        if(repoContent && repoContent.content.version.active.length > 1){
+          pathPrefix = `${this.currentJob.payload.repoName.replace('docs-','')}/docsworker-xlarge/${this.currentJob.payload.branchName}`;
```
skerschb, line 57: I'm not sure that our already-published prefixes always match the reponame. Is this verified? For example, cloud-docs, mms-docs, etc.

Also, we don't typically deploy with a user in the prefix. Only stage.

worker/jobTypes/githubJob.js
```diff
@@ -40,6 +42,56 @@ class GitHubJobClass {
...
+          repoOwner: this.currentJob.payload.repoOwner, repoName: this.currentJob.payload.repoName,
+        };
+        const repoContent = await workerUtils.getRepoPublishedBranches(repoObject)
+        //versioned repo
+        if(repoContent && repoContent.content.version.active.length > 1){
+          pathPrefix = `${this.currentJob.payload.repoName.replace('docs-','')}/docsworker-xlarge/${this.currentJob.payload.branchName}`; 
+        }
+        //non-versioned repo
+        else{
+          pathPrefix = `${this.currentJob.payload.repoName.replace('docs-','')}`
```
skerschb, line 61: Same question here.

worker/jobTypes/githubJob.js
```diff
@@ -40,6 +42,56 @@ class GitHubJobClass {
...
+        }
+      }
+      // server staging commit jobs
+      else if(this.currentJob.payload.patch && this.currentJob.payload.patchType === 'commit'){
+        pathPrefix = `${this.currentJob.payload.repoName.replace('docs-','')}/${this.currentJob.user}/${this.currentJob.payload.localBranchName}/docsworker-xlarge/master`;
```
skerschb, line 66: And here.

worker/jobTypes/githubJob.js
skerschb, line 57: It seems like sometimes we use the current user and sometimes we hardcode the user. We should not hardcode the user -- when we aren't using current user we should be using system user.

worker/jobTypes/githubJob.js
```diff
@@ -268,6 +309,34 @@ class GitHubJobClass {
...
+      ];
+
+      // check if need to build next-gen
+      if (this.buildNextGen() && !isProdDeployJob) {
+        commandsToBuild[commandsToBuild.length - 1] = 'make next-gen-html';
+      }
+
+      //check if prod deploy job
+      if (isProdDeployJob) {
+          commandsToBuild[commandsToBuild.length - 1] = 'make download-published-branches';
```
skerschb, line 337: This target is very specific in name -- can we call it something like get-build-dependencies?

worker/jobTypes/githubJob.js
```diff
@@ -2,6 +2,8 @@ const fs = require('fs-extra');
 const workerUtils = require('../utils/utils');
 const simpleGit = require('simple-git/promise');
 const request = require('request');
+const logger = require('../utils/logger');
+const S3Publish = require('./S3Publish');
```
skerschb, line 6: We have avoided putting a binding to S3 in the Github job because it implies a dependency on S3 which should not be the case for these operations. I didn't see a reference to this const in this PR though so do we need it?

madelinezec: We do not need logger no - I'm sorry i should have closed this PR after that dummy spec I wrote for this (and then reopened it when officially ready). I was still in the process of cleaning everything up!

worker/jobTypes/githubJob.js
madelinezec: this is also a clean up issue! This will not be the case after this afternoon

madelinezec:
This PR should not have been opened originally - it was used in a dummy spec to give concrete examples of my plans. 

I will reopen once we've decided on a way forward with the repo name to path prefix dilemma

worker/jobTypes/githubJob.js
skerschb: Yes, please close the PR and reopen. Let me know when ready to review (I'm assuming it's still not ready because these lines are still in the code).

worker/jobTypes/githubJob.js
```diff
@@ -268,6 +318,36 @@ class GitHubJobClass {
...
+          `make html`
+      ];
+
+      // check if need to build next-gen
+      if (this.buildNextGen() && !isProdDeployJob) {
+        commandsToBuild[commandsToBuild.length - 1] = 'make next-gen-html';
+      }
+
+      //check if prod deploy job
+      if (isProdDeployJob) {
```
skerschb, line 347: we need to check if this is a next gen build.

worker/jobTypes/githubJob.js
```diff
@@ -268,6 +318,36 @@ class GitHubJobClass {
                 'ERROR: makefile does not exist in /makefiles directory on meta branch.'
             );
         }
+        //set up env vars for all jobs
+        const pathPrefix = await this.writeEnvFile(isProdDeployJob)
```
skerschb, line 322: should only run for next gen jobs

worker/jobTypes/githubJob.js
```diff
@@ -40,7 +42,66 @@ class GitHubJobClass {
...
+    
+    async writeEnvFile(isProdDeployJob){
+      //download published branches file to retrieve prefix and check if repo is versioned 
+      const repoObject = {
+        repoOwner: this.currentJob.payload.repoOwner, repoName: this.currentJob.payload.repoName,
+      };
+      const repoContent = await workerUtils.getRepoPublishedBranches(repoObject)
+      var pathPrefix = repoContent.content.prefix
+
+      if(isProdDeployJob){
```
skerschb, line 54: I would pull this out into a function -- get the path prefix from a compartmentalized function.

worker/jobTypes/githubJob.js
```diff
@@ -40,7 +42,66 @@ class GitHubJobClass {
...
+        //versioned repo
+        if(repoContent && repoContent.content.version.active.length > 1){
+          pathPrefix += `/${this.currentJob.payload.branchName}`; 
+        }
+      }
+      // server staging commit jobs
+      else if(this.currentJob.payload.patch && this.currentJob.payload.patchType === 'commit'){
+        const server_user = await this.getUser()
+        pathPrefix += `/${this.currentJob.user}/${this.currentJob.payload.localBranchName}/${server_user}/master`; 
+      }
```
skerschb, line 64: can't hardcode the branch here (versioned branches)

madelinezec: I believe the branch name can be `{this.currentJob.payload.localBranchName}` for `make stagel commit` jobs as they will be working from the branch they wish to stage

worker/jobTypes/githubJob.js
madelinezec: got it - I mistakenly thought we only published next-gen

madelinezec:
@skerschb ready for review

worker/jobTypes/githubJob.js
```diff
@@ -41,13 +41,97 @@ class GitHubJobClass {
...
+            return mutPrefix
+          }
+          return null
+        } catch (error) {
+         console.log(error)
+         throw error 
+        }
+
+  }
+    async getUser(){
```
skerschb, line 115: This function should live somewhere a little more non-build-specific -- in the utils.js perhaps?

worker/jobTypes/githubJob.js
```diff
@@ -268,6 +338,40 @@ class GitHubJobClass {
                 'ERROR: makefile does not exist in /makefiles directory on meta branch.'
             );
         }
+        //set up env vars for all next-gen jobs
+        if(this.buildNextGen()){
+          const pathPrefix = await this.writeEnvFile(isProdDeployJob)
```
skerschb, line 343: We have an opportunity here -- to split out the writeEnvFile into a Gatsby adapter. That way all things Gatsby-related will be in one module. Then we know exactly where to go when things related to Gatsby change, *and* we can easily swap out Gatsby for another provider, maybe even at run time eventually.

Could we change writeEnvFile to constructing our GatsbyAdapter object -- by passing in the job -- and writing the file with an initEnv method? Then we can persist the prefix to the job object as well instead of returning it in this method.

When we eliminate the makefiles we will abstract these calls and allow the instantiated builder to manage the differences between the different build types.

worker/jobTypes/githubJob.js
```diff
@@ -41,13 +41,97 @@ class GitHubJobClass {
...
+      }catch(error){
+        console.log(error)
+        throw error
+      }
+        
+    }
+
+    async writeEnvFile(isProdDeployJob){
+        try {
+          const [pathPrefix, mutPrefix, server_user] = await this.constructPrefix(isProdDeployJob)
```
skerschb, line 80: split out into a Gatsby adapter as mentioned below.

worker/jobTypes/githubJob.js
```diff
@@ -41,13 +41,97 @@ class GitHubJobClass {
...
+          let envVars;
+
+          if(pathPrefix !== null){
+            envVars = 
+            `GATSBY_PARSER_USER=${server_user}
```
skerschb, line 85: since both conditional clauses write the same thing for GATSBY_PARSER_USER and GATSBY_PARSER_BRANCH, this can be condensed.

worker/jobTypes/githubJob.js
```diff
@@ -41,13 +41,97 @@ class GitHubJobClass {
...
+          }
+    
+          fs.writeFile(`repos/${this.getRepoDirName()}/.env.production`, envVars,  { encoding: 'utf8', flag: 'w' }, function(err) {
+              if(err) {
+                console.log(`error writing .env.production file: ${err.stderr}`);
+                throw err;
+              }
+          }); 
+          //pass mutprefix back to caller to save in prefix field of currentJob, which we pass to stage and deploy targets
+          if(mutPrefix !== null){
```
skerschb, line 105: we can persist this to the job object directly when its passed to the GatsbyAdapter.

madelinezec:
Update:

`constructPrefix` now constructs the path prefix and mut prefix and save both to the job object. 
`writeEnvFile` has been replaced by `GatsbyAdapter.initEnv()`. `GithubJob.buildRepo()` calls `constructPrefix` and then passes the current job to`GatsbyAdapter.initEnv()`, which grabs the path prefix and saves it to .env.production. Both of these calls happen before `next-gen-html`, this order is necessary as this target runs the front end which needs `.env.production` copied to `snooty/` before `npm run build` is called

worker/jobTypes/productionDeployJob.js
```diff
@@ -68,7 +68,7 @@ function safeGithubProdPush(currentJob) {
 async function startGithubBuild(job, logger) {
   const buildOutput = await workerUtils.promiseTimeoutS(
     buildTimeout,
-    job.buildRepo(logger),
+    job.buildRepo(logger, true),
```
skerschb, line 71: we don't need the adapter here?

worker/utils/utils.js
```diff
@@ -168,7 +182,7 @@ module.exports = {
   },
 
   async getRepoPublishedBranches(repoObject) {
-    const pubBranchesFile = `https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/${repoObject.repoName}.yaml`;
+    const pubBranchesFile = `https://raw.githubusercontent.com/madelinezec/docs-worker-pool/meta-prefix-work/publishedbranches/${repoObject.repoName}.yaml`;
```
skerschb, line 185: we won't merge with your personal id in this path.

PR #290, opened 2020-08-18
add commit hash

skerschb:


skerschb:
For https://jira.mongodb.org/browse/DOP-1335

Adding the --redirects-prefix flag.

PR #302, opened 2020-09-03
DOP-1453

madelinezec:


skerschb (approved):
There is a format issue with the brackets but let's test this and fix the cosmetics on next iteration.

PR #306, opened 2020-09-09
add filter for slack output in logger

skerschb:


madelinezec (approved):
really great, makes sense for us to get some use out of that logger class. lgtm

PR #307, opened 2020-09-10
Fix for devhub-content deploy

madelinezec:


makefiles/Makefile.devhub-content
```diff
@@ -62,13 +62,14 @@ next-gen-stage: ## Host online for review
 	@:
 
 next-gen-deploy:	
-	@yes | mut-publish public ${PRODUCTION_BUCKET} --prefix="$(filter-out $@,$(MAKECMDGOALS))" --deploy --deployed-url-prefix=https://docs.mongodb.com --json --all-subdirectories ${ARGS};
-	@echo "Hosted at ${PRODUCTION_URL}/$(filter-out $@,$(MAKECMDGOALS))/index.html";
-	$(MAKE) deploy-search-index
```
madelinezec, line 67: removed the call to deploy-search-index bc I checked previous version of this makefile and saw the deploy target did not call deploy-search-index.

makefiles/Makefile.docs-landing
```diff
@@ -68,7 +68,7 @@ next-gen-publish: build
 
 next-gen-deploy: build
 	@yes | mut-publish public ${PRODUCTION_BUCKET} --prefix="$(filter-out $@,$(MAKECMDGOALS))" --deploy --deployed-url-prefix=https://docs.mongodb.com --json --all-subdirectories ${ARGS};
-	@echo "Hosted at ${PRODUCTION_URL}/$(filter-out $@,$(MAKECMDGOALS))";
+	@echo "Hosted at ${PRODUCTION_URL}";
```
madelinezec, line 71: here I just wanted to clean up the echo

makefiles/Makefile.devhub-content
```diff
@@ -62,13 +62,14 @@ next-gen-stage: ## Host online for review
...
+	@yes | mut-publish public ${PRODUCTION_BUCKET} --prefix="$(filter-out $@,$(MAKECMDGOALS))" --deploy --deployed-url-prefix=https://developer.mongodb.com --json --all-subdirectories ${ARGS};
+	@echo "Hosted at ${PRODUCTION_URL}/index.html";
```
skerschb, line 66: Should deployed url prefix be the  PRODUCTION_URL var?

skerschb (commented):
just one nit

makefiles/Makefile.devhub-content
madelinezec: we can do that - but that is not the case in the other makefiles that I used as a model

madelinezec: what would you like to do in this case?

skerschb: even docs-landing? This should follow the docs-landing model.

PR #308, opened 2020-09-11
Release 0.4.6

skerschb:
Autobuiler 0.4.6

- Remove extra slack output at production deploy (https://jira.mongodb.org/browse/DOP-1457)


worker/utils/logger.js
```diff
@@ -1,19 +1,39 @@
...
-  }
+    save(message) {
+        workerUtils.logInMongo(this.currentJob, message);
+    }
+
+    static trimOutputForUserFacingLogMessages(output) {
+        let trimmedOutput = '';
+        const splitOutput = output.split('\n');
+        splitOutput.forEach((line) => {
+            if (line.indexOf('WARNING') > -1 || line.indexOf('ERROR') > -1 || line.indexOf('INFO') > -1) {
```
i80and, line 21: Having an INFO message probably doesn't require that we dump the entire build log, but the point is a little moot because we don't currently *have* any INFO messages :D

i80and (approved):

worker/utils/logger.js
skerschb: It would only publish the INFO log line -- and I did see one in testing. It might have been from a legacy build.

skerschb: Correction - it's next gen: the snooty parser version is an INFO log.

PR #309, opened 2020-09-14
DOP-1487: add legacy build target to next-gen-stage target

schmalliso:
I pulled out the targets that were specific to giza since they're not being used as well as the `next-gen-publish` that Maddie mentioned was deprecated.

Please let me know if I was overzealous anywhere.

### DIFF
--- a/makefiles/Makefile.docs-landing
+++ b/makefiles/Makefile.docs-landing
@@ -27,8 +27,7 @@ help:
 get-build-dependencies: 
 	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-landing.yaml > ${REPO_DIR}/published-branches.yaml
 
-next-gen-html:
-	# snooty parse and then build-front-end
+next-gen-html: ## snooty parse and then build-front-end
 	@echo ${SNOOTY_DB_PWD} | snooty build "${REPO_DIR}" "mongodb+srv://${SNOOTY_DB_USR}:@cluster0-ylwlz.mongodb.net/snooty?retryWrites=true" --commit "${COMMIT_HASH}"; \
 	if [ $$? -eq 1 ]; then \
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-landing
```diff
@@ -43,64 +42,31 @@ next-gen-html:
...
-next-gen-stage: ## Host online for review
+next-gen-stage: build  ## Host online for review
 	if [ $(filter-out $@,$(MAKECMDGOALS)) ]; then \
 		mut-publish public ${STAGING_BUCKET} --prefix="$(filter-out $@,$(MAKECMDGOALS))" --stage ${ARGS}; \
 		echo "Hosted at ${STAGING_URL}$(filter-out $@,$(MAKECMDGOALS))/${USER}/${GIT_BRANCH}/"; \
 	else \
 		mut-publish public ${STAGING_BUCKET} --prefix="${COMMIT_HASH}/${PROJECT}" --stage ${ARGS}; \
 		echo "Hosted at ${STAGING_URL}/${COMMIT_HASH}/${PROJECT}/${USER}/${GIT_BRANCH}/"; \
+      @echo "Hosted at ${STAGING_URL}/${COMMIT_HASH}/${PROJECT}/${USER}/${GIT_BRANCH}/cloud"; \
+	   @echo "Hosted at ${STAGING_URL}/${COMMIT_HASH}/${PROJECT}/${USER}/${GIT_BRANCH}/tools";
```
madelinezec, line 53: looks like this will give a syntax error

madelinezec: (line 53 not matched up to line 52, maybe extra tab?)

schmalliso: In fact, I was using spaces rather than tabs. The oldest problem in programming 😁 I've updated accordingly!

PR #316, opened 2020-09-18
fixing indexing for node, java

skerschb:
https://jira.mongodb.org/browse/DOP-1456

Found the same problem in the java makefile

### DIFF
--- a/makefiles/Makefile.docs-java
+++ b/makefiles/Makefile.docs-java
@@ -66,6 +66,6 @@ next-gen-deploy:
 
 deploy-search-index: ## Update the search index for this branch
 	@echo "Building search index"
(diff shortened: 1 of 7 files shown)


makefiles/Makefile.docs-java
```diff
@@ -66,6 +66,6 @@ next-gen-deploy:
 
 deploy-search-index: ## Update the search index for this branch
 	@echo "Building search index"
-	mut-index upload build/public -o ${PROJECT}-${GIT_BRANCH}.json -u ${PRODUCTION_URL}/${PROJECT} -g -s --exclude build/public/sdk/iOS
+	mut-index upload public -o ${PROJECT}-${GIT_BRANCH}.json -u ${PRODUCTION_URL}/${PROJECT} -s
```
i80and, line 69: Did you mean to remove `-g`?

I don't think new search even uses global search, so I don't think it actually matters.

skerschb: Well, might as well make it consistent. How is this?

madelinezec (approved):
lgtm

PR #322, opened 2020-09-30
mongocli configs

skerschb:
Per https://jira.mongodb.org/browse/DOP-1542

* fixed up makefile
* checked publishedbranches.yaml

### DIFF
--- a/makefiles/Makefile.docs-mongocli
+++ b/makefiles/Makefile.docs-mongocli
@@ -110,8 +110,8 @@ next-gen-html:
 	else \
 		exit 0; \
 	fi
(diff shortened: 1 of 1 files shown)


madelinezec (commented):
this looks good - have we tested it yet? we shouldnt need the line cont symbols. just want to make sure we test with them in if we are sure we want to include them.

madelinezec:
also re the jira ticket - is there supposed to be a publishedbranches.yaml file in this PR?

skerschb:
The publishedbranches.yaml is already in the repo.

I have had some wonky behaviors on ec2 with makefiles that do not have these appending slashes on large file copies.

PR #324, opened 2020-10-06
Fix unnamed security issue

skerschb:
No ticket, no title, top secret. Suffice it to say our branch names are vulnerable.


madelinezec (approved):
ty for linting - i wonder how the safeString(branchname) wasn't in there before?

PR #335, opened 2020-10-21
V0.4.7

madelinezec:


madelinezec:
spreadsheet of deploy tests (all passed): https://docs.google.com/spreadsheets/d/1ZpKmPOvMY6YV90TBKlcayZp71eoMZ2stfef-GBuzUNw/edit#gid=0

worker/jobTypes/S3Publish.js
```diff
@@ -80,9 +80,15 @@ class S3PublishClass {
     try {
       const exec = workerUtils.getExecPromise();
       const command = deployCommands.join(' && ');
-      const { stdout } = await exec(command);
+      const { stdout, stderr } = await exec(command);
       let stdoutMod = stdout;
-      
+
+      if (stderr && stderr.indexOf('ERROR') !== -1) {
```
i80and, line 86: Would it be possible to factor this out, since I'm seeing it a couple times?

i80and (approved):

worker/jobTypes/S3Publish.js
madelinezec: Hmm how so? This line tests for stderr from running makefile commands. We have two separate functions - one that calls the deploy target and one that calls the stage target (depending on the job type). Each function has to have this stanza

PR #341, opened 2020-11-03
Fix realm search index URL

i80and:
We're going to need to apply something similar to many other makefiles I believe, but we *need* to change how the prefix is passed in for this to not be really weird


madelinezec:
This looks good - before I approve and merge, did we test this?

madelinezec (approved):
lgtm

PR #348, opened 2020-11-25
Please review each record that pertains to a repo that you oversee

madelinezec:
We use an atlas collection to both populate the options in the dropdown menu of the /deploy app, and to correctly enter fields in the job document that the autobuilder picks up. **To support alias publishing, we have to update this collection. This PR contains all the changes to this collection.**

Eventually, a slack app will be created for adding/updating aliases. In the meantime, we are providing these docs so docs-leads can understand the data model for this collection (** eventually I will publish this info in a wiki, after Thanksgiving). 

Overall, the new data model looks like this:

```
repoName: string 
branches: array (of objects)
[
    {
      Name: string
      publishOriginalBranchName: boolean 
      active: boolean 
      aliases: null or array
    },
   Obj2, 
   Obj3 
]

```

**We need an entry for every repo that we publish on the autobuilder.** As this project has spanned many sprints, and we have been rapidly moving more projects onto the autobuilder, **I may have missed some repos. Please use this PR to correct my work.**

We have versioned and non-versioned repos, both of which can be aliased. 

To correctly publish each site (w/ or w/o aliases), we have to consider four things: 

1. Are we publishing the original branch name in addition to the branch's aliases? ie we are publishing docs.mongodb.com/charts/master along with /saas and /atlas
- For aliased repos, if you want to publish the orig branch name, you must have the "publishOriginalBranchName" field set to true
- Non-versioned and non-aliased repos (such as devhub, docs-realm, docs-ecosystem) must have the "publishOriginalBranchName" field set to false (we have never had /master in the url for these sites)

2. What (if any) are the aliases for this branch? You can also think of this question as, "When I deploy X github branch, what do I want the suffix of the url to be?"  For the master branch for docs-charts, the aliases are saas and atlas.
- For a branch with no aliases, the "aliases" field must be of type null 

3. Which is the primary alias? 
- The primary alias is the one which will trigger reindexing the repo for global search. Traditionally, we reindexed the stable branch (defined by the stableBranch field in the published branch files on docs-tools) for global search. Now that the stable branch may have multiple aliases, we put the primary alias as the first entry in the array of the aliases field.

4. Is every publishable branch represented? There should be an object for every publishable branch for a given repo document.

### DIFF
--- a/repos_branches.json
+++ b/repos_branches.json
@@ -259,7 +259,7 @@
       "publishOriginalBranchName": true,
       "active": true,
       "aliases": [
-        "v1.9",
+        "v1.10",
         "upcoming"
       ]
     },
--- a/repos_branches.json
+++ b/repos_branches.json
@@ -259,6 +259,59 @@
       "aliases": [
         "stable"
       ]
+        {
+      "name": "v1.8",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.7",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.6",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.5",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.4.0",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.3.0",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.2.0",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.1.0",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
+    },
+    {
+      "name": "v1.0.0",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": null
     }
   ]
 },{
--- a/repos_branches.json
+++ b/repos_branches.json
@@ -181,42 +181,6 @@
       "publishOriginalBranchName": true,
       "active": true,
       "aliases": null
-    },
-    {
-      "name": "v1.1",
-      "publishOriginalBranchName": true,
-      "active": true,
-      "aliases": null
-    },
-    {
-      "name": "v0.12",
-      "publishOriginalBranchName": true,
-      "active": true,
-      "aliases": null
-    },
-    {
-      "name": "v0.11",
-      "publishOriginalBranchName": true,
-      "active": true,
-      "aliases": null
-    },
-    {
-      "name": "v0.10",
-      "publishOriginalBranchName": true,
-      "active": true,
-      "aliases": null
-    },
-    {
-      "name": "v0.9",
-      "publishOriginalBranchName": true,
-      "active": true,
-      "aliases": null
-    },
-    {
-      "name": "v0.8",
-      "publishOriginalBranchName": true,
-      "active": true,
-      "aliases": null
     }
   ]
 },{
--- a/repos_branches.json
+++ b/repos_branches.json
@@ -144,7 +144,10 @@
       "name": "master",
       "publishOriginalBranchName": true,
       "active": true,
-      "aliases": null
+      "aliases": [
+        "v1.9",
+        "upcoming"
+      ]
     },
     {
       "name": "v1.8",
--- a/repos_branches.json
+++ b/repos_branches.json
@@ -147,6 +147,14 @@
       "aliases": null
     },
     {
+      "name": "v1.8",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": [
+        "current"
+      ]
+    },
+    {
       "name": "v1.7",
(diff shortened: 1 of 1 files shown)


jeff-allen-mongo (approved):

andf-mongodb (approved):
LGTM! And thanks for the incredibly-helpful description to this PR! Made understanding and reviewing an absolute breeze.

melissamahoney-mongodb (commented):
A few tweaks and some open questions.

1. There are a bunch of branches that we can be removed from the `docs-k8s-enterprise` drop-down. Should we mark them as `"active": false` or remove them entirely?

2. We are missing `mms-docs` from this list, which does stage using the autobuilder. I think the entry would look like what's below. My question here is, is there something special we need to do since this repo builds both CM and OM docs?
```
{
  "_id": {
    "$oid": "ADD OID HERE"
  },
  "repoName": "mms-docs",
  "branches": [
    {
      "name": "master",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": [
        "v5.0",
        "upcoming"
      ]
    },
    {
      "name": "v4.9",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": [
        "rapid"
      ]
    },
    {
      "name": "v4.4",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": [
        "current"
      ]
    },
    {
      "name": "v4.2",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v4.0",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    }
  ]
},
```

madelinezec:
> A few tweaks and some open questions.
> 
> 1. There are a bunch of branches that we can be removed from the `docs-k8s-enterprise` drop-down. Should we mark them as `"active": false` or remove them entirely?
> 2. We are missing `mms-docs` from this list, which does stage using the autobuilder. I think the entry would look like what's below. My question here is, is there something special we need to do since this repo builds both CM and OM docs?
> 
> ```
> {
>   "_id": {
>     "$oid": "ADD OID HERE"
>   },
>   "repoName": "mms-docs",
>   "branches": [
>     {
>       "name": "master",
>       "publishOriginalBranchName": true,
>       "active": true,
>       "aliases": [
>         "v5.0",
>         "upcoming"
>       ]
>     },
>     {
>       "name": "v4.9",
>       "publishOriginalBranchName": true,
>       "active": true,
>       "aliases": [
>         "rapid"
>       ]
>     },
>     {
>       "name": "v4.4",
>       "publishOriginalBranchName": true,
>       "active": true,
>       "aliases": [
>         "current"
>       ]
>     },
>     {
>       "name": "v4.2",
>       "publishOriginalBranchName": true,
>       "active": true,
>       "aliases": null
>     },
>     {
>       "name": "v4.0",
>       "publishOriginalBranchName": true,
>       "active": true,
>       "aliases": null
>     }
>   ]
> },
> ```

for transparency: @melissamahoney-mongodb and I touched base. since mms-docs is still only being published locally, it does not make sense to add it to this collection yet (which is for publishing purposes only). We will cross this bridge when we get closer to migrating mms-docs

repos_branches.json
```diff
@@ -0,0 +1,439 @@
...
+    {
+      "name": "v1.8",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": [
+        "stable"
+      ]
+    }
+  ]
+},{
```
melissamahoney-mongodb, line 264:
```suggestion
        {
      "name": "v1.7",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.6",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.5",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.4.0",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.3.0",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.2.0",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.1.0",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    },
    {
      "name": "v1.0.0",
      "publishOriginalBranchName": true,
      "active": true,
      "aliases": null
    }
  ]
},{
```

repos_branches.json
```diff
@@ -0,0 +1,439 @@
...
+    "$oid": "5f6aaed282989d521a60636b"
+  },
+  "repoName": "docs-mongocli",
+  "branches": [
+    {
+      "name": "master",
+      "publishOriginalBranchName": true,
+      "active": true,
+      "aliases": [
+        "v1.9",
```
melissamahoney-mongodb, line 251:
```suggestion
        "v1.10",
```

repos_branches.json
```diff
@@ -0,0 +1,439 @@
...
+        "upcoming"
+      ]
+    },
+    {
+      "name": "v1.8",
```
melissamahoney-mongodb, line 256:
```suggestion
      "name": "v1.9",
```

melissamahoney-mongodb:
@madelinezec Apologies that my comments aren't grouped, I didn't realize I hadn't selected "add a review" until I had already done them. :(

melissamahoney-mongodb (approved):

madelinezec:
We don't actually need to store this info on the nodeJS app (i just wanted to use github for ease of getting reviews!). As such, I am going to close without merging, but no need to be nervous about it :)

PR #349, opened 2020-11-25
fixing search requires updating all published branches yaml files

madelinezec:
Hello! 

Please double check that my updates to the stable branch fields are correct and that I have no introduced any new errors (and alert me if I have!) 

Again, front-end currently does not use the stable or upcoming field. The autobuilder uses the stable field, so we need stable field to exactly match the name of the github branch (ie we need to have it `v1.8` if the branch indeed has a `v`)

Thank you!

(and I know, ugly amount of commits, I always squash before merging)


schmalliso (approved):

andf-mongodb (approved):
LGTM! Thank you!

publishedbranches/docs-k8s-operator.yaml
```diff
@@ -25,8 +25,8 @@ version:
     - '1.2'
     - '1.1'
     - '1.0'
-  stable: '1.7'
-  upcoming: '1.8'
+  stable: 'v1.8'
+  upcoming: 'v1.8'
```
melissamahoney-mongodb, line 29:
```suggestion
  upcoming: 'v1.9'
```

publishedbranches/docs-mongocli.yaml
```diff
@@ -27,7 +27,7 @@ version:
     - '1.2'
     - '1.1'
     - '1.0'
-  stable: '1.8'
+  stable: 'v1.8'
   upcoming: '1.9'
```
melissamahoney-mongodb, line 31: Should this be:
```suggestion
  upcoming: 'v1.9'
```
?

melissamahoney-mongodb:
Also I think `mms-docs` should be in this list?

publishedbranches/docs-mongocli.yaml
melissamahoney-mongodb, line 31: I'm actually about to do a version bump right now and this will be:
```suggestion
  stable: 'v1.9'
  upcoming: '1.10'
```

madelinezec:
> Also I think `mms-docs` should be in this list?

Great catch! We don't need a published branches file for mms-docs until we have it publishing on next-gen.

jeff-allen-mongo (approved):

publishedbranches/docs-mongocli.yaml
```diff
@@ -27,8 +27,8 @@ version:
     - '1.2'
     - '1.1'
     - '1.0'
-  stable: '1.8'
-  upcoming: '1.9'
+  stable: 'v1.9'
+  upcoming: '1.10'
```
melissamahoney-mongodb, line 31: I know `upcoming` isn't a used field here, but maybe this should be `v1.10` for consistency?
```suggestion
  upcoming: 'v1.10'
```

melissamahoney-mongodb (approved):
One tiny nit then LGTM!

publishedbranches/docs-mongocli.yaml
madelinezec: I think that's smart! Thanks for catching it

PR #354, opened 2020-12-01
(DOCSP-13395): Version bump to v1.9

melissamahoney-mongodb:
[JIRA](https://jira.mongodb.org/browse/DOCSP-13395)

### DIFF
--- a/publishedbranches/docs-mongocli.yaml
+++ b/publishedbranches/docs-mongocli.yaml
@@ -1,7 +1,8 @@
 prefix: 'mongocli'
 version:
   published:
-    - '1.9'
+    - '1.10 (upcoming)'
+    - '1.9 (stable)'
     - '1.8'
     - '1.7'
     - '1.6'
@@ -17,7 +18,8 @@ version:
     - '0.0.4'
     - '0.0.3'
   active:
-    - '1.9'
+    - '1.10 (upcoming)'
+    - '1.9 (stable)'
     - '1.8'
     - '1.7'
     - '1.6'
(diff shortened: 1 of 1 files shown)


publishedbranches/docs-mongocli.yaml
```diff
@@ -27,13 +29,14 @@ version:
     - '1.2'
     - '1.1'
     - '1.0'
-  stable: '1.8'
-  upcoming: '1.9'
+  stable: 'v1.9'
+  upcoming: 'v1.10'
```
melissamahoney-mongodb, line 33: Should I just remove this line if snooty doesn't use it?

publishedbranches/docs-mongocli.yaml
```diff
@@ -27,13 +29,14 @@ version:
...
 git:
   branches:
-    manual: 'v1.9'
+    manual: 'v1.10'
```
melissamahoney-mongodb, line 36: Should I just remove this line if snooty doesn't use it?

publishedbranches/docs-mongocli.yaml
madelinezec: You could keep upcoming for now as Jon is considering if we want to display it in the future

publishedbranches/docs-mongocli.yaml
madelinezec: I think its fine to keep for now - we haven't rehauled these published branches files. So I think its ok to wait and just keep everything to be safe. I can always do clean up in future if we decide to formally get rid of a field

madelinezec (approved):

publishedbranches/docs-mongocli.yaml
melissamahoney-mongodb: Sounds good, thanks.

publishedbranches/docs-mongocli.yaml
melissamahoney-mongodb: Sounds good, thanks.

madelinezec:
@melissamahoney-mongodb for my own context - when do you next plan on deploying mongocli?

PR #360, opened 2020-12-08
DOP-1732: add published branches file for commandline tools

schmalliso:


publishedbranches/docs-commandline-tools.yaml
```diff
@@ -0,0 +1,15 @@
+prefix: '/'
```
madelinezec, line 1: I believe this should be 'tools'

madelinezec (changes requested):

publishedbranches/docs-commandline-tools.yaml
madelinezec: or 'database-tools' (depending on which site hosts the content of this repo) -- but certainly not '/' :)

schmalliso: oh lord, of course. cloud-docs's prefix is `/` because it's at its own weird url, right?

madelinezec (approved):

publishedbranches/docs-commandline-tools.yaml
madelinezec: yes exactly!!

PR #364, opened 2020-12-17
(DOCSP-13663): version bump mongocli

jwilliams-mongo:


madelinezec (approved):
I'm going to wait to merge this until we get the aliases sorted out. I'm checking with our product manager about how we want users to request or implement alias updates for the time being. I'll update you ASAP!

PR #366, opened 2021-01-14
fix prod bucket for commandline tools

schmalliso:


i80and (approved):
Our buckets are weird and need to be cleaned up, but I think this is right?

(We don't have to change the prefix, right?)

madelinezec (approved):

PR #368, opened 2021-01-21
Install mut from requirements.txt

i80and:


madelinezec (approved):
LGTM 

this will require an autobuilder release - so getting out these beautiful changes will take a little longer than just rebuilding

which is all good! but I may push to later in the afternoon

i80and:
There is no rush on this -- it's just to prevent problems in the unknowable future :D

PR #369, opened 2021-01-27
point docs-commandline-tools at new bucket for deploy

schmalliso:


madelinezec (approved):
LGTM 

does/did this require any changes to fastly stuff?

PR #372, opened 2021-01-28
(DOCSP-14178): version bump mongocli docs

jwilliams-mongo:


madelinezec (commented):
@jwilliams-mongo @melissamahoney-mongodb 

so the /v1.12 site and the /upcoming site will show content hosted on master branch?

and the /stable branch will show /v1.11

and v1.10 will have no aliases. it will only publish to /v1.10

jwilliams-mongo:
@madelinezec thanks for reviewing! I'm not sure that we need a /v1.12 alias yet. Not sure what the next release version will be. /upcoming should host master. All else looks good. @melissamahoney-mongodb can you confirm?

madelinezec (approved):

PR #376, opened 2021-02-17
Makefile side changes for DOP-1915, DOP-1858, DOP-1812 + Beginning shared makefile

madelinezec:
This PR contains the necessary makefile side changes to fix bugs associated with generating the search index manifest filename on the autobuilder. 

I also used this as an opportunity to start the work of the shared makefile, as this solution required adding an identical target to all of our makefiles (get-project-name). 

This sheet contains all the functional tests I did to verify the new work:
https://docs.google.com/spreadsheets/d/1R1Jqx1OOZX37JzTppouNMxwckJCpXtOAQXGScM09bwI/edit?usp=sharing

PR for the node side changes to come ASAP

### DIFF
--- a/makefiles/Makefile.devhub-content
+++ b/makefiles/Makefile.devhub-content
@@ -11,7 +11,6 @@ PRODUCTION_URL=https://developer.mongodb.com
 
 PROJECT=devhub
 MUT_PREFIX ?= $(PROJECT)
-PREFIX=devhub
 REPO_DIR=$(shell pwd)
 
 COMMIT_HASH=$(shell git rev-parse --short HEAD)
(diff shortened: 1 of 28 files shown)


makefiles/Makefile.devhub-content
```diff
@@ -11,7 +11,6 @@ PRODUCTION_URL=https://developer.mongodb.com
 
 PROJECT=devhub
 MUT_PREFIX ?= $(PROJECT)
-PREFIX=devhub
```
madelinezec, line 14: PREFIX is never referenced in this makefile (only MUT_PREFIX is) thus why I am removing it!

i80and (approved):

PR #377, opened 2021-02-17
Node side changes to implement DOP-1915, DOP-1858, DOP-1812 + Beginning shared makefile

madelinezec:


Dockerfile
```diff
@@ -33,6 +33,9 @@ USER docsworker-xlarge
 
 WORKDIR /home/docsworker-xlarge
 
+# get shared.mk
+RUN curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/makefiles/shared.mk >> shared.mk
```
i80and, line 37: A couple notes:
* `>>` works, but isn't necessary here: it'll *append* to the current file instead of overwriting it
* The `--output` option for curl is actually preferable over shell redirection: because file creation is handled by the shell, if curl fails, shell redirection will still end up creating an empty file.

worker/jobTypes/githubJob.js
```diff
@@ -40,7 +41,16 @@ class GitHubJobClass {
         }
         return false;
     }
-
+    async constructManifestIndexPath(){
+        try {
+            const snootyName = await utils.getSnootyProjectName(this.getRepoDirName());
+            this.currentJob.payload.manifestPrefix = snootyName + '-' + (this.currentJob.payload.alias ? this.currentJob.payload.alias : this.currentJob.payload.branchName)
+        } catch (error) {
+            console.log(error)
```
i80and, line 49: `console.error()` is preferable to `console.log()` for errors, unless it actually needs to be on stdout

worker/jobTypes/githubJob.js
```diff
@@ -312,14 +321,24 @@ class GitHubJobClass {
...
-          commandsToBuild[commandsToBuild.length - 1] = 'make get-build-dependencies';
-          commandsToBuild.push(`make next-gen-html`)
+        // we only generate a single search index per branch to ensure we do not have duplicate search indexes
+        // duplicate indexes === only differ by the suffix of the url, /atlas vs /saas vs /master
+        // if a branch is not aliased (and therefore not duplicated) or if this is the primary alias of a branch, construct a path for search index
+        if( ! this.currentJob.payload.aliased || ( this.currentJob.payload.aliased && this.currentJob.payload.primaryAlias ) ) {
+            await this.constructManifestIndexPath(); 
+        }
+        commandsToBuild[commandsToBuild.length - 1] = 'make get-build-dependencies';
+        commandsToBuild.push(`make next-gen-html`)
```
i80and, line 340: Don't need backtick quotes here

worker/utils/utils.js
```diff
@@ -3,7 +3,7 @@ const path = require('path');
 const fs = require('fs-extra');
 const request = require('request');
 const yaml = require('js-yaml');
-// const git  = require('nodegit');
+const toml = require('toml');
```
i80and, line 6: Is used anywhere?

worker/utils/utils.js
```diff
@@ -180,6 +180,28 @@ module.exports = {
     }
     return returnObject;
   },
+  
+  async getSnootyProjectName(repoDirName){
```
i80and, line 184: We can't actually just read snooty.toml for this, right? `get-project-name` just reads the makefile variable which has to match the snooty project name, right?

worker/jobTypes/githubJob.js
i80and: Actually, maybe `console.trace()` would be appropriate here?

worker/utils/utils.js
madelinezec: Yeah we can't read snooty.toml because at the point in the autobuilder process where we need to have the project name, we haven't yet downloaded the repo (which contains the snooty.toml file)

worker/jobTypes/githubJob.js
madelinezec: Hmm we use console.log error in all of our try/catch blocks (probably not a good thing?). I see one instance of console.error already in the repo, so I'll use console.error here I think.

Dockerfile
madelinezec: got it! I've changed it to curl <link> -o shared.mk

worker/jobTypes/githubJob.js
madelinezec: On second thought - I think we should use our logging class for this!

madelinezec: (not marking this as resolved just so you can give thumbs up or down)

i80and (approved):

PR #381, opened 2021-02-24
Correct mut-index invocation in vscode makefile

i80and:
Currently the wrong URL is passed to `mut-index`, so search results 404.


madelinezec (approved):
LGTM

PR #383, opened 2021-03-03
begin branch versioning for docs-node

terakilobyte:
It's finally happened, there's a new version of the node driver coming out.

### DIFF
--- a/publishedbranches/docs-node.yaml
+++ b/publishedbranches/docs-node.yaml
@@ -10,7 +10,7 @@ version:
   upcoming: 'v4.0'
 git:
   branches:
-    manual: 'v4.0'
+    manual: 'master'
     published:
       - 'master'
       - 'v3.6'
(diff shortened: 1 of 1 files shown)


publishedbranches/docs-node.yaml
```diff
@@ -1,16 +1,19 @@
...
-  stable: 'master'
-  upcoming: master
+    - '4.0 (upcoming)'
+    - '3.6 (stable)'
+  stable: 'v3.6'
+  upcoming: 'v4.0'
 git:
   branches:
-    manual: 'master'
+    manual: 'v4.0'
```
sophstad, line 13: This field should correspond to the branch name, which I presume to be `master` from line 15!
```suggestion
    manual: 'master'
```

sophstad (changes requested):

sophstad (approved):

sophstad:
@terakilobyte When are the v4 docs set to be published? I want to make sure we schedule the alias branch work appropriately!

terakilobyte:
@sophstad Right now we think they'll go GA by the 20th of March, though that could slip. If it'd help we can cut a placeholder v4.0 branch now?

schmalliso (approved):

PR #386, opened 2021-03-09
DOP-2025: Add Go docs build pipeline

sophstad:
[[DOP-2025](https://jira.mongodb.org/browse/DOP-2025)]
- Do not include commit hashes in staging jobs
- The [mongodb/docs-golang](https://github.com/mongodb/docs-golang) repo will be using `main` as its primary branch name instead of `master`; I tried to reflect that here but not sure if there is anywhere else where that would have an impact
- It looks like every repo has its own integration bucket, so perhaps we need to make one?

### DIFF
--- a/makefiles/Makefile.docs-golang
+++ b/makefiles/Makefile.docs-golang
@@ -12,8 +12,6 @@ else
   PRODUCTION_BUCKET=docs-go-integration
 endif
 
-COMMIT_HASH=$(shell git rev-parse --short HEAD)
-
 PREFIX=/drivers/go
 PROJECT=golang
 MUT_PREFIX ?= $(PROJECT)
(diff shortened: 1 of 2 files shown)


madelinezec:
Oh hmm I think I'd like to investiage if the master -> main change will work or if it'll require changes on the autobuilder. How soon do they need this by?

sophstad:
@terakilobyte said in the ticket

> We'd like to start work on this next week.

madelinezec:
Oh goodness ok - well, I can definitely get this out by Thursday. Because this slightly differs from the git process we expect, I'd have to figure out what can work (they definitely won't have access to pushless staging for example).

It probably would have been good to let them know last week in the ticket that this might not be able to turn right around given the master -> main change (but also I was out, and I understand that this would not occur to everyone!)

makefiles/Makefile.docs-golang
```diff
@@ -0,0 +1,55 @@
...
+	echo "GATSBY_SITE=${PROJECT}" >> .env.production; \
+	npm run build; \
+	cp -r "${REPO_DIR}/snooty/public" ${REPO_DIR};
+
+next-gen-stage: ## Host online for review
+	mut-publish public ${STAGING_BUCKET} --prefix="${MUT_PREFIX}" --stage ${ARGS};
+	echo "Hosted at ${STAGING_URL}/${MUT_PREFIX}/${USER}/${GIT_BRANCH}/";
+
+next-gen-deploy:	
+	if [ ${GIT_BRANCH} = main ]; then mut-redirects config/redirects -o public/.htaccess; fi
```
madelinezec, line 48: @terakilobyte has updated it to master to get this out asap - so let's make that change reflected here please :)

madelinezec (changes requested):
just one small change and we're good to go

madelinezec (approved):

makefiles/Makefile.docs-golang
sophstad: Oh, I think perhaps the published branches file should have been updated as well

madelinezec: Oh my - I'll get that out one sec

madelinezec: we're good to go https://github.com/mongodb/docs-worker-pool/blob/meta/publishedbranches/docs-golang.yaml

PR #391, opened 2021-03-16
DOP-2043 allow multiple autobuilder staging instances

madelinezec:
In the current iteration of the docs-worker-pool, every instance of the autobuilder is spun up to listen to updates to the same, single collection. This is great for production - we have four autobuilder instances listening to the same collection, thus building sites faster and more robustly. However, it represents a huge bottleneck for staging. 

If multiple staging instances of the autobuilder are up, each instance will pick up a job whenever it becomes available despite the fact that a job may have been intended to run on a different staging instance that has different tests implemented.  This is why I've always had to ask when the team wanted to test on integration - because I needed my tests jobs to be run through my staging instance (especially if I was testing publish/deploys!). 

This PR reflects the node-js side changes required to enable multiple autobuilder staging instances running at the same time, without picking up each other's jobs. 

The other work required will be: 
1. Create a collection in the testing database that reflects your name. 
2. Create a webhook in the staging realm app for the autobuilder that adds jobs to your personal collection. 
3. Add a monitor to your personal collection so that slack is triggered to send you a message on any update to a doc in your personal collection (I have yet to do this step for myself - we should probably create a ticket for this, as the queue dashboard right now only works for the single staging collection and i'm not sure how it would handle multiple / what changes would be required)

### DIFF
--- a/worker/utils/environment.js
+++ b/worker/utils/environment.js
@@ -5,6 +5,7 @@ const atlasUsername = process.env.MONGO_ATLAS_USERNAME;
 const atlasPassword = process.env.MONGO_ATLAS_PASSWORD;
 const xlarge = process.env.XLARGE;
 const jobDb = process.env.DB_NAME;
+const jobCol = process.env.COL_NAME;
 
 class EnvironmentClass {
   static getDB() {
@@ -12,6 +13,15 @@ class EnvironmentClass {
       return 'pool_test';
     }
     return jobDb;
+  }
+
+  /* collection name is dynamic to allow mult staging autobuilder instances running
+     at same time without picking up each other's jobs */
+  static getCollection() {
+    if (jobCol === undefined) {
+      return 'queue';
+    }
+    return jobCol;
   }
 
   static getXlarge() {
--- a/worker/utils/mongo.js
+++ b/worker/utils/mongo.js
@@ -9,8 +9,8 @@ const runXlarge = EnvironmentClass.getXlarge();
 const url = `mongodb+srv://${username}:${password}@cluster0-ylwlz.mongodb.net/admin?retryWrites=true`;
 
 // Collection information
-const DB_NAME = EnvironmentClass.getDB(); // Database name of the queue in MongoDB Atlas
-const COLL_NAME = 'queue'; // Collection name of the queue in MongoDB Atlas
+const DB_NAME = EnvironmentClass.getDB(); // Database name for autobuilder in MongoDB Atlas
+const COLL_NAME = EnvironmentClass.getCollection(); // Collection name in MongoDB Atlas
 const META_NAME = 'meta';
 const MONITOR_NAME = 'monitor';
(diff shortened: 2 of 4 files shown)


worker/utils/mongo.js
```diff
@@ -91,9 +91,14 @@ module.exports = {
     };
 
     const update = { $set: { startTime: new Date(), status: 'inProgress' } };
-    const options = { sort: { priority: -1, createdTime: 1 } };
-    await queueCollection.findOne({ status: 'inQueue' });
```
GuruPKK, line 95: Curious to know why this has been removed?

GuruPKK (approved):

worker/utils/environment.js
```diff
@@ -14,6 +15,15 @@ class EnvironmentClass {
     return jobDb;
   }
 
+  /* collection name is dynamic to allow mult staging autobuilder instances running
+     at same time without picking up each other's jobs */
+  static getCollection() {
```
i80and, line 20: Does this need to be a static function, or could it just be a constant? It doesn't seem to be overridden AFAICS, and it looks like it will always return the same thing?

worker/utils/mongo.js
```diff
@@ -91,9 +91,14 @@ module.exports = {
...
-    return queueCollection.findOneAndUpdate(query, update, options);
+    const options = { sort: { priority: -1, createdTime: 1 }, returnNewDocument: true };
+
+    try {
+      return queueCollection.findOneAndUpdate(query, update, options);
+    } catch (error) {
+      console.log(error)
```
i80and, line 99: `console.trace()` could be more helpful?

i80and (dismissed):

worker/utils/mongo.js
madelinezec: options is still in the code base ([look at this link](https://github.com/mongodb/docs-worker-pool/pull/391/files#diff-4d2a99f21e295622e09d0249fd63dcf7d8c620ce42ebd2528c7d6384523218acR94)

madelinezec: And findOne is removed because we also already have findOneAndUpdate - so the extra findOne seemed redundant to me! :)

worker/utils/environment.js
madelinezec: If we forget to update the env vars in the script on one of our production sites for instance (such that jobCol === undefined), without this static function, the autobuilder will fail and there will be an outage. I think its smart to keep it as a static function in the interest of safety check

i80and (approved):

GuruPKK (approved):

worker/utils/environment.js
madelinezec: For visibility/longevity for anyone in the future who may spy on this PR: I misunderstood Andrew's point, however it was agreed that it made more sense to keep with the established style of the codebase (even if its a java-ism) than to change how we retrieve one var (`const collection = (jobCol === undefined?) "queue" : jobCol;`).

PR #392, opened 2021-03-22
DOP-1323 Remove triple process on build failures

GuruPKK:
Reduced the number of retries from 2 to 1. 
Failed builds will be retried only once instead of twice. 
Tested with random repo's, 
simulated failure by switching Credentials so that publishing fails 
Test results are https://docs.google.com/spreadsheets/d/1n3U_Ko42uvElB4bJfUDR3QJawcgBn-RF-agv4tVbKkU/edit#gid=0


madelinezec (changes requested):
Hey Guru this is a great start! However, we want failed builds to process only once. That means after it fails the first time, it is not retried. By looking at[ your tests](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?jobId=605911fa2723a9e448b0cc6a), I can see each jobs was attempted twice before the autobuilder considered it completed/finished. 

We want it to fail once and only once 😀

GuruPKK:
> Hey Guru this is a great start! However, we want failed builds to process only once. That means after it fails the first time, it is not retried. By looking at[ your tests](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?jobId=605911fa2723a9e448b0cc6a), I can see each jobs was attempted twice before the autobuilder considered it completed/finished.
> 
> We want it to fail once and only once 😀

Oh Ok I got it, So we are removing retry altogether. if it is failed once we are marking it as failed instead of retrying..  I got confused will update

madelinezec (commented):
This is looking great! Thanks for updating the spreadsheet with you new tests

Before this PR can get approved, let's also fix up the code on the realm app. We definitely want to change this logic now that there will be no retries: https://github.com/mongodb/snooty-workerpool-stitchapp/blob/master/functions/handleSlackTrigger/source.js#L70-L92

We also want to change the structure of the job document itself. Now that `failures` will no longer be plural, we should change the field to `error` and have it be an object (rather than an array of objects as failures currently it) 

We should also change `logs` so that its an array and make it so that there are no try objects (ie `try0`) within the `logs`: https://github.com/mongodb/docs-worker-pool/blob/master/worker/utils/mongo.js#L140-L156

madelinezec (commented):
This looks great! One note: lets fix the now outdated comment above this function ("  //Sends Job To inQueue Status or Failed Status Depending on job.numFailures")

Let's also remove the numFailures - as if it fails now, we know it fails once and only once :)

madelinezec (approved):
LGTM! Lets merge this and then re-test once integration is spun up (we test **a lot** :D )

PR #396, opened 2021-03-30
DOP-1974: update docs published branches for snooty deploy

schmalliso:


madelinezec (approved):
holy moly

schmalliso:
@madelinezec We're only _actually_ publishing master today, so it's not as much as it seems :)

PR #402, opened 2021-04-16
Continuous Deployment via Github Actions

casthewiz:
### What this does:
Adds a series of github actions to automatically deploy to dedicated EC2 instances for respective environments. 

All hosts, credentials, etc. are protected in Github Secrets. If a secret value is not present at runtime of the action, the deploy will fail without running any commands on the remote instance.

### Before merging
Reconciliation of `Dockerfile` `git diff` for each respective instance may be required for these actions to successfully work.

### Other notes
We should really get a dedicated key pair for this action to use when SSH'ing in the future. 

Hopefully this can all go away and get replaced with building the docker images and sending them ECR, then having a simple action either in this repo or in CodePipeline to handle any container up/down required.


.github/workflows/deploy-staging-ec2.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install SSH Key
+        uses: shimataro/ssh-key-action@v2.3.0
+        with:
+          key: ${{ secrets.EC2_RSA_KEY }}
+          known_hosts: ${{ secrets.EC2_KNOWN_HOSTS }}
+          name: id_rsa
+      - name: Deploy Development to Staging
+        run: |
+          ssh -l ubuntu ${{ secrets.STAGING_HOST}} -T '
+          cd docker/cs-stage-docs-worker-pool
```
casthewiz, line 21:
```suggestion
          cd docker/docs-worker-pool
```
CHANGE THIS BEFORE MERGING

Dockerfile
```diff
@@ -42,14 +42,14 @@ RUN python3 -m pip install pip==20.2 flit==3.0.0
...
-	git checkout <TAG> && \
+	git checkout v0.9.6 && \
 	FLIT_ROOT_INSTALL=1 python3 -m flit install
 
 # install snooty front-end
 RUN git clone https://github.com/mongodb/snooty.git snooty
 RUN cd snooty && \
 	git fetch --all && \
-	git checkout <TAG> && \	
+	git checkout v0.9.8 && \
```
casthewiz, line 52: With these actions in place, we can deploy via bumping the tags in the Dockerfile via a pull request.

.github/workflows/deploy-production-ec2.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: SSH to Production 1
+        run: |
+          ssh ${{ secrets.PRODUCTION_HOST_1}} -T '
+          cd docker/docs-worker-pool
+          git pull origin
+          git checkout $(git describe --tags)
+          sudo docker system prune
+          sudo docker build --no-cache --tag production .
+          sudo docker stop $(sudo docker ps -a -q)
+          sudo sh rundocker.sh
```
casthewiz, line 28: in a perfect world, we can run `sudo docker build --no-cache --tag production . ` in this action and then send the image to ECR - and then fetch the new image from ECR and do any stopping, pruning, etc. on the remote **instead** of building the new docker image on the remote instance

for now, this works though.

.github/workflows/deploy-staging-ec2.yml
```diff
@@ -0,0 +1,34 @@
+on:
+  push:
+    branches:
+      - development
```
casthewiz, line 4: The trigger for this deployment to staging is very open to debate - I've used a `development` branch that's essentially a feature branch of of `master` to trigger deploys to staging envs in the past, but in a perfect world we might want something that can deploy an ephemeral instance for each pull request in this repository.

Right now this seemed like the best setup given our current infrastructure.

Dockerfile
madelinezec: I just want @sophstad and @i80and to see this and confirm we're comfortable with this change :)

.github/workflows/deploy-integration-ec2.yml
```diff
@@ -0,0 +1,31 @@
+on:
+  push:
+    branches:
+      - master
+name: Deploy Integration
+jobs:
+  deploy-integration:
+    if: github.ref == 'refs/heads/master'
```
madelinezec, line 8: If I'm understanding the logic here (which I could entirely not be), we're going to deploy integration on merges to master? But also, wouldn't merges to master deploy to production as well?

It would be great to deploy integration on merges to stage (or create a new branch, integration). And deploy to production on mergers to master.

madelinezec: (completely tangential, we really should move off of master soon bc that terminology is horrible)

.github/workflows/deploy-staging-ec2.yml
```diff
@@ -0,0 +1,34 @@
...
+name: Deploy Staging
```
madelinezec, line 5: I don't think we would ever want to deploy to staging environment. Staging is for live work / live debugging. We use it to run/test the autobuilder while we are still implementing a fix/feature bc it can be faster than running the autobuilder on our local machines. Having a merge to deploy on staging doesn't make sense for our workflow

madelinezec (changes requested):

.github/workflows/deploy-integration-ec2.yml
casthewiz: Sure! so there's two logic blocks in each workflow file - lines 1-4 which trigger the action, and then the block here on line 7 which is defining the job `deploy-integration`. 
```
on:
  push:
    branches:
      - master
name: Deploy Integration
jobs:
  deploy-integration:
    if: github.ref == 'refs/heads/master'
```
the if statement here is admittedly a little redundant, but serves as another check to make sure that we run this action only if the [context](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context) in which this action was run is indeed the head of `master`. 
Practically, it's just another guard to confirm that we're only running our job in the contexts we deem appropriate. 


as far as the production deploy goes
```
on:
  push:
    tags:
      - 'v*'
name: Deploy Production
jobs:
  deploy-production-1:
    if: github.ref == 'refs/heads/master'
```
this is a good catch because this logic block will always eval to false - the `github.ref` should equal something in the format of `refs/tags/<tag number>`. The upside here is that we just wouldn't run the production release jobs - the action would eval the if as false and then skip running the job.

casthewiz: > we're going to deploy integration on merges to master? But also, wouldn't merges to master deploy to production as well? It would be great to deploy integration on merges to stage (or create a new branch, integration). And deploy to production on mergers to master.

I think my intent is to try to mirror our current branching strategy -> deployment process as much as possible. I'm under the current impression that the workflow is to test `master` on integration and create release tags for production (and deploy those tags to production) if everything looks good in integration. If that's not the case then I'd want to revisit the triggers for these jobs - which wouldn't be hard to do at all.

madelinezec: Let's revisit yes! I became the only engineer on the server back in October - so to create less PRs for non-autobuilder folks to review, I would just PR against master. Rather than bug them to review a PR against staging branch, test staging branch and then have to bug them against to review essentially same/similar PR against master.

However, the ideal/original workflow was to open a PR against staging and then test staging on integration environment. With multiple people working on the autobuilder, and thus more engineering resources for reviewing, I think it makes most sense to return to our original workflow.

madelinezec: (also to make the terminology more clear, we could create a branch Integration and associate that with testing on the integration environment, rather than using stage as we did previously)

casthewiz: That sound perfect - we're actually really close to https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow then. I'll make some changes to the triggers.

.github/workflows/deploy-integration-ec2.yml
```diff
@@ -0,0 +1,31 @@
+on:
+  push:
+    branches:
+      - master
```
casthewiz, line 4:
```suggestion
      - integration
```

.github/workflows/deploy-integration-ec2.yml
casthewiz, line 8:
```suggestion
    if: github.ref == 'refs/heads/integration'
```

.github/workflows/deploy-production-ec2.yml
```diff
@@ -0,0 +1,53 @@
+on:
+  push:
+    tags:
+      - 'v*'
+name: Deploy Production
+jobs:
+  deploy-production-1:
+    if: github.ref == 'refs/heads/master'
```
casthewiz, line 8:
```suggestion
    if: startsWith(github.ref, 'refs/tags/v')
```

.github/workflows/deploy-production-ec2.yml
```diff
@@ -0,0 +1,53 @@
...
+          cd docker/docs-worker-pool
+          git pull origin
+          git checkout $(git describe --tags)
+          sudo docker system prune
+          sudo docker build --no-cache --tag production .
+          sudo docker stop $(sudo docker ps -a -q)
+          sudo sh rundocker.sh
+          '
+  deploy-production-2:
+    if: github.ref == 'refs/heads/master'
```
casthewiz, line 31:
```suggestion
    if: startsWith(github.ref, 'refs/tags/v')
```

.github/workflows/deploy-integration-ec2.yml
casthewiz: @madelinezec I think I'll keep the trigger for production deploys as push of new release tags - this lets us keep the intentionality we have with manual deploys to production.

madelinezec (approved):
LGTM! Excited to test this out next week ahead of our release (and hopefully employ it for the release on thursday!!!!)

Let's remember to create a ticket or two to document the new process / steps for releasing the toolchain. I'll then edit the current wiki to say "manual release process" etc.

casthewiz:
I **normally** wouldn't do this, but commits post approval were to beef up the README.md for incoming changes and resolve a handful of minor discrepancies in action verbiage vs. what was actually executing - changing names to `Integration` from `Master` in action titles for integration, for example.

Also changed filepath in integration to accurately reflect file structure of that instance after doublechecking.

Merging this changeset post approval so that our actions are present within the repository for testing purposes.

PR #405, opened 2021-04-21
DOP-1311 Fix Path Overwrite during Regression Testing

GuruPKK:


madelinezec (changes requested):
this looks good so far but we also want to remove the lines where we currently set STAGING_URL and STAGING_BUCKET

madelinezec (commented):
this looks really good! let's poll the team to make sure we're good with bucket name and url

PR #406, opened 2021-04-22
DOP-2082 - Manual staging links should not contain the commit hash

GuruPKK:


madelinezec (commented):
on initial pass, this looks good but can you stage a job just to test on integration or staging ec2 before we merge? thank you!

PR #407, opened 2021-04-22
remove legacy targets

madelinezec:
hoping to clean up server manual's makefile! Am I good to remove all these legacy targets?


schmalliso (changes requested):
No! There's a bunch of branches that are still on legacy that haven't moved over yet.

madelinezec:
@schmalliso thank you for the quick response  🙌!!! I did not know realize they publish legacy branches of the server manual on the autobuilder!

PR #408, opened 2021-04-23
DEVHUB-634: Webhooks for Devhub Production/Staging

jestapinski:
Hi team!

This PR adds two webhooks into the Workerpool Realm App for DevHub to use with our new CMS on deploys to staging and production. We have one in there for Staging but I thought I would take this chance to clean that up while adding a similar one for production jobs.

We will handle calling these, I'm just hoping that these make sense and maybe there aren't any potential updates we could make? Any thoughts would be great! I'm also just going a bit blind here.

Thanks a ton!

### DIFF
--- a/services/strapi-build-webhook/incoming_webhooks/strapi-staging-webhook/source.js
+++ b/services/strapi-build-webhook/incoming_webhooks/strapi-staging-webhook/source.js
@@ -1,0 +1,34 @@
+// This function is the webhook's request handler.
+exports = async function (payload, response) {
+  const { email, name } = await JSON.parse(await payload.body.text());
+
+  try {
+    let jobTitle = "DevHub CMS Staging Build";
+    let jobUserName = name;
+    let jobUserEmail = email;
+    const newPayload = {
+      jobType: "githubPush",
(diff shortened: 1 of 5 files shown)


services/strapi-build-webhook/incoming_webhooks/strapi-staging-webhook/source.js
```diff
@@ -0,0 +1,34 @@
...
+    const newPayload = {
+      jobType: "githubPush",
+      source: "strapi",
+      action: "push",
+      repoName: "devhub-content-integration",
+      branchName: "test-strapi-webhook",
+      isFork: true,
+      private: true,
+      isXlarge: true,
+      // Can this be generalized to use the devhub-content master branch instead of my fork?
```
jestapinski, line 18: Curious if there are any options for this comment on line 18?

jestapinski:
Hey @madelinezec and @GuruPKK I don't seem to have access to assign reviewers, so if you have some time I would love a look. Thanks!

madelinezec:
Alrighty! I will take a look on Monday! Have a good weekend!

PR #412, opened 2021-04-28
DEVHUB-636: Consume STRAPI_PUBLICATION_STATE

jestapinski:
This PR attempts (keyword!) to use the `STRAPI_PUBLICATION_STATE` variable defined in [this branch](https://github.com/mongodb/docs-worker-pool/compare/master...madelinezec:DOP-2144?expand=1) and pipe it to the `env.production` file. I also opted to wrap in a conditional to only add the field should it exist.

This _should_ do the trick?

### DIFF
--- a/makefiles/Makefile.devhub-content
+++ b/makefiles/Makefile.devhub-content
@@ -5,6 +5,7 @@ STAGING_URL="http://docs-devhub-staging.s3-website-us-east-1.amazonaws.com"
 STAGING_BUCKET=docs-devhub-staging
 
 STRAPI_URL=http://54.219.137.111:1337
+STRAPI_PUBLICATION_STATE = $(shell printenv STRAPI_PUBLICATION_STATE)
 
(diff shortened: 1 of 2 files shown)


makefiles/Makefile.devhub-content
```diff
@@ -5,6 +5,7 @@ STAGING_URL="http://docs-devhub-staging.s3-website-us-east-1.amazonaws.com"
 STAGING_BUCKET=docs-devhub-staging
 
 STRAPI_URL=http://54.219.137.111:1337
+STRAPI_PUBLICATION_STATE = $(shell printenv STRAPI_PUBLICATION_STATE)
```
madelinezec, line 8: I don't believe we need this line actually!

madelinezec (changes requested):

makefiles/Makefile.devhub-content
jestapinski: TIL!

PR #422, opened 2021-05-11
Dop 1986: Instantly see changes on docs on /deploy

madelinezec:
background on the issue: https://docs.google.com/document/d/1n_KuSuz_4RbCtYR45WT-ormZo5X1RA_r80ioVmfVZFQ/edit

context on implementation: https://docs.google.com/document/d/1IxnFhyTcDVRLBECH6EXQIrc9ksFHTYrFe3biO2ELAQY/edit?usp=sharing


worker/utils/fastlyJob.js
```diff
@@ -21,51 +45,71 @@ class FastlyJobClass {
...
     let urlCounter = urlArray.length;
     let purgeMessages = [];
+    const fastly_service_id = environment.getFastlyServiceId();
+  
+      for (let i = 0; i < urlArray.length; i++) {
+        //retrieve surrogate key
+        try {
+          const surrogateKey = await this.retrieveSurrogateKey(urlArray[i])
+          // perform request to purge
+          await this.requestPurgeOfSurrogateKey(surrogateKey)
```
madelinezec, line 55: awaiting insde a for loop breaks the no-await-in-loop ESLint rule. If we should indeed honor no-await-in-loop, would love your advice on how to refactor it

worker/utils/fastlyJob.js
```diff
@@ -1,92 +1,128 @@
...
-    const headers = {
-      'fastly-key': environment.getFastlyToken(),
-      'accept': 'application/json',
-      'Fastly-Soft-Purge': '1',
-    };
+                  // warm cache
+                  await this.warmCache(urlArray[i]);
+              } catch (error) {
+                  console.trace(error);
+                  // should return reject here?
```
madelinezec, line 41: same issue here, do we need to be issuing rejects where we issue resolves? ie do we need to have a condition for rejecting? or is catching and throwing an error enough?

worker/jobTypes/S3Publish.js
```diff
@@ -93,57 +94,37 @@ class S3PublishClass {
...
-      } catch (e) {
-        // if not JSON, then it's a normal string output from mut
-        // get only last part of message which includes # of files changes + s3 link
-        if (stdout.indexOf('Summary') !== -1) {
-          stdoutMod = stdout.substr(stdout.indexOf('Summary'));
-        }
-      }
 
-      return new Promise((resolve) => {
-        logger.save(`${'(prod)'.padEnd(15)}Finished pushing to production`);
```
madelinezec, line 131: 1. Do we want to record URLs that have been purged? Ofc I'm assuming yes
2. How do we want to log these URLs to user? Do we want to have a field in the job object, purgedUrls? I don't think it makes sense to add it to the already long `logs` array field that users have to scroll thru to find info that is relevant to their needs
3. Do we want to slack them when URLs have been purged? Also for context, we are now defining a job as finishing after it has pushed to S3 AND purged URLs associated with updated content and warmed the cache (GET'd the purged URLs)

worker/utils/fastlyJob.js
casthewiz: [`Promise.All`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) in general would be the way to batch out async executions for an iterable 👍

casthewiz: ^^ That isn't necessarily compile + run ready, but would be a good example of how we could do this. 

Batch promise execution in stages like this can be a bit messy - it might make more sense to make a single function that cascades the `retrieveSurrogateKey()` -> `requestPurgeOfSurrogateKey` -> `warmCache` sequence of execution with appropriate awaits, and then batch execution of that function with a `Promise.All()`.

GuruPKK (approved):

PR #433, opened 2021-06-03
Meta

GuruPKK:


casthewiz:
With moving the `next-gen-deploy` out of the Makefiles, do we have to also set `CUSTOM_NEXT_GEN_DEPLOY=true` for all of them, or only for `docs-404?`

makefiles/shared.mk
```diff
@@ -7,8 +8,15 @@ endif
 get-project-name:
 	@echo ${PROJECT};
 
+ifndef CUSTOM_NEXT_GEN_DEPLOY
+next-gen-deploy:
+	if [ -f config/redirects -a "${GIT_BRANCH}" = master ]; then mut-redirects config/redirects -o public/.htaccess; fi	
+	yes | mut-publish public ${PRODUCTION_BUCKET} --prefix="${MUT_PREFIX}" --deploy --deployed-url-prefix=${PRODUCTION_URL} --json --all-subdirectories ${ARGS};
+	@echo "Hosted at ${PRODUCTION_URL}/${MUT_PREFIX}";
+	if [ ${MANIFEST_PREFIX} ]; then $(MAKE) next-gen-deploy-search-index; fi
+endif
```
casthewiz, line 17: If this is only running next-gen-deploy if `CUSTOM_NEXT_GEN_DEPLOY`, do we need to specify and set `CUSTOM_NEXT_GEN_DEPLOY` to true in the Makefiles where next-gen-deploy was removed?

GuruPKK: If you look next-gen-deploy is available only for repos where CUSTOM_NEXT_GEN_DEPLOY is not defined. for docs-404 we are defining it so the local deploy is called.

PR #400, opened 2021-04-09
DEVHUB-542, DOP-2093: Push DevHub Deploys to an Additional S3 Bucket

jestapinski:
[JIRA (DOP)](https://jira.mongodb.org/browse/DOP-2093)
[JIRA (DevHub)](https://jira.mongodb.org/browse/DEVHUB-542)
[Project Tech Spec](https://docs.google.com/document/d/11zgUY7l6oprkcfQU0Nex4nOha9Ztm_tgqOlaLYux0yM/edit#heading=h.bn5fffrhnzxo) (look for Step 1 - Create parallel publishing workflow to a new S3 bucket)

Hi team! For our subdomain consolidation, we need to push content to another S3 bucket in addition to the existing one. This (should) take care of it, but not totally sure, hence the WIP. Happy to chat as needed if something else is required.

Thanks!
Jordan

PR #434, opened 2021-06-04
DEVHUB-542, DOP-2093: Additional DevHub Deploy Job

jestapinski:
(Continuation of https://github.com/mongodb/docs-worker-pool/pull/400)
[JIRA (DOP)](https://jira.mongodb.org/browse/DOP-2093)
[JIRA (DevHub)](https://jira.mongodb.org/browse/DEVHUB-542)
[Project Tech Spec (look for Step 1 - Create parallel publishing workflow to a new S3 bucket)](https://docs.google.com/document/d/11zgUY7l6oprkcfQU0Nex4nOha9Ztm_tgqOlaLYux0yM/edit#heading=h.bn5fffrhnzxo)

After going back and forth on requirements, we need a new job to build the DevHub to prepare for our subdomain consolidation. We need a new job because:
- We need Gatsby to run with a new config file to change our prefix path ([see our front-end PR](https://github.com/mongodb/devhub/pull/856))
- We need to specify a new path for the deploy to go to in S3

Still awaiting getting the 10gen-noc canonical ID for s3 access to the bucket but wanted to get this out there. Ideally, we run this job in parallel or right after the existing devhub content production deploy job.


jestapinski:
(Seems `meta` is being worked on but this is a net new job so just going off that for now)

makefiles/Makefile.devhub-content-new
```diff
@@ -0,0 +1,59 @@
...
+	# snooty parse and then build-front-end
+	@echo ${SNOOTY_DB_PWD} | snooty build "${REPO_DIR}" "mongodb+srv://${SNOOTY_DB_USR}:@cluster0-ylwlz.mongodb.net/snooty?retryWrites=true" --commit "${COMMIT_HASH}"; \
+	if [ $$? -eq 1 ]; then \
+		exit 1; \
+	else \
+		exit 0; \
+	fi
+	rsync -az --exclude '.git' ${REPO_DIR}/../../snooty-devhub ${REPO_DIR}; 
+	cp ${REPO_DIR}/.env.production ${REPO_DIR}/snooty-devhub; 
+	curl https://raw.githubusercontent.com/mongodb/devhub/master/gatsby-config.prod-new.js > ${REPO_DIR}/snooty/gatsby-config.js
```
jestapinski, line 44: This is important, as in this new config file we set the path prefix we will need for prod. We may also need to tweak some other settings.

casthewiz, line 44:
```suggestion
	curl https://raw.githubusercontent.com/mongodb/devhub/master/gatsby-config.prod-new.js > ${REPO_DIR}/snooty-devhub/gatsby-config.js
```

casthewiz: We'll need to make sure it curls to the `snooty-devhub` dir for this approach, but I have another concern in that these job runs aren't atomic - after this gets executed, we'd overwrite the `gatsby-config.js` for every following job, until the docker container gets rebuilt.

@GuruPKK correct me if I'm wrong 🤔

jestapinski: Oh interesting, curious what we do now to distinguish between gatsby-config and gatsby-config.prod?

casthewiz: > Oh interesting, curious what we do now to distinguish between gatsby-config and gatsby-config.prod?

We could be extra safe here by changing line 42 to 
```
rsync -az --exclude '.git' ${REPO_DIR}/../../snooty-devhub/ ${REPO_DIR}/snooty-devhub-tmp;
```
and then changing further references to `snooty-devhub` after line 42  to `snooty-devhub-tmp`, so we're extra confident that everything is happening in a new directory, and only this job run is occurring in that directory.

GuruPKK: @jestapinski cassidy's suggestion needs to be applied otherwise we will be overwriting the config

GuruPKK (changes requested):

makefiles/Makefile.devhub-content-new
casthewiz: Yeah, in general I think we'd be much more confident in this if we rsync to a directory specific to this job run.

This would also help us during any debugging of a job run that needs to happen - and would ensure that we don't run into issues if this makefile runs in parallel with the base devhub job run.

GuruPKK: Yeah it will be better to keep it in different dir, and curl to that specific dir. this will have the changes local to devhub and doesnt impact other builds if we turn on multi processing in auto builder

jestapinski:
Thanks all, made the change

GuruPKK (approved):

casthewiz (approved):

jestapinski:
Hey had to make a one-line change for the bucket subdirectory destination as per a techops request and pushed that up here.

makefiles/Makefile.devhub-content-new
```diff
@@ -0,0 +1,59 @@
...
+	npm run build; \
+	cp -r "${REPO_DIR}/snooty-devhub-tmp/public" ${REPO_DIR};
+  
+next-gen-stage: ## Host online for review
+	mut-publish public ${STAGING_BUCKET} --prefix="${COMMIT_HASH}/${MUT_PREFIX}" --stage ${ARGS};
+	echo "${COMMIT_HASH}/${MUT_PREFIX}"
+	echo "Hosted at ${STAGING_URL}/${COMMIT_HASH}/${MUT_PREFIX}/${USER}/${GIT_BRANCH}/";
+
+next-gen-deploy:	
+	yes | AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID} AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY} mut-publish public ${PRODUCTION_BUCKET} --prefix=/developer --deploy --deployed-url-prefix=https://mongodb.com/developer --json --all-subdirectories ${ARGS};
```
jestapinski, line 58: Hey team, after speaking with IT Systems Eng, they recommended we do something like this where we parameterize the AWS credentials we are using for the job and pass in two new ones so we properly upload as the AWS user. This is also important since docs is going to consolidate infrastructure soon and the user associated with buckets may change.

Of course I can hand these over via 0bin when it is time to deploy but wanted to get a check on this just to be sure this is the way to do it here (unless we can set ENV secrets before running this portion of the job)

casthewiz: I think this is a good idea, but let's scope the parameters as well to something like `AWS_SECRET_ACCESS_KEY_DEVHUB` and `AWS_ACCESS_KEY_ID_DEVHUB` 

Just to be hyper specific and avoid any scope collisions, or a regression from an inadvertent change to the values in a config in the future when the infra consolidation occurs.

jestapinski: Great idea, just added it thanks

GuruPKK: Looks good

jestapinski:
Thanks all, this is ready for us to merge/test. I am merging the new config file (not used yet) from the front end shortly so whenever the team has some time I would like to get this out there.

Happy to set up a 30min meeting as well when ready

jestapinski:
Okay took an attempt at combining the configurations and marking dependencies as we discussed, let me know what you think and thanks!

jestapinski:
Commits seem fine to me, thanks @GuruPKK . Anything else you have seen?

PR #466, opened 2021-07-14
DOP-2343: fix mut prefix for manual to align with legacy

schmalliso:
Legacy writes the manual's search indexes to manual-, so this makes us do the same.


schmalliso:
5gh3kg

PR #467, opened 2021-07-14
DOP-2343: (TEMP HACK) fix manifest prefix for manual to align with legacy

schmalliso:
Ugh.

Relevant context: for reasons that are hard to deduce, we upload the Manual's search index manifest as docs-<something> rather than manual-<Something>. Because of the way the front-end works, that means we have "Docs" as a "Product" and "Manual" as a product, which is obviously silly. This will (very hackily) fix that situation while we sort out a more permanent solution as part of the Version Management project.

### DIFF
--- a/makefiles/Makefile.docs
+++ b/makefiles/Makefile.docs
@@ -23,10 +23,10 @@ SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
 SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
 PUSHLESS_DEPLOY_SHARED_DISABLED=true
 
-include ~/shared.mk
+CUSTOM_SEARCH_INDEX=true
+CUSTOM_NEXT_GEN_DEPLOY=true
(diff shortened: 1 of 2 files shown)


makefiles/Makefile.docs
```diff
@@ -86,6 +86,22 @@ next-gen-stage: ## Host online for review
 		echo "Hosted at ${STAGING_URL}/${MUT_PREFIX}/${USER}/${GIT_BRANCH}/"; \
 	fi
 
+next-gen-deploy:
```
GuruPKK, line 89: should we add this? cant we re use the one in shared.mk? i assume since we define CUSTOM_SEARCH_INDEX, next-gen-deploy-search-index defined here will be called?

schmalliso: Reasonable question! I was worried the "local" next-gen-deploy-search-index wouldn't be called, but I'm not sure why. We can give it a shot without this dupe.

schmalliso: Updated :)

PR #468, opened 2021-07-15
(TEMP WORKAROUND) DOP-2343: exclude landing and rename cloud-docs to Atlas

schmalliso:


makefiles/Makefile.cloud-docs
```diff
@@ -15,12 +15,20 @@ PROJECT=cloud-docs
...
+CUSTOM_SEARCH_INDEX=true
 
 SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
 SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
 
 include ~/shared.mk
 
+next-gen-deploy-search-index: ##THESE ARE TERRIBLE HACKS THAT WE SHOULD BE (AND ARE) ASHAMED OF
+	@echo "Building search index"
+	MANUAL_INDEX=$(subst cloud-docs,atlas,${MANIFEST_PREFIX})
```
i80and, line 27: Fair, but ow

i80and: (if we care this could be `INDEX_NAME` instead of `MANUAL_INDEX`)

i80and (approved):

makefiles/Makefile.cloud-docs
schmalliso: I shall change to INDEX_NAME since that's more practical now that it's not just for the manual :sob:

PR #469, opened 2021-07-15
DOP-2313 Create buckets with right policies

GuruPKK:
We are trying to have a single bucket instead of buckets for each repo. as there is a soft limit of 100 buckets and hard limit of 1000 buckets per account, we are exploring the option of serving all repos from single bucket


.github/workflows/deploy-dev-ecs.yml
```diff
@@ -0,0 +1,42 @@
...
+        with:
+          node-version: '14.x'
+      - name: Install Serverless Framework
+        run: npm install -g serverless
+      - name: Serverless AWS authentication
+        run: sls config credentials --provider aws --key ${{ secrets.AWS_ACCESS_KEY_ID }} --secret ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+      - name: Create ECS Repo
+        run: |
+          cd infrastructure/ecs-repo
+          npm install
```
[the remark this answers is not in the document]
casthewiz: We should use `npm ci` 😄

.github/workflows/deploy-dev-ecs.yml
```diff
@@ -0,0 +1,42 @@
...
+jobs:
+  deploy-dev:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '14.x'
+      - name: Install Serverless Framework
+        run: npm ci -g serverless
```
i80and, line 15: `npm ci` does not work for global packages

.github/workflows/deploy-dev-ecs.yml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Create ECS Repo
+        run: |
+          cd infrastructure/ecs-repo
+          npm install
+          sls deploy --stage dev
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY}}
+          IMAGE_TAG: ${{ github.sha }}
+      - uses: mr-smithers-excellent/docker-build-push@v5
```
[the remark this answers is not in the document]
GuruPKK: i think for lower environments, we can have just the latest tag and for prod we can tag it with the released version. i will remove it.

.github/workflows/deploy-dev-ecs.yml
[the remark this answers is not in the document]
GuruPKK: normally i dont check in package lock , should i do it?

worker/worker.js
```diff
@@ -173,6 +179,10 @@ module.exports = {
         // Sanitize the job (note that jobs that do not implement the sanitize function
         // will not proceed
 
+        if (currentJob.payload.regression) {
+          process.env.REGRESSION = true;
```
i80and, line 183: May I have the context on what this is intended to do?

i80and: I think it would be better to pass this explicitly into the child environments, rather than using environment inheritance

.github/workflows/deploy-dev-ecs.yml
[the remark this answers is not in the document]
i80and: package-lock should always be checked in, yes

Dockerfile
```diff
@@ -35,7 +35,7 @@ USER docsworker-xlarge
 WORKDIR /home/docsworker-xlarge
 
 # get shared.mk
-RUN curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/makefiles/shared.mk -o shared.mk
+RUN curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/DOP-2357/makefiles/shared.mk -o shared.mk
```
i80and, line 38: Probably not immediately actionable, but shared.mk should probably be pulled when we pull the entrypoint makefile, not during image creation

i80and: Also, is it desirable/possible to parameterize where we pull makefiles from?

worker/worker.js
GuruPKK: when we run our regressions, currently it overwrites the same bucket which beats the purpose of regressions. with this flag, we will be able to run the regressions so that the outputs are written to two different buckets and helps us in comparing the results. 

These are two webhooks for regressions 
https://webhooks.mongodb-stitch.com/api/client/v2.0/app/workerpoolstaging-qgeyp/service/githubRegressionTest/incoming_webhook/enqueueRegressionTestJobsInTest

https://webhooks.mongodb-stitch.com/api/client/v2.0/app/workerpool-boxgs/service/githubRegressionTest/incoming_webhook/enqueueRegressionTestJobsInProd

One operates out of prod and other operates out of staging. but they both treat it as STAGE job and writes it to the same bucket.

Dockerfile
GuruPKK: sure

worker/worker.js
GuruPKK: We do pass the regression flag as part of the payload but the buckets are chosen in make file and i thought having it in environment is the easy way to get it accessed in make file

.github/workflows/deploy-dev-ecs.yml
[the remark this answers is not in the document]
GuruPKK: Fixed it

Dockerfile
GuruPKK: We will address these as part of our rearchitecture project

worker/worker.js
i80and: We'd want to pass it in when we spawn the makefile process. But right now that would involve untangling some dragons, so this is fine I guess.

i80and (approved):

PR #477, opened 2021-08-17
DOP-2294 handle page data .json purges

GuruPKK:


casthewiz (approved):
Tentatively approving

We should watch the process.env behavior here - it's likely that any specified value for PURGE_ALL will be truthy, since I believe it should be coming in as a `string` if it's loading straight from the env.

This might make the `shouldPurgeAll()` logic redundant

GuruPKK:
> Tentatively approving
> 
> We should watch the process.env behavior here - it's likely that any specified value for PURGE_ALL will be truthy, since I believe it should be coming in as a `string` if it's loading straight from the env.
> 
> This might make the `shouldPurgeAll()` logic redundant

OK Fixed it, we will move to config package as part of our refactoring

casthewiz (approved):

PR #484, opened 2021-08-30
DOP-2415: Take 2 at fetching git submodules

i80and:
The `get-build-dependencies` target is only run in deploy builds, it seems,
so we have to do this to add the git submodule invocation.

### DIFF
--- a/makefiles/Makefile.docs-mongocli
+++ b/makefiles/Makefile.docs-mongocli
@@ -23,8 +23,52 @@ MUT_PREFIX ?= $(PROJECT)
 SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
 SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
 
+# Necessary to add the fetch-submodule dependency
+PUSHLESS_DEPLOY_SHARED_DISABLED=true
+
 include ~/shared.mk
+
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-mongocli
```diff
@@ -23,8 +23,52 @@ MUT_PREFIX ?= $(PROJECT)
...
+	cd snooty; \
+	echo "GATSBY_SITE=${PROJECT}" >> .env.production; \
+	if [ -n "${PATCH_ID}" ]; then \
+		echo "COMMIT_HASH=${COMMIT_HASH}" >> .env.production && \
+		echo "PATCH_ID=${PATCH_ID}" >> .env.production; \
+	fi && \
+	npm run build; \
+	cp -r "${REPO_DIR}/snooty/public" ${REPO_DIR};
+
+next-gen-stage: ## Host online for review
```
GuruPKK, line 59: Should this be moved here?

i80and: Unfortunately, it has to be -- because we're using GNU make instead of BSD make, targets can't be redefined, and the logic in shared.mk bundles these two targets together: `PUSHLESS_DEPLOY_SHARED_DISABLED` disables defining both `next-gen-html` and `next-gen-stage`

GuruPKK (approved):

PR #500, opened 2021-09-13
DOP-2420: Add feature flag to serverless.yml and ecs_service.yml

casthewiz:
Adds an environment variable definition for the consistent nav feature flag, so that it can pull values from parameter store.

### DIFF
--- a/infrastructure/ecs-main/serverless.yml
+++ b/infrastructure/ecs-main/serverless.yml
@@ -80,6 +80,7 @@ custom:
   npmEmail: ${ssm:/env/${self:provider.stage}/docs/worker_pool/npm/email}
   awsKeyDevhub: ${ssm:/env/${self:provider.stage}/docs/worker_pool/devhub/aws/key}
   awsSecretDevhub: ${ssm:/env/${self:provider.stage}/docs/worker_pool/devhub/aws/secret}
+  gatsbyFeatureFlagConsistentNavigation: ${ssm:/env/${self:provider.stage}/docs/worker_pool/npm/con_nav_flag}
   env:
     stg: "staging"
     prd: "production"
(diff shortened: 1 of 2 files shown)


GuruPKK (approved):

GuruPKK:
> I wouldn't do that - ideally we can just change the value in parameter store then redeploy, so we can toggle flags without a code change

So you mean every feature flag will be of convention flag/{feature}? I was thinking enabling a feature is also a property and {feature}/{property} can be better fit as there may be more properties for the same feature. even with this we wont be changing the code right? i am just talking about the convention.

casthewiz:
Flag exists in every 

> > I wouldn't do that - ideally we can just change the value in parameter store then redeploy, so we can toggle flags without a code change
> 
> So you mean every feature flag will be of convention flag/{feature}? I was thinking enabling a feature is also a property and {feature}/{property} can be better fit as there may be more properties for the same feature. even with this we wont be changing the code right? i am just talking about the convention.

I think that {feature}/{property} would make sense if we introduce complex, multi property feature flags, but for now just having the mapping be 1:1 makes the most sense to me.

We'll also be removing the flag after the launch of the nav, so timeline for this being present is only 2 weeks at most 😄

infrastructure/ecs-main/serverless.yml
```diff
@@ -80,6 +80,7 @@ custom:
   npmEmail: ${ssm:/env/${self:provider.stage}/docs/worker_pool/npm/email}
   awsKeyDevhub: ${ssm:/env/${self:provider.stage}/docs/worker_pool/devhub/aws/key}
   awsSecretDevhub: ${ssm:/env/${self:provider.stage}/docs/worker_pool/devhub/aws/secret}
+  gatsbyFeatureFlagConsistentNavigation: ${ssm:/env/${self:provider.stage}/docs/worker_pool/flag/consistent_nav}
```
casthewiz, line 83:
```suggestion
  gatsbyFeatureFlagConsistentNavigation: ${ssm:/env/${self:provider.stage}/docs/worker_pool/flag/consistent_nav/enabled}
```

GuruPKK (approved):

PR #504, opened 2021-09-17
Increase number of staging + production instances

casthewiz:
Parallelize all the things for our nav launches!


schmalliso (approved):
Oh I definitely fixed our fastServiceId in the wrong place yesterday.

PR #506, opened 2021-09-20
DOP-2464 Publishing docs-ecosystem blows docs-java and docs-golang away

GuruPKK:
Create new buckets for Java and go lang


makefiles/Makefile.docs-golang
```diff
@@ -0,0 +1,38 @@
+GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
```
rayangler, line 1: Should the makefile be in this PR?

PR #511, opened 2021-09-21
DOP-2481 Fix purging issues

GuruPKK:
Current purge implementation is not working. for the following reaosns

- We are using dochub service id and credentials for all other services we have in the system 
- We are purging by surrogate key which is implemented wrongly, unless a key is associated with a url we wont be able to use it. 

Implementation includes the following 

- pick ids and token based on the repo
- purge by url  , urls are returned by mut publish


worker/utils/environment.js
```diff
@@ -69,12 +82,82 @@ class EnvironmentClass {
...
+      return fastlyAtlasToken;
+    }  
+    if (repoName === "devhub-content") {
+      return fastlyDevhubToken;
+    }
+    return fastlyMainToken;
+  }
+
+  static isAtlasRepo(repoName) {
+    return repoName == "cloudgov-docs" || repoName == "cloud-docs" || repoName == "cloud-docs-osb";
```
casthewiz, line 158:
```suggestion
    return ["cloudgov-docs","cloud-docs","cloud-docs-osb"].includes(repoName);
```

worker/utils/environment.js
```diff
@@ -69,12 +82,82 @@ class EnvironmentClass {
...
+
+  static getFastlyMainToken() {
+    return fastlyMainToken;
+  }
+
+
+  static getFastlyDochubServiceId() {
+    if (fastlyDochubServiceId === undefined) {
       return 'testId';
     }
```
casthewiz, line 97: I feel like this should probably log an error if it's important, or otherwise should just be removed - idk how much utility 'testId' provides here

casthewiz (approved):
Nitpick for extensibility in `isAtlasRepo`, and potential redundancy in `getFastlyDochubServiceId()`.

Looks good and functional, but is there a reason why we don't directly reference the env var in the static getters?  e.g.
```
 static getFastlyDochubToken() {
    return process.env.FASTLY_DOCHUB_TOKEN;
  }
```

GuruPKK:
> Nitpick for extensibility in `isAtlasRepo`, and potential redundancy in `getFastlyDochubServiceId()`.
> 
> Looks good and functional, but is there a reason why we don't directly reference the env var in the static getters? e.g.
> 
> ```
>  static getFastlyDochubToken() {
>     return process.env.FASTLY_DOCHUB_TOKEN;
>   }
> ```

Not sure the cost involved with accessing the process environment every time over the cached values.

PR #514, opened 2021-09-27
Refactoring Autobuilder PHASE-1

GuruPKK:
This Includes the following Tickets
[DOP-2406](https://jira.mongodb.org/browse/DOP-2406)
[DOP-2407](https://jira.mongodb.org/browse/DOP-2407)
[DOP-2408](https://jira.mongodb.org/browse/DOP-2408)


src/job/jobValidator.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export class JobValidator implements IJobValidator {
+    _fileSystemService: IFileSystemServices;
+    _repoEntitlementRepository: RepoEntitlementsRepository;
+    constructor(fileSystemService: IFileSystemServices, repoEntitlementRepository: RepoEntitlementsRepository) {
+        this._fileSystemService = fileSystemService;
+        this._repoEntitlementRepository = repoEntitlementRepository;
+    }
+
+    async throwIfUserNotEntitled(job: IJob): Promise<void> {
```
schmalliso, line 22: I know we have this in the existing code, but how would this ever happen? 

Folks deploy via /deploy, which is populated based on a user's entitlements. How could they try to deploy a repo that isn't listed in the /deploy modal?

schmalliso (commented):
I have left one puny little comment, which is hilarious given how many files this touches :D

src/job/jobValidator.ts
GuruPKK: Yeah the idea is to make the code agnostic of the callers. if we dont check it here , then our code will be with an assumption that validations are done by caller. 
this assumption might impact us when we want to introduce an API to trigger the build where the payload will be provided by the caller directly( in this case it will be the writers). 

Said that, I am open for discussions on it

schmalliso: Reasonable rationale. I'm not enthusiastic about the idea of providing an API through which writers deploy -- in fact, I'm actively against it, but I suppose there's no harm in double-checking.

src/app.ts
```diff
@@ -0,0 +1,65 @@
...
+let fileSystemServices: FileSystemServices;
+let jobCommandExecutor: JobSpecificCommandExecutor;
+let githubCommandExecutor: GithubCommandExecutor;
+let jobRepository: JobRepository;
+let hybridJobLogger: HybridJobLogger;
+let repoEntitlementRepository: RepoEntitlementsRepository;
+let jobValidator: JobValidator;
+let cdnConnector: FastlyConnector;
+let repoConnector: GitHubConnector;
+let jobHandletFactory: JobHandlerFactory;
```
mango-db, line 27: nit: `jobHandletFactory` -> `jobHandlerFactory`

schmalliso:
@GuruPKK 👍🏻 +1 to Claire's suggestion to add a linter -- it seems sensible to handle that at this point.

PR #515, opened 2021-10-01
DOP-2483 Lambda for DOCHUB edge dictionary updates

GuruPKK:
**Changes:**

- Modify the Dochub trigger in Staging APP to send event over the event bridge on key collection update 
- Enable the partner source event which is in pending state from AWS console
- Introduce a lambda which will be triggered by the partner source event which is activated 
- Wire the lambda to invoke the upsert method implemented it in fastly connector.


serverless.yml
```diff
@@ -0,0 +1,56 @@
...
+
+default-env: &default-env
+  STAGE: ${self:provider.stage}
+  FASTLY_DOCHUB_MAP: ${self:custom.fastlyDochubMap}
+  FASTLY_DOCHUB_SERVICE_ID: ${self:custom.fastlyDochubServiceId}
+  FASTLY_DOCHUB_TOKEN: ${self:custom.fastlyDochubToken}
+
+
+functions:
+    v1UpsertEdgeDictionaryItem:
```
casthewiz, line 47: I'm not very enthusiastic about versioned api's when our own applications are the only customers for the api - mostly because it muddies the waters around expectations of upholding a contract

Just food for thought - it's a strong opinion, but weakly held 😅

src/services/cdn.ts
```diff
@@ -1,38 +1,33 @@
...
-import { IJobRepoLogger } from "./logger";
+import { ILogger } from "./logger";
 
 export const axiosApi = axios.create();
 
 
 export interface ICDNConnector {
     purge(jobId: String, urls: Array<string>): Promise<void>;
-    purgeAll(jobId: String, creds: any): Promise<any>;
+    purgeAll(creds: any): Promise<any>;
```
casthewiz, line 9: An interface for `creds` might be useful here - it seems like we're always expecting `id` and `token` to be present?

api/controllers/v1/dochub.ts
```diff
@@ -0,0 +1,10 @@
+import { FastlyConnector } from "../../../src/services/cdn";
+import c from "config";
+import { ConsoleLogger } from "../../../src/services/logger";
+export const UpsertEdgeDictionaryItem = async (event: any = {}): Promise<any> => {
+    const pair = {
+        key: event.detail.fullDocument.name,
+        value: event.detail.fullDocument.url
+    }
+    await new FastlyConnector(new ConsoleLogger).upsertEdgeDictionaryItem(pair, c.get('fastlyDochubMap'), c.get('cdn_creds')['dochub']);
```
casthewiz, line 9: do we need to `try` & `catch` here, or will any exceptions automagically bubble up into CloudWatch? 

I'm not 100% on our configuration or how much data would be in a failure of this lambda, if we need to debug

GuruPKK: exceptions will be logged in the cloudwatch. i can add try catcgh

PR #518, opened 2021-10-15
ACL Update for buckets to include list objects

GuruPKK:


i80and (commented):
We only need this for a single bucket -- is it desirable to turn it on for buckets search will never look at?

i80and (approved):

PR #532, opened 2021-11-14
DOP-2564 Worker pool changes to include buckets and url ondemand

GuruPKK:


src/repositories/repoBranchesRepository.ts
```diff
@@ -0,0 +1,22 @@
...
+        super(config, logger, "RepoEntitlementsRepository", db.collection(config.get("repoBranchesCollection")));
+    }
+
+    async getRepoBranchesByRepoName(repoName: string): Promise<any> {
+        const query = { 'repoName': repoName };
+        const repoDetails = await this.findOne(query, `Mongo Timeout Error: Timedout while retrieving Repoinformation for ${repoName}`);
+        if (repoDetails && repoDetails.bucket && repoDetails.url) {
+            return repoDetails
+        } else {
+            return { status: 'failure' };
```
casthewiz, line 19: We might want to make this a little more robust with error messaging - using `this._logger` would be useful here, especially to message to a writer that no results were found for the repo

PR #533, opened 2021-11-15
add-plugin: Add plugin-transform-runtime so tests run out-of-the-box

mango-db:
Adds the @babel/plugin-transform-runtime dependency, so tests run out-of-the-box


mango-db:
Closing due to https://github.com/mongodb/docs-worker-pool/pull/543

PR #538, opened 2021-11-19
Update Makefile.docs-mongodb-internal

rayangler:
Updated the Makefile to allow for old gen staging. I included the [lines](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/Makefile.docs#L100-L119) in the Makefile for docs that seemed relevant to staging, minus `publish` since I believe that would be for deploy jobs.


casthewiz (approved):
That looks right 👀

PR #592, opened 2022-02-28
DOP-2549-2: Absolute minimal Search infrastructure POC

mango-db:


src/job/jobHandler.ts
```diff
@@ -410,6 +414,50 @@ export abstract class JobHandler {
...
+      title: this._currJob.title + ' - search manifest generation',
+      user: this._currJob.user,
+      manifestPrefix: this._currJob.manifestPrefix,
+      pathPrefix: this._currJob.pathPrefix,
+      mutPrefix: this._currJob.mutPrefix,
+      buildCommands: [],
+      deployCommands: [],
+    };
+    try {
+      this._jobRepository.insertJob(manifestJob, this._config.get('jobsQueueUrl'));
```
GuruPKK, line 451: await?

mango-db: Ah, good catch - adding, thanks!

src/job/jobHandler.ts
```diff
@@ -410,6 +414,50 @@ export abstract class JobHandler {
...
+      title: this._currJob.title + ' - search manifest generation',
+      user: this._currJob.user,
+      manifestPrefix: this._currJob.manifestPrefix,
+      pathPrefix: this._currJob.pathPrefix,
+      mutPrefix: this._currJob.mutPrefix,
+      buildCommands: [],
+      deployCommands: [],
+    };
+    try {
+      await this._jobRepository.insertJob(manifestJob, this._config.get('jobsQueueUrl'));
```
kumaraguru-pk, line 451: Log the job id returned

GuruPKK (changes requested):
Manifest handler should have specific implementation rather than leaking it to based job handler

src/job/jobHandler.ts
```diff
@@ -414,6 +418,50 @@ export abstract class JobHandler {
     }
   }
 
+  // Creates and pushes the related manifestJob
+  @throwIfJobInterupted()
+  public async queueManifestJob(): Promise<void> {
+    // Rudimentary error prevention
+    if (this._currJob.payload.jobType.includes('manifestGeneration')) {
```
GuruPKK, line 425: if i understand it right, we are not allowed to queue multiple manifest generation for same job?

mango-db: Yes, that's the idea - does that seem incorrect?

GuruPKK (approved):

PR #598, opened 2022-03-03
manifest-prefix-fix: Add conditional to manifestPrefix function

mango-db:
[DOP-2801](https://jira.mongodb.org/browse/DOP-2801). This attempts to solve this issue of incorrect generation of search manifests e.g. `realm-.json `

- Also adds a conditional to the job validator to check that the project string exists. 
- Also moves the function declaration around to be more local to where it is used, and removes Promise / async
Is this appropriate @GuruPKK?

### DIFF
--- a/src/job/jobHandler.ts
+++ b/src/job/jobHandler.ts
@@ -217,7 +217,7 @@ export abstract class JobHandler {
       await this.constructPrefix();
       // if this payload is NOT aliased or if it's the primary alias, we need the index path
       if (!this.currJob.payload.aliased || (this.currJob.payload.aliased && this.currJob.payload.primaryAlias)) {
-        await this.constructManifestIndexPath();
+        this.currJob.payload.manifestPrefix = this.constructManifestPrefix();
       }
 
       this.prepStageSpecificNextGenCommands();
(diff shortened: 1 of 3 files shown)


src/job/jobHandler.ts
```diff
@@ -286,8 +286,12 @@ export abstract class JobHandler {
     return '';
   }
 
-  protected constructManifestIndexPath(): Promise<void> {
-    return Promise.resolve();
+  // For certain unversioned properties, urlSlug is null; use branchName instead
+  protected constructManifestPrefix(): string {
```
GuruPKK, line 290: doing it here means it will hide derived class implementation, is this intended

GuruPKK: Also i believe manifest generation done only for prod jobs

mango-db: I figured since the manifest index path should be constructed in the same way by any derived class, it should be OK to move it to the parent class, where it is called from. I also put functionality to kick-off manifestJobs from githubPush just for now (in the other PR) with the thought that it might be easier for us to test. We can remove that functionality once we are confident with the search manifests results. Thoughts?

GuruPKK (approved):

PR #599, opened 2022-03-03
DOP-2797 - Local trigger

GuruPKK:


api/controllers/v1/jobs.ts
```diff
@@ -146,6 +148,22 @@ function limit_message_size(message) {
...
+      console.log(ret);
+      if (ret) {
+        urls.push(...ret);
+      }
+    }
+    return urls;
+  }
+  return [];
+}
+
```
casthewiz, line 166:
```suggestion
function extractUrlFromMessage(fullDocument) {
  const { logs } = fullDocument;
  let urls = (logs?.length > 0 )
    ? logs.flatMap(log => log.match(/\bhttps?:\/\/\S+/gi) || []);
    : [];
  return urls;
}

```
Optional suggestion for readability + null safety

this seems like a good case for [flatMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap#for_adding_and_removing_items_during_a_map)

repl as PoC: https://replit.com/@casthewiz/BoringVioletTag#index.js

api/controllers/v1/jobs.ts
```diff
@@ -157,21 +175,23 @@ async function prepSummaryMessage(
...
+  const urls = extractUrlFromMessage(fullDocument);
+  let mms_urls = [null, null];
   if (repoName === 'mms-docs') {
-    lastMessage = `${lastMessage.slice(lastMessage.indexOf('mut-publish'))}\n\n${lastMessage.slice(
-      lastMessage.lastIndexOf('Summary')
-    )}`;
+    if (urls.length >= 2) {
+      mms_urls = urls.slice(-2);
+    }
+  }
```
casthewiz, line 184: I'd add a comment here explaining why we need to do this for mmsdocs

alternately we could move lines 179, and 181 through 183 to line 194 and consolidate this boolean

casthewiz (commented):
Just some comments for improving readability - I think we're getting to the point where we could move some stuff around or make some of the logic more concise

casthewiz (approved):

PR #605, opened 2022-03-15
Update slack.ts

GuruPKK:


api/controllers/v1/slack.ts
```diff
@@ -156,7 +156,7 @@ export const DeployRepo = async (event: any = {}, context: any = {}): Promise<an
     let urlSlug = branchObject.aliasObject.urlSlug; //string or null, string must match value in urlAliases or gitBranchName
     const isStableBranch = branchObject.aliasObject.isStableBranch; // bool or Falsey
     aliases = aliases?.filter((a) => a);
-    if (!urlSlug && non_versioned == false) {
+    if (!urlSlug || urlSlug === '') {
```
casthewiz, line 159: I don't think `urlSlug === ''` will ever evaluate - `''` will be falsey here

it might be worth running a `.trim()` on `urlSlug` in case it's just whitespace here, and then explicitly asserting truthiness? Something like `if (!urlSlug || !!urlSlug.trim())`?

casthewiz (approved):
I think this is good from what I know about version handling and slug behavior - added a suggestion to catch a potential edge case

api/controllers/v1/slack.ts
GuruPKK: done, updated it

PR #616, opened 2022-03-29
DOP-2680 Disable parallel Prod

GuruPKK:


.github/workflows/deploy-prd-ecs.yml
```diff
@@ -58,9 +45,8 @@ jobs:
         run: |
           cd infrastructure/ecs-main
           npm ci
-          sls deploy --stage prd
```
casthewiz, line 61: I would wait to change the `prd` deploy actions until we've deployed the code change and verified that behavior is as expected in production

casthewiz (commented):
I think we should wait on changing the production deploy actions until we verify that the code changes work as expected when deployed in production - I think we also need to keep dotcomstg until we're more confident that we've cleaned up the `docs.` vs `dotcom` logic in the parser and frontend

casthewiz (approved):

PR #626, opened 2022-04-14
Formalize job types

mango-db:
Changes job types from `string` to enum `JobType` to codify & reduce job type errors


mango-db:
Closing due to out-of-date

PR #627, opened 2022-04-19
Add a makefile for docs-meta

i80and:


makefiles/Makefile.docs-meta
```diff
@@ -0,0 +1,35 @@
...
+
+.PHONY: help html stage
+
+## Builds this branch's HTML under build/<branch>/html
+html:
+	giza make html
+
+
+get-build-dependencies:
+	curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-k8s-operator.yaml > ${REPO_DIR}/published-branches.yaml
```
rayangler, line 30: We're grabbing `docs-k8s-operator`'s yaml file here. Is this intentional for now? I'm unsure if we're actually using the information from the yaml file in the build anyways though.

rayangler (commented):
Just one question

PR #632, opened 2022-04-20
DOP-2866: removes continue when active != true

schmalliso:
Inactive branches may, regrettably, still need to be deployed. Inactive branches that build on Snooty need to be deployed
via the Autobuilder. This continue causes attempts to deploy active: false branches to silently fail to push to the queue,
which makes EOL-ing 'legacy' docs impossible.


casthewiz (approved):
💡 it may be worth exploring a change to the schema in the future to support `active` and `eol` as seperate concepts, but this should be fine for now!

PR #635, opened 2022-04-21
DOP-2866: permit building to '/' if urlSlug == null and publishOriginalBranchName == false

schmalliso:
Why would you want to do this? You're Charts and you've only got 1 actual version, but also a legacy version that you want to build. You actual version should stay at www.mongodb.com/docs/charts/ not at www.mongod.com/docs/charts/master, while 19.12 should build to www.mongodb.com/docs/charts/19.12/.


casthewiz (approved):
we should probably codify these cases outside of a mono function at some point
But that's a conversation for the future 🚀 LGTM!

PR #637, opened 2022-04-22
DOP-2902: Perform wildcard invalidations when invalidating more than 250 objects

gjchong25:
[DOP-2902](https://jira.mongodb.org/browse/DOP-2902)

This ticket allows for a single wildcard invalidation in the CloudFront CDN instead of multiple individual objects being invalidated, which is an expensive operation. 

Notes: 
Before merge, the temporary workaround for the staging pipeline and the threshold for URL wildcard purging (which is currently set to 1 for testing purposes, should be 250) will be changed. Extraneous console.log()s will also be removed, which were added for greater visibility in the logs.  

[Invalidation Status for Atlas](https://cdnvalidator.devops.staging.corp.mongodb.com/api/v1beta1/distributions/mongodbcom-staging-docs/invalidations/I2HFI80AX9C73S)
[Invalidation Status for Docs](https://cdnvalidator.devops.staging.corp.mongodb.com/api/v1beta1/distributions/mongodbcom-staging-docs/invalidations/IR806NQ7UPO0U)

### DIFF
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,7 +3,6 @@ on:
     branches:
       - "master"
       - "integration"
-      - "DOP-2902"
 concurrency:
   group: environment-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-2902"
 concurrency:
   group: environment-${{ github.ref }}
   cancel-in-progress: true
--- a/src/job/productionJobHandler.ts
+++ b/src/job/productionJobHandler.ts
@@ -100,13 +100,8 @@ export class ProductionJobHandler extends JobHandler {
       const updatedURLsArray = stdoutJSON.urls;
       // purgeCache purges the now stale content and requests the URLs to warm the cache for our users
       await this.logger.save(this.currJob._id, JSON.stringify(updatedURLsArray));
-      console.log('this.currJob.payload.pathPrefix : ' + this.currJob.payload.pathPrefix);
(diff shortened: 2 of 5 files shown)


schmalliso (approved):
Looks good to me! @i80and can you do a second pass, please?

src/services/cdn.ts
```diff
@@ -139,6 +139,13 @@ export class K8SCDNConnector implements ICDNConnector {
     urls = urls.map((item) => {
       return `/${item}`;
     });
+
+    // perform wildcard invalidations when invalidating >= 250 objects
+    if (urls.length >= 250) {
+      const wildcardPatternURL = '/' + pathPrefix + '/*';
```
i80and, line 145: Should we have an assertion here that pathPrefix isn't falsey? Are there circumstances where it would be okay for pathPrefix to be null in the first place?

.github/workflows/deploy-stg-ecs.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-2902"
```
casthewiz, line 6: We'll want to drop this before merging - just flagging, sure you're on top of it 😄

src/job/productionJobHandler.ts
```diff
@@ -100,7 +100,8 @@ export class ProductionJobHandler extends JobHandler {
       const updatedURLsArray = stdoutJSON.urls;
       // purgeCache purges the now stale content and requests the URLs to warm the cache for our users
       await this.logger.save(this.currJob._id, JSON.stringify(updatedURLsArray));
-      const id = await this._cdnConnector.purge(this.currJob._id, updatedURLsArray);
+      console.log('this.currJob.payload.prefix : ' + this.currJob.payload.prefix);
```
casthewiz, line 103: If we want to retain this log statement, we might want to log out a little more context here, to help anyone reading the log outputs in CloudWatch.

casthewiz (approved):
Two small nits, but looks great!

PR #638, opened 2022-04-28
DOP-2711: add makefile & published branches for cluster-sync

schmalliso:
Also adds a dodgy little publishedbranches since there's a build step that looks for it even though it's unnecessary.

### DIFF
--- a/makefiles/Makefile.docs-cluster-to-cluster-sync
+++ b/makefiles/Makefile.docs-cluster-to-cluster-sync
@@ -1,0 +1,21 @@
+GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
+
+ifeq ($(STAGING_USERNAME),)
+	USER=$(shell whoami)
+else
+	USER=$(STAGING_USERNAME)
+endif
+
+PROJECT=cluster-sync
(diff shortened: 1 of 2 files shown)


makefiles/Makefile.docs-cluster-to-cluster-sync
```diff
@@ -0,0 +1,21 @@
...
+REPO_DIR=$(shell pwd)
+
+SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
+SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
+
+include ~/shared.mk
+
+
+get-build-dependencies: 
+	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-cluster-to-cluster-sync.yaml > ${REPO_DIR}/published-branches.yaml
```
i80and, line 21: Having a `@` sigil here REALLY bugs me, but it's also consistent with our other files.

I think it's also much better to use curl's `-o` option rather than shell redirection as a matter of course. It shouldn't really matter here, but if curl fails, as-is this will still create that file.

i80and (approved):

PR #642, opened 2022-05-04
DOP-2935

schmalliso:


makefiles/Makefile.docs-app-services
```diff
@@ -0,0 +1,21 @@
+GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
+
+ifeq ($(STAGING_USERNAME),)
+	USER=$(shell whoami)
+else
+	USER=$(STAGING_USERNAME)
+endif
+
+PROJECT=atlas-app-services
```
i80and, line 9: Frankly, I thinkwe should just move every single line of this file except `PROJECT` and a definition for `get-build-dependencies`

i80and (approved):

makefiles/Makefile.docs-app-services
schmalliso: I mean, we should delete the publishedbranches bit too since it's not used anymore, but who has time to make the relevant updtes :)

PR #643, opened 2022-05-04
DOP-N/A: Use repoBranches url instead of payload url

casthewiz:
Uses repoBranches to source the basename url instead of the payload url (which is currently the source url for the content repo)


schmalliso:
Noice.

schmalliso (approved):

rkhullar-mongodb (approved):

PR #644, opened 2022-05-05
[DOP-2946] stabilize staging pipeline

rkhullar-mongodb:
This PR updates the triggers for the dev and staging pipeline so that we don't deploy into staging via pull request.
We'll still have to revisit the pipeline, but this should make staging reflect the main branch.

If we need to deploy into staging of a feature branch, then the pipeline trigger can be manually updated for that branch.

### DIFF
--- a/.github/workflows/deploy-dev-ecs.yml
+++ b/.github/workflows/deploy-dev-ecs.yml
@@ -1,9 +1,12 @@
 on:
   push:
-    branches-ignore:
-      - "integration"
+    branches:
       - "master"
-      - "meta"
+      - "integration"
+  pull_request:
+    branches:
+      - master
+      - integration
 concurrency:
   group: environment-${{ github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-dev-ecs.yml
```diff
@@ -1,9 +1,12 @@
 on:
   push:
-    branches-ignore:
-      - "integration"
+    branches:
       - "master"
-      - "meta"
+      - "integration"
```
schmalliso, line 5: @casthewiz Are there situations where you foresee we'd want the dev deploy to run prior to a PR being opened? My feeling is that folks use the dev builds to confirm that their work is - vaguely - working prior to opening up a PR, but I'm not sure that's actually an accurate assessment.

If it *is* an accurate assessment, we may want to revisit the changes to this file.

rkhullar-mongodb: So the pipeline definitely needs followup changes IMO. I'd opt for have the build/test pipelines only run for commits that are part of a PR. I think running the pipeline for any (non-PR) branch can get out of control.

casthewiz (approved):

PR #645, opened 2022-05-06
DOP-N/A: Only enqueue a manifest for primaryAliases

casthewiz:
Adds a boolean to check if the property being deployed is aliased, and whether it is the primary alias.

Only enqueues a manifest generation on primary alias jobs for aliased properties.

### DIFF
--- a/src/job/jobHandler.ts
+++ b/src/job/jobHandler.ts
@@ -446,6 +446,11 @@ export abstract class JobHandler {
     if (doNotSearchProperties.includes(this.currJob.payload.repoName)) {
       return false;
     }
+    // Ensures that we only build a manifest for aliased properties if the job
+    // is for a primary alias
+    if (this.currJob.payload.aliased && !this.currJob.payload.primaryAlias) {
+      return false;
+    }
     // Edit this if you want to generate search manifests for dev environments, too
     if (this.currJob.payload.jobType !== 'productionDeploy') {
       return false;


src/job/jobHandler.ts
```diff
@@ -446,6 +446,11 @@ export abstract class JobHandler {
     if (doNotSearchProperties.includes(this.currJob.payload.repoName)) {
       return false;
     }
+    // Ensures that we only build a manifest for aliased properties if the job
+    // is for a primary alias
+    if (this.currJob.payload.aliased && !this.currJob.payload.primaryAlias) {
+      return false;
+    }
```
rkhullar-mongodb, line 453: interesting. looks fine to me. curious if we want the default behavior of the whole function to return true. working with the inverse logic for when we don't want to build the manifest seems tricky.

rkhullar-mongodb (approved):

src/job/jobHandler.ts
casthewiz: I think that by default we want to build the manifest except in exclusion cases
and that the exclusion cases are themselves non-typical, e.g. we don't want to index our homepage and a subset of properties have aliasing schemes

So I err on the side of this should return default true, that way we can track exclusion cases in the code, since they are non-typical and fairly static

the only 'nontypical' case here is checking if it's a production deploy vs. staging deploy, but we can always move this function into the `productionJobHandler` in the future

PR #656, opened 2022-05-20
Update Makefile.devhub-content

casthewiz:


makefiles/Makefile.devhub-content
```diff
@@ -56,6 +56,8 @@ next-gen-stage: ## Host online for review
 	mut-publish public ${BUCKET} --prefix=/developer --stage ${ARGS};
 	echo "${COMMIT_HASH}/${MUT_PREFIX}"
 	echo "Hosted at ${URL}/developer";
+	mut-publish public developer-test --prefix="${COMMIT_HASH/MUT_PREFIX}" --stage ${ARGS}
```
ajhenri, line 59: Should this be the same as line 56?
```suggestion
	mut-publish public developer-test --prefix=/developer --stage ${ARGS}
```

casthewiz: we can do it that way to ensure that it goes to the right spot, sure 👍

PR #658, opened 2022-05-24
bugfix iam policy for reading params from ecs

rkhullar-mongodb:
- https://github.com/mongodb/docs-worker-pool/pull/653


rkhullar-mongodb:
also saw a connection error in the ecs logs, so updated the ondemandapp script

PR #659, opened 2022-05-25
[DOP-2993] remove aws account ids - part 1

rkhullar-mongodb:
The purpose of this PR is to remove our aws configuration from three types of files: github workflows, serverless framework files, and cloud-formation stacks. Eventually we should remove hardcoded aws config from the whole codebase.

While the AWS account id and region aren't totally sensitive values, there's no benefit to having them hardcoded. Within the github workflows the account id can be accessed as an output of the aws config step. Within the serverless yaml files the account id can actually be access via `${aws.accountId}`.


casthewiz (approved):
Nice! Definitely a step in the right direction for making us less hardcoded.

casthewiz (approved):

PR #663, opened 2022-06-01
[DOP-3006] add redirect for realm 

rkhullar-mongodb:
## Related Links
- [DOP-3006](https://jira.mongodb.org/browse/DOP-3006)
- [DOP-2956](https://jira.mongodb.org/browse/DOP-2956)

## Description
Add redirect from `mongodb.com/docs/realm/cloud/*` to `mongodb.com/docs/app-services/*`.


rkhullar-mongodb:
For now I've manually updated the s3 config for the two buckets with the redirect rules.

schmalliso:
@rkhullar-mongodb Revert the manual changes. They're redirecting pages that should not be included in the redirects.

infrastructure/ecs-main/serverless.yml
```diff
@@ -156,6 +156,19 @@ custom:
     prd: "prd"
     dotcomstg: "preprd"
     dotcomprd: "prd"
+  site:
+    host:
+      dev: "docs-dev.mongodb.com" # NOTE: cannot be empty in redirect
```
casthewiz, line 161: Is this intended to be a future dev environment?

infrastructure/ecs-main/serverless.yml
```diff
@@ -156,6 +156,19 @@ custom:
...
+      stg: "mongodbcom-cdn.website.staging.corp.mongodb.com"
+      dotcomstg: "mongodbcom-cdn.website.staging.corp.mongodb.com"
+      prd: "www.mongodb.com"
+      dotcomprd: "www.mongodb.com"
+    prefix:
+      dev: "docs-dev"
```
casthewiz, line 167: I think this should be just `docs` - that way we can have `docs-dev.mongodb.com/docs`

casthewiz (approved):
I think this is good - we should be able to test the redirects in staging at `https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/realm/` and its cloud paths 😄

infrastructure/ecs-main/serverless.yml
rkhullar-mongodb: So this is intentionally not valid url currently. If we setup a future domain for dev, then I think this one would be a good fit. I could have used the actual s3 website url domain here, but that would include our aws account id.

infrastructure/ecs-main/serverless.yml
rkhullar-mongodb: I'll update this. Would we want to eventually do the same for staging? Move from `{stage-domain}/docs-qa` to `{stage-domain}/docs`?

PR #668, opened 2022-06-14
[DOP-2715] create site archive after publish

rkhullar-mongodb:
This PR hooks into the job completion logic in order to trigger an archival system implement with AWS Batch. The static content for published sites are uploaded to s3 in the form of zip files and are named according to date time. 

The archive buckets are `docs-archive-{env}-mongodb` for `dev`, `dotcomstg`, and `dotcomprd`. If we need to rollback a site we'll be able to manually submit that work within the AWS Batch console.


api/controllers/v1/jobs.ts
```diff
@@ -69,6 +71,10 @@ export const HandleJobs = async (event: any = {}): Promise<any> => {
           case JobStatus[JobStatus.completed]:
             queueUrl = c.get('jobUpdatesQueueUrl');
             await NotifyBuildSummary(jobId);
+            const environments = new Set(['dev', 'dotcomstg', 'dotcomprd']);
+            if (environments.has(c.get('env'))) {
+              await SubmitArchiveJob(jobId);
```
rkhullar-mongodb, line 76: should i wrap this in a try catch?

api/controllers/v1/jobs.ts
```diff
@@ -9,6 +9,8 @@ import { JobQueueMessage } from '../../../src/entities/queueMessage';
 import { Job, JobStatus } from '../../../src/entities/job';
 import { ECSContainer } from '../../../src/services/containerServices';
 import { SQSConnector } from '../../../src/services/queue';
+import { Batch } from '../../../src/services/batch';
+import {IConfig} from "config";
```
drew-beckmen, line 13: nit but to ensure consistent spacing inside braces, consider `{ IConfig }`

api/controllers/v1/jobs.ts
```diff
@@ -222,3 +228,27 @@ async function NotifyBuildProgress(jobId: string): Promise<any> {
     statusCode: 200,
   };
 }
+
+function getMongoClient(config: IConfig): mongodb.MongoClient {
```
drew-beckmen, line 232: We initialize a `MongoClient` in a few different places using the same connection string. might be worth consolidating the logic into a single helper function

drew-beckmen (commented):
LGTM, but I'll leave the final approval to the experts. I got the demo yesterday, so I'm convinced it works :)

api/controllers/v1/jobs.ts
rkhullar-mongodb: Yeah I think that's a good idea, but should probably be in a different PR.

Most of the initializations happen with `new mongodb.MongoClient(c.get('dbUrl'))`.  I could have used that here as well, but I'm leaning towards building the the connection url using the individual parts (host, username, password, e tc). We may also want to consider explicitly closing the connection in code.

api/controllers/v1/jobs.ts
```diff
@@ -69,6 +71,10 @@ export const HandleJobs = async (event: any = {}): Promise<any> => {
           case JobStatus[JobStatus.completed]:
             queueUrl = c.get('jobUpdatesQueueUrl');
             await NotifyBuildSummary(jobId);
+            const environments = new Set(['dev', 'dotcomstg', 'dotcomprd']);
+            if (environments.has(c.get('env'))) {
+              await SubmitArchiveJob(jobId);
+            }
```
casthewiz, line 77: I feel that this logic should exist in the `submitArchiveJob` function and not in this function - the batch service can hold the responsibility of determining whether or not the environment is valid for batch submission

Not a strongly held opinion, let me know if you disagree! 😄

api/controllers/v1/jobs.ts
```diff
@@ -222,3 +228,33 @@ async function NotifyBuildProgress(jobId: string): Promise<any> {
...
 }
+
+function getMongoClient(config: IConfig): mongodb.MongoClient {
+  const url = `mongodb+srv://${config.get('username')}:${config.get('password')}@${config.get(
+    'host'
+  )}/?retryWrites=true&w=majority`;
+  return new mongodb.MongoClient(url);
+}
+
+async function SubmitArchiveJob(jobId: string) {
```
casthewiz, line 239: Is there a reason for why we can't use the client defined on line 16 here, passed in as an arg?

api/controllers/v1/jobs.ts
rkhullar-mongodb: So i also think this logic is a little whacky, but i'd like to keep it as is. The problem is having environment conditional logic. But assuming that we only want the archive / rollbacks to happen for `dev` `dotcomstg` and `dotcomprd`, this place in the flow makes the most sense to me. I think it'd be worse to push it to the lower level functions, since it'd be confusing to see a function invocation `await SubmitArchiveJob(jobId)` and then potentially that function doesn't actually do anything.

Do you think it'd be better to remove the conditional logic all-together and create archive / deploy resources for `stg` and `prd`?

drew-beckmen (approved):

api/controllers/v1/jobs.ts
casthewiz: I fully agree that ideally we don't have to manage a set of strings to match on 
I think the archive/deploy resources for environments would be ideal tbh

but I also think that `await SubmitArchiveJob(jobId)` wouldn't be confusing at all, since we don't assign a response or expect that function that mutate state within the function that's invoking it - as is the higher level function doesn't care or have responsibility over the results of the invocation

I'd rather have the environmental string matching in the lower level function, since if we truly want to only run on environment match, it makes sense to enforce it in the function executing the logic, not the invoker

But I'd rather not deal with the environment matching at all, so if it seems a reasonable lift, adding archive / deploy resources for `stg` and `prd` seems best

rkhullar-mongodb: sounds good. i'll create the resources and remove the conditional logic from this code.

src/services/batch.ts
```diff
@@ -0,0 +1,29 @@
...
+      jobName: 'archive',
+      jobQueue: `docs-archive-${this.environment}`,
+      jobDefinition: `docs-archive-${this.environment}`,
+      parameters: {
+        'source-bucket': sourceBucket,
+        'target-bucket': targetBucket,
+        'site-name': siteName,
+      },
+    });
+    return await this.client.send(command);
```
casthewiz, line 27: nit: we don't need to `return await` here

https://eslint.org/docs/latest/rules/no-return-await

casthewiz (approved):
just one nitpick

src/services/batch.ts
rkhullar-mongodb: updated

casthewiz (approved):

PR #669, opened 2022-06-16
[DOP-3076] bucket policies

rkhullar-mongodb:
## Related Links
- https://github.com/mongodb/docs-worker-pool/pull/668

## Description
This PR follows up to #668 by updating the site bucket policies to allow rollbacks with AWS batch.
If we need to rollback a site, we can submit the job within the AWS console and specify the following job parameters:
- source-bucket
- target-bucket
- site-name
- version

The job runs with an IAM role called `docs-archive-deploy-poc`. Eventually we should create roles for each environment. The bucket policies are updated such that services using that role are allowed to push files into s3. There are seven buckets used for site hosting in each environment.


casthewiz (approved):
Nit: Can we change the role name to `docs-archive-deploy-root` or another status that denotes broad permissioning?

rkhullar-mongodb:
> Nit: Can we change the role name to `docs-archive-deploy-root` or another suffix that denotes broad permissioning?

@casthewiz would it be ok to handle that as follow up to this PR. I'm actually thinking we should have roles for each environment and job type. For example `docs-archive-dev` `docs-deploy-dotcomstg` etc

casthewiz:
Doing it as a follow up is fine 😄 👍

PR #674, opened 2022-06-24
DOP-3071: Update docs-404 Makefile to upload built assets to multiple environment-specific S3 buckets

drew-beckmen:
### Ticket

[DOP-3071](https://jira.mongodb.org/browse/DOP-3071?src=confmacro)

### Notes

To add custom 404 pages back to our site post subdomain consolidation, we need to upload the built version of the `docs-404` content repo to each of the 7 buckets that hosts docs content (`docs-java-{env}`, `docs-mongodb-org-{env}`, etc). Since we only need to display a custom 404 page in production and pre-production environments, the Makefile only uploads the built files using `mut` if `SNOOTY_ENV` is equal to `dotcomprd` or `dotcomstg`.

### Testing Story

After this PR has been merged, we will need to run a `/test-deploy` in Slack on `docs-404` to see if the files get uploaded successfully.

- [ ] Before running `/test-deploy`, we'll need to add `docs-404` to the `pool.repo_branches` collection in order to satisfy the autobuilder. Here's my suggested document to insert:

```
{
  "repoName": "docs-404",
  "branches": [
    {
      "publishOriginalBranchName": false,
      "active": true,
      "gitBranchName": "master",
      "isStableBranch": false,
      "urlAliases": null,
      "urlSlug": null,
      "versionSelectorLabel": "master",
      "buildsWithSnooty": true
    }
  ],
  "bucket": {
    "dotcomstg": "docs-mongodb-org-dotcomstg",
    "dotcomprd": "docs-mongodb-org-dotcomprd"
  },
  "url": {
    "dotcomprd": "http://mongodb.com/",
    "dotcomstg": "https://mongodbcom-cdn.website.staging.corp.mongodb.com/"
    
  },
  "prefix": {    
    "dotcomstg": “docs/404”,
    "dotcomprd": "docs/404”
  },
  "project": “docs-404”
}
```

In the Makefile, you'll see that we leverage the `BUCKET` environment variable, which will be either `docs-mongodb-org-dotcomstg` or `docs-mongodb-org-dotcomprd` since the autobuilder sources the value from `repo_branches`.

PR #679, opened 2022-06-30
DOP-3071: Update CloudFormation template to specify custom S3 error document

drew-beckmen:
### Ticket

[DOP-3071](https://jira.mongodb.org/browse/DOP-3071)

### Notes

A follow-up to #674. Successfully tested the Makefile in preprod `dotcomstg` environment. We now need to update our CloudFormation stack to ensure the 404 page is served in both `dotcomstg` and `dotcomprd` environments.

Goal is to only set the `ErrorDocument` property in `dotcomstg` and `dotcomprd`. Outside of those environments, we will continue to serve the default S3 error page.


rkhullar-mongodb (approved):
LGTM

PR #681, opened 2022-07-05
DOP-2866: fix semi-versioned repo logic harder

schmalliso:
Previously 'fixed' in https://github.com/mongodb/docs-worker-pool/pull/635 but actually this case hits earlier in the loop logic and then getting `continue`'d because the `master` branch doesn't have any aliases.

### DIFF
--- a/api/controllers/v1/slack.ts
+++ b/api/controllers/v1/slack.ts
@@ -144,7 +144,8 @@ export const getDeployableJobs = async (values, entitlement, branchRepository: B
     );
 
     if (!aliases || aliases.length === 0) {
-      if (non_versioned) {
+        // handle non-versioned repos AND repos where only 1 version is active where there are no aliases
+      if (non_versioned || (!publishOriginalBranchName && urlSlug === null))) {
         newPayload.urlSlug = '';
       }
       deployHelper(deployable, newPayload, jobTitle, jobUserName, jobUserEmail);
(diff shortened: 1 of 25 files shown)


schmalliso:
Note to self: stop making edits in github, it only leads to annoyance when the lint doesn't pass.

api/controllers/v1/slack.ts
```diff
@@ -144,7 +144,8 @@ export const getDeployableJobs = async (values, entitlement, branchRepository: B
     );
 
     if (!aliases || aliases.length === 0) {
-      if (non_versioned) {
+      // handle non-versioned repos AND repos where only 1 version is active where there are no aliases
+      if (non_versioned || (!publishOriginalBranchName && urlSlug === null)) {
```
drew-beckmen, line 148: nit: this is the same logic (and comment) as line 167. perhaps toss it into a helper function?

schmalliso: Reasonable nit! Broadly we should refactor all of this logic so it isn't hardcoded in the way it currently is, but that's probably out of the scope here. You'll note it's a SLIGHTLY different comment, since this mentions aliases while the other doesn't, but it is a very subtle difference :)

schmalliso:
@drew-beckmen  re: not seeing your changes

You can pick which commits you're diffing from in the top left drop-down menu if you're so inclined.

api/controllers/v1/slack.ts
drew-beckmen: Agreed on the refactor. I believe we're tracking some similar work in [DOP-3079](https://jira.mongodb.org/browse/DOP-3079), and Raymund and I recently had a [discussion on GitHub](https://github.com/mongodb/snooty/pull/644#discussion_r906333253) for a similar edge case bug fix in `snooty`

drew-beckmen (approved):

PR #690, opened 2022-07-25
DOP-2246: Remove publishedbranches

i80and:


schmalliso (approved):
This looks good, though we should ensure we've pulled out the corresponding stuff in master before we actually merge.

PR #696, opened 2022-09-09
DOP-3188: add bucket definitons for csharp to severless defs

schmalliso:
I *think* these things are managed by the mysterious serverless framework now, and want to ensure our buckets end up with the correct policy, so updating here accordingly.


schmalliso:
YOLO merging so can test in preprd

PR #698, opened 2022-10-03
Increase CPU units for prd instance

schmalliso:
Based on cursory investigation, Gatsby seems to make good use of additional CPUs and doesn't appear to be memory-bound. Testing out increasing the number of available CPUs on `prd` to assess its impact.


casthewiz (approved):

casthewiz (commented):
Might be worth upping `dotcomstg` also.

schmalliso:
I wanna compare 😁

PR #703, opened 2022-10-25
Dop 3203

casthewiz:
#### Associated Products Merging Logic

Snapshot test suite will be added in DOP-3370. For now, we can focus on making sure that abstractions are clean and clearly partitioned.

Staging behavior to be accommodated in https://jira.mongodb.org/browse/DOP-3410


modules/persistence/src/services/assets/index.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export const upsertAssets = async (zip: AdmZip) => {
+  try {
+    const assets = await assetsFromZip(zip);
+    return Promise.all(assets.map((asset) => upsert(asset, COLLECTION_NAME, asset._id)));
+  } catch (error) {
+    console.error(`Error at upsertion time for ${COLLECTION_NAME}: ${error}`);
+    throw error;
+  }
+};
```
casthewiz, line 23: We weren't persisting `assets` as a part of the initial work, which was causing issues on trying to build from entries sourced via DB. This replicates logic from the parser for handling asset upsertion.

modules/persistence/src/services/metadata/associations/index.ts
```diff
@@ -0,0 +1,147 @@
...
+// Convert our cursor from the shared ToC aggregation query into a series of ToC objects
+const shapeToCsCursor = async (
+  tocCursor: AggregationCursor,
+  repoBranchesMap
+): Promise<{ tocInsertions: ToCInsertions; tocOrderInsertions: TocOrderInsertions }> => {
+  let tocInsertions, tocOrderInsertions;
+  await tocCursor.forEach((doc) => {
+    // TODO: If we want staging builds with embedded versions, it needs to be added here
+    if (repoBranchesMap[doc._id.project][doc._id.branch]) {
+      tocInsertions[doc._id.project][doc._id.branch] = doc.most_recent.tocTree;
```
seungpark, line 88: seems like these likes will throw errors when reading project from tocInsertions and tocOrderInsertions (undefined values in line 84)

seungpark (commented):
at first pass i couldn't test for metadata merge since there was no call to insertUmbrellaMetadata. left a few styling comments below but still have to look through the toc merging logic

modules/persistence/src/services/metadata/associations/index.ts
```diff
@@ -0,0 +1,147 @@
...
+const sharedMetadataEntry = async (metadata): Promise<SharedMetadata> => {
+  try {
+    const snooty = await db();
+    const entry = await snooty
+      .collection('metadata')
+      .find({
+        $elemMatch: { associated_products: metadata.project },
+      })
+      .sort({ build_id: -1 })
+      .limit(1)
```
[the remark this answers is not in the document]
casthewiz: Unfortunately findOne() doesn't let us sort since it doesn't return a cursor 😬

modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -0,0 +1,76 @@
...
+// BFS through the toctree from the metadata entry provided as an arg
+// and insert matching tocInsertion entries + tocOrders
+export const traverseAndMerge = (
+  metadata: SharedMetadata,
+  tocInsertions: ToCInsertions,
+  tocOrderInsertions: TocOrderInsertions
+) => {
+  const { toctree, associated_products } = metadata;
+
+  let queue = [toctree];
```
seungpark, line 66: guessing fifo/filo doesn't matter here since we have to traverse all the children?

modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -0,0 +1,76 @@
...
+  while (queue) {
+    let next = queue.pop();
+    if (next && isInsertionCandidateNode(next, associated_products)) {
+      next = mergeNode(next, tocInsertions);
+      metadata.toctreeorder = mergeTocTreeOrder(metadata, next, tocOrderInsertions);
+    }
+    if (next?.children) queue = [...queue, ...next.children];
```
seungpark, line 73: also here, are we supporting more than one level deep for merging toc?

modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -0,0 +1,76 @@
...
+// and insert matching tocInsertion entries + tocOrders
+export const traverseAndMerge = (
+  metadata: SharedMetadata,
+  tocInsertions: ToCInsertions,
+  tocOrderInsertions: TocOrderInsertions
+) => {
+  const { toctree, associated_products } = metadata;
+
+  let queue = [toctree];
+  while (queue) {
```
seungpark, line 67: infinite loop for empty array here

seungpark (changes requested):
ran some tests with associated product, will be running additional tests with umbrella product

modules/persistence/src/services/metadata/index.ts
```diff
@@ -23,3 +23,13 @@ export const insertMetadata = async (buildId: ObjectId, zip: AdmZip) => {
     throw error;
   }
 };
+
+export const insertUmbrellaMetadata = async (buildId: ObjectId, zip: AdmZip) => {
+  try {
+    const umbrellaMetadata = await mergeAssociatedToCs(metadataFromZip(zip)[0]);
+    return umbrellaMetadata ? insert([umbrellaMetadata], COLLECTION_NAME, buildId) : undefined;
```
seungpark, line 30: getting an error here for same document._id. should strip, or upsert?

modules/persistence/src/services/metadata/ToC/index.ts
casthewiz: fifo matters here for future optimizations, ideally we'll stop bfs after hitting all association matches specified in the `associated_products` list

and all matches should be high in the toctree, thus bfs and not dfs - I'll switch this to use `.shift()` instead of `.pop()`, very nice catch

modules/persistence/src/services/metadata/ToC/index.ts
casthewiz: Yep - we support more than one level deep. If my logic's right this is extending the queue so long as `children` are present 🤔, so we should be traversing the entire toctree

modules/persistence/src/services/metadata/associations/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    .collection('repos_branches')
+    .find({ project: { $in: associatedProductNames } })
+    .toArray();
+  return entries as unknown as ReposBranchesDocument[];
+};
+
+const mapRepoBranches = (repoBranches: ReposBranchesDocument[]) =>
+  Object.fromEntries(
+    repoBranches.map((entry) => {
+      const branches = Object.fromEntries(entry.branches.map((branch) => [branch.name, branch]));
```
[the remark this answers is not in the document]
casthewiz: Changed to `gitBranchName` ... prop overload, running out of RAM in my head to keep it all straight 😅

modules/persistence/src/services/metadata/associations/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    }
+  });
+
+  return { tocInsertions, tocOrderInsertions };
+};
+
+const queryForAssociatedProducts = async (metadata, sharedMetadataEntry) => {
+  try {
+    // Do a comparison between the local metadata entry and the shared metadata
+    const localNewerThanShared = metadata.created_at > sharedMetadataEntry.created_at;
```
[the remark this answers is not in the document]
casthewiz: Revisiting this case - I vaguely remember thinking we need to do a time compare here for most recent, but may not be necessary.

modules/persistence/src/services/metadata/index.ts
casthewiz: This is now stripping the `_id` field, since we want to treat it as a new document entirely 👍 - thanks for flagging!

modules/persistence/src/services/connector/pool/index.ts
```diff
@@ -0,0 +1,17 @@
+import { MongoClient, Db } from 'mongodb';
+
+// cached db object, so we can handle initial connection process once if unitialized
+let dbInstance: Db;
+// Handles memoization of db object, and initial connection logic if needs to be initialized
+export const db = async (client: MongoClient) => {
```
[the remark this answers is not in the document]
casthewiz: I think that in a larger application or monolith I would have gone down that route, but since this is mostly a bounded context I've taken the approach of adding additional comments where it's exported in the higher level `connector/index.ts`.

The intent is that we shouldn't really have to make modifications in this file, so I prefer this approach - I'd rather hide the connection implementation/helpers here, and just export the `pool` connection from the `connector` with comment caveats around not touching it unless absolutely necessary.

modules/persistence/src/services/metadata/associations/index.ts
```diff
@@ -0,0 +1,147 @@
...
+  const db = await pool();
+  const entries = await db
+    .collection('repos_branches')
+    .find({ project: { $in: associatedProductNames } })
+    .toArray();
+  return entries as unknown as ReposBranchesDocument[];
+};
+
+const mapRepoBranches = (repoBranches: ReposBranchesDocument[]) =>
+  Object.fromEntries(
```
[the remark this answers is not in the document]
casthewiz: I see it a bit better now but I think that I'd prefer the fromEntries for semantic readability - happy to revisit if we hit a bottleneck here 🧐

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -0,0 +1,157 @@
...
+    }
+  });
+
+  return { tocInsertions, tocOrderInsertions };
+};
+
+const getAssociatedProducts = async (metadata, sharedMetadataEntry) => {
+  try {
+    // Do a comparison between the local metadata entry and the shared metadata
+    const localNewerThanShared = metadata.created_at > sharedMetadataEntry.created_at;
```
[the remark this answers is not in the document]
seungpark: discussed offline. can bypass the merge process, and this check, if not an umbrella

seungpark (changes requested):
minor required changes below.

also, getting this error and im suspecting i've had too many open connections...

```
{
  errno: undefined,
  code: 'ETIMEOUT',
  syscall: 'querySrv',
  hostname: '_mongodb._tcp.cluster0.ylwlz.mongodb.net'
}
```
may need to close client connections on success/fail of app wrapper

modules/persistence/src/services/metadata/associated_products/index.ts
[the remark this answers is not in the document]
casthewiz: Looking further, I need to remove this entirely - I'm already checking and assigning via this statement `const umbrellaMetadata = hasAssociations(metadata) ? metadata : await umbrellaMetadataEntry(metadata);` for if the local entry is the umbrella entry 😵‍💫

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -135,21 +133,21 @@ const getAssociatedProducts = async (metadata, sharedMetadataEntry) => {
...
   // If we want to embed with staging builds, then this needs to be turned off
   // or converted so that local metadata ToC is added to tocInsertions
   const isStagingBranch = await !getRepoBranchesEntry(project, branch);
   if (isStagingBranch) return;
 
-  const repoBranchesEntries = await getAllAssociatedRepoBranchesEntries(sharedMetadata);
+  const repoBranchesEntries = await getAllAssociatedRepoBranchesEntries(umbrellaMetadata);
   const repoBranchesMap = mapRepoBranches(repoBranchesEntries);
-  const tocsCursor = await getAssociatedProducts(metadata, sharedMetadata);
+  const tocsCursor = await getAssociatedProducts(metadata, umbrellaMetadata);
```
seungpark, line 148: minor. have one ordered param here for getAssociatedProducts

modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  // Options might be undefined, so safely cast to {} if nullish
+  node.options = node.options ?? {};
+
+  const project = tocs[node?.options?.project];
+  const branches = Object.keys(project);
+  node.options.versions = branches;
+  node.children = branches.map((branch) => {
+    const child = project[branch];
+    const options = {
+      ...child[0].options,
```
seungpark, line 43: we are using index numbers for an object here (child[0] is throwing an error)

```
tocs is in form of tocInsertions:
{
  project: {
    branchName: ToC
  }
}
```

so a child is a ToC is an object, not a list. not sure exactly what the intent is here

seungpark (approved):
changes LGTM. last minor changes requested below for mergeNode 

i am still getting the db connection error Error at db client connection: Error: queryTxt ETIMEOUT cluster0.ylwlz.mongodb.net

wondering if it is because i wrapped my invocation of app with a timeout (for debugging purposes) - update: used VPN and this seems to have mitigated the issue

modules/persistence/src/services/metadata/associated_products/index.ts
casthewiz: Taken care of 💯

PR #704, opened 2022-10-26
DOP-3338: remove old search manifest generation step from Makefiles

schmalliso:
As of earlier this year, search manifest generation is handled by manifest build jobs rather than as a step in the Makefile. This has proven efficient and effective and stable, so we should remove this extraneous build step. Moreover, continued support of this build stage (which serves no purpose) is blocking progress on updating `mut` for manifest jobs.

The analogous changes have [already happened in cloud-docs-osb for testing purposes](https://github.com/mongodb/docs-worker-pool/commit/541ab6e810467a6cda2e772a1b7f0121a537f999), which behaved as expected ([staging link](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63596a79dafcf7c33d890275)). 

Given that Makefile changes are scary, though, extra attention to review detail would be appreciated :)

### DIFF
--- a/makefiles/Makefile.cloud-docs
+++ b/makefiles/Makefile.cloud-docs
@@ -14,21 +14,13 @@ PROJECT=cloud-docs
 MUT_PREFIX ?= $(PROJECT)
 REPO_DIR=$(shell pwd)
 
-CUSTOM_SEARCH_INDEX=true
-
 SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
 SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
 
 include ~/shared.mk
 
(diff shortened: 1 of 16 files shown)


i80and (approved):
Oh this pleases me mightily!

casthewiz (approved):
Just asking for some clarification on K8's, Ops + Cloud manager build/deploy step removal - but otherwise very happy that `##THESE ARE TERRIBLE HACKS THAT WE SHOULD BE (AND ARE) ASHAMED OF` are getting removed!

schmalliso:
@casthewiz We don't support deploying any of the old gen properties via the autobuilder so it seems weird to have steps in their Makefiles that strongly imply we would support that… so mostly for clarity for inbound folks / folks who aren't intimately familiar to not get the wrong impression of expected behavior.

PR #705, opened 2022-10-26
DOP-3304, DOP-3335, DOP-3359: update manifest job handler to use snooty manifest

schmalliso:


Dockerfile
```diff
@@ -34,7 +34,9 @@ RUN /venv/bin/pip install -r https://raw.githubusercontent.com/mongodb/docs-tool
 RUN apt-get update && apt-get install -y python3 python3-dev python3-pip
 RUN apt-get -y install vim
 RUN apt-get -y install git pkg-config libxml2-dev
-RUN python3 -m pip install -r https://raw.githubusercontent.com/mongodb/mut/master/requirements.txt
+RUN python3 -m pip install https://github.com/mongodb/mut/releases/download/v0.10.1/mut-0.10.1-py3-none-any.whl
```
i80and, line 37: Tasty.

There's the theoretical issue that we're no longer pinning dependencies, but I don't think we need to overthink this.

i80and (approved):

Dockerfile
schmalliso: Are they not semi-pinned in [pyproject.toml](https://github.com/mongodb/mut/blob/master/pyproject.toml) ?

i80and: that's a dependency specification, not pinning per se. The idea of pinning is to express "developers have tested *this exact* dependency version and it is known to be good", preventing a regression in a dependency from hosing things.

schmalliso:
@i80and Rerequesting since I've touched a bunch of other code since your original LGTM as I found _opportunities_ for enhancement and also bugs.

api/controllers/v1/slack.ts
```diff
@@ -140,9 +140,11 @@ export const getDeployableJobs = async (values, entitlement, branchRepository: B
       urlSlug,
       false,
       false,
-      '-g'
+      false
```
i80and, line 143: Do I want to know? 😅

api/controllers/v1/slack.ts
```diff
@@ -140,9 +140,11 @@ export const getDeployableJobs = async (values, entitlement, branchRepository: B
...
     );
 
+    newPayload.stable = isStableBranch ? true : false;
```
i80and, line 146: I think the `!!isStableBranch` incantation is idiomatic for this. or `new Boolean(isStableBranch)`.

api/controllers/v1/slack.ts
schmalliso: You do not.

api/controllers/v1/slack.ts
schmalliso: ooh nice. done.

schmalliso:
Checks passed on the non-forked Draft. https://github.com/mongodb/docs-worker-pool/pull/711 Awaiting Cassidy feedback before I click the 'Merge without waiting for requirements' button like a champ.

casthewiz (approved):
LGTM - I'm increasingly questioning the utility of the `createPayload` function in the slack controller, but that's a revision for another PR imo.

PR #706, opened 2022-10-27
DOP-3335: separate parse and front-end stages in makefiles

schmalliso:
We don't need to build the front-end for manifestJobs anymore, so might as well separate it into two steps in the Makefile so we're not needlessly building a front-end each deploy.


i80and (approved):
This idea is clutch I love it :+1:

PR #714, opened 2022-11-17
DOP-3340: Initial implementation of OAS page builder module

rayangler:
### Ticket

[DOP-3340](https://jira.mongodb.org/browse/DOP-3340)

### Testing

Follow the steps below if you're interested in testing locally. It's also advised to read the README added in this PR for additional context and/or instructions.

1) Use the parser (with the [following commit](https://github.com/mongodb/snooty-parser/commit/144386d522c8693dc4a02424ef7786ed0bd133ff)) to parse a docs repo that has one or more OpenAPI content pages. Examples include `cloud-docs` and `docs-app-services`. (Using the aforementioned commit instead of the `HEAD` is important here due to changes and experimentation of ideal metadata outputs. The module will not work if used with the latest commits.)
2) Clone the team's [fork of Redoc](https://github.com/mongodb-forks/redoc). Set up by installing dependencies for both the root folder (Redoc) and within the `cli/` folder (Redoc CLI).
3) Build and compile Redoc and Redoc CLI by running `npm run bundle` in the root directory of the Redoc fork. IF there are any issues building, then you might need to use `npm link ../` to link the local copy of `redoc` as the CLI's dependency. `npm run bundle` should include a build for the CLI, which will be output as `cli/index.js`.
4) (Optional) Use the frontend to build the docs repo you parsed. All files should be built to the `public/` directory. The autobuilder module should output the generated HTML files from Redoc to the same `public/` directory output from Snooty. However, for testing purposes, you could always just have it output to wherever you want.
5) Install the dependencies of this autobuilder module and run using the following command:

```
npm run build && npm run start -- --bundle <path-to-zipped-output-from-parser> --output <path-to-target-directory> --redoc <path-to-local-js-file-for-redoc-cli> --repo <path-to-docs-repo-directory-you-parsed>
```

### Notes

* Adds initial implementation for the new OpenAPI page builder module.
* Not yet callable or usable by the autobuilder; this functionality should be added in [DOP-3341](https://jira.mongodb.org/browse/DOP-3341). The module is usable locally for testing purposes (see: testing instructions above). If having difficulties with testing locally and/or if the instructions are unclear/inaccurate, please feel free to ping me!


modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+
+      const finalFilename = normalizePath(`${output}/${pageSlug}/index.html`);
+      await execRedoc(spec, finalFilename, redocPath);
+    } catch (e) {
+      console.error(e);
+      // Continue to try to build other pages since it's possible that mut will
+      // still upload existing HTML files
+      continue;
+    }
```
rayangler, line 80: Currently catching here and continuing because it's unclear what the interaction will be when this module is called by the autobuilder. There have been instances in the past where a failure in the autobuilder still resulted in the build continuing... so having pages continue to build under this condition to minimize missing OpenAPI content pages might be okay.

Happy to test this interaction as part of DOP-3341 and change this to throw

modules/oas-page-builder/src/services/database.ts
```diff
@@ -0,0 +1,47 @@
...
+    try {
+      await client.connect();
+      dbInstance = client.db(DB_NAME);
+    } catch (error) {
+      console.error(`Error at db client connection: ${error}`);
+      throw error;
+    }
+  }
+  return dbInstance;
+};
```
rayangler, line 32: This logic is the same as the persistence module's. I wanted a way to sort of share the logic between the two, but might be difficult if we wanted to treat the 2 as independent modules

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+      console.log(`Using ${oasFileURL}`);
+    } else {
+      throw new Error(`Could not find a saved hash for API: ${apiKeyword}`);
+    }
+  }
+
+  return oasFileURL;
+};
+
+export const buildOpenAPIPages = async (
```
rayangler, line 52: Most of the logic in this file is similar to the logic introduced in https://github.com/mongodb/snooty/pull/719, but modified for the autobuilder's use case

rayangler (commented):
Annotations

modules/oas-page-builder/src/services/database.ts
casthewiz: 🤔 it might make sense to manage this dependency in a shared lib in the future, but I think this is sufficient for now. We can always revisit if we add a 3rd module and feel that the logic is starting to get copy-pasted. 😄

modules/oas-page-builder/src/services/pageBuilder.ts
casthewiz: I think catch, log the error, and continue is the best course of action here. Even if building the OAS file fails, I think we need the remainder of the build to continue - especially since I could see a case where a content team member wants to stage content, and an error in the OAS build step shouldn't impact that process.

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const getOASFileUrl = (gitHash: string) =>
+  `https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/${gitHash}.json`;
+
+const getAtlasSpecUrl = async (apiKeyword: string) => {
+  // Currently, the only expected API fetched programmatically is the Cloud Admin API,
+  // but it's possible to have more in the future with varying processes.
+  if (apiKeyword !== 'cloud') {
+    throw new Error(`${apiKeyword} is not a supported API for building.`);
+  }
```
casthewiz, line 25: minor nitpick: if we want more than one keyword, may be useful to follow the
````
const keywords = ['cloud'];
if keywords.includes(apiKeyword)
```

pattern

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -0,0 +1,82 @@
...
+  const res = await fetch(url);
+  if (!res.ok) {
+    // Error should be caught when creating pages.
+    throw new Error(`${errMsg}; ${res.statusText}`);
+  }
+  return res.text();
+};
+
+const getOASFileUrl = (gitHash: string) =>
+  `https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/${gitHash}.json`;
```
casthewiz, line 18: We might want to consider parameterizing, or at least storing the url for this string this in a constant e.g. `const OAS_FILE_SERVER = https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/`

casthewiz (approved):
This is looking good, just a few nits that don't need to be strictly addressed in this pr

modules/oas-page-builder/src/services/pageBuilder.ts
rayangler: Done. Leaving as a constant for now. We should be able to easily parameterize it in the future depending on what URLs we need

PR #722, opened 2022-12-14
DOP-3370

seungpark:
DOP-3370: Construct unit / snapshot tests to modules within Persistence Module.

utilizes [jest-mongodb](https://github.com/shelfio/jest-mongodb) to run a local MongoDB memory server

- [x] add mock data to represent samples of prod `metadata` and `repos_branches` collections
- [x] snapshots to cover associated_products module
- [x] unit tests to cover connector module
- [x] convert above to snapshots(SEE TODO in connector.test.ts)
- [x] snapshots to cover toc module


modules/persistence/tests/services/connector.test.ts
```diff
@@ -0,0 +1,140 @@
...
+    });
+
+    test('returns an error if client fails', async () => {
+      /**
+       * Question for error catching:
+       * why are they not catching in functions (ie. db, insert) when throwing mocked rejections
+       * see connection/db line 32
+       *
+       */
+      mockConnect.mockRejectedValueOnce(new Error('test error') as never);
```
seungpark, line 86: @casthewiz i was not seeing this rejected value being logged within [connector ](https://github.com/mongodb/docs-worker-pool/blob/master/modules/persistence/src/services/connector/index.ts#L33)

upsert and insert are written differently (throw in catch vs non) but the `return Promise` seems to bypass the catch block? 

```
try {
  return await insertSession
      .collection(collection)
      .insertMany(docs.map((d) => ({ ...d, build_id: buildId, created_at: buildId.getTimestamp() })));
} catch (error) {
   // error will be caught here and propagated
}

casthewiz: Looks fine to me - we're pretty deep in nested awaits here, so makes sense to preserve error context by `return await`. Good change!

casthewiz (approved):

PR #723, opened 2022-12-15
DOP-3341: Integrate OAS Page Builder module as part of build process

rayangler:
### Ticket

[DOP-3341](https://jira.mongodb.org/browse/DOP-3341)

### Current Behavior

[Atlas Cloud Admin API](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/)
[Atlas App Services API](https://www.mongodb.com/docs/atlas/app-services/admin/api/v3/)
[Landing OpenAPI Preview](https://www.mongodb.com/docs/openapi/preview/)
[BI Connector](https://www.mongodb.com/docs/bi-connector/current/)

### Staging Links

The following staging links are built with our fork of Redoc. Notable differences:
* The UnifiedNav component stays at the top of the page instead of being sticky/fixed when scrolling down the page. There is custom logic on our frontend that accomplishes this interaction, but has not yet been applied to our fork of Redoc for simplicity. We can go back and apply this logic in the fork, if desired
* Missing Feedback Widget since it's currently strictly a frontend component

[Atlas Cloud Admin API](https://docs-atlas-dotcomstg.s3.us-east-2.amazonaws.com/docs-qa/atlas/reference/api-resources-spec/index.html) (S3 link since docs-qa link is messed up) - using latest git hash at the time
[Atlas Cloud Admin API](https://docs-atlas-dotcomstg.s3.us-east-2.amazonaws.com/cloud-docs/docsworker-xlarge/2023-01-03-test-staging/reference/api-resources-spec/index.html) - using staging webhook
[Atlas App Services API](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/atlas/app-services/admin/api/v3/) - using local spec file from docs repo
[Landing OpenAPI Preview](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/openapi/preview/) - should have no change
[BI Connector](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/bi-connector/current/) - should have no change; does NOT use OpenAPI

### Notes

* Adds build step for building OpenAPI content pages through the Autobuilder.
* Adds build steps for OAS Page Builder module in Dockerfile.
* Adds installation steps for Redoc (CLI) fork in Dockerfile.
   - This adds about 6-10 minutes to the Docker image build time. This range is based on build logs + local Docker image build times.
   - This additional build time could potentially be reduced if we install the package/module as a global CLI tool through Artifactory. We should explore this option in the future.
* Changes env variables for database access in the OAS Page Builder module to better fit existing production variables needed by the autobuilder. 
   - This was tested as part of the Atlas Cloud Admin API staging link. See [build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=639b7b815fc5aca436fd443e) - "Using https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/a646b0cf64fe70b213d206768900eb5e4982a476.json". An invalid git hash was purposely used for testing to forcefully trigger the logic for fetching the last good git hash for the Cloud API's spec file.
   - Logs are out of order probably because of stdout vs. stderr?
* Tested with`/test-deploy` on repos with OpenAPI pages. I've already test deployed `cloud-docs`, `docs-app-services`, and `docs-landing` (which should have no change in behavior)
* Tested with pre-prd's writers staging via webhook.

### TODO before merging
- [x] Remove this branch from preprd/integration/stg image build
- [x] Update Dockerfile's parser version. The current version of the parser generates OpenAPI metadata that is only compatible with the [frontend implementation](https://github.com/mongodb/snooty-parser/pull/428). This change should be reverted and should be accompanied by a new parser version
- [x] Clean up branch names of cloned and depended Redoc branches
- [x] Test with preprd staging webhook?
- [x] Update `docs-landing`'s OpenAPI preview page to use the `:preview:` option to avoid static rendering. The frontend should continue to handle building the preview page since it already supports on-demand spec file parsing.

### Open Questions

* Should we wrap the call to `make oas-page-build` in a feature flag to allow for more time testing in preprd/stg/integration?


mmeigs (approved):
LGTM. 
All of this looks good! I am guessing you're aware of the slight styling differences between the current versions and your staging links, though? Although that has nothing to do with this PR :)

rayangler:
> the slight styling differences between the current versions and your staging links

Yup, some of the styling differences are intentional, such as menu items being narrower/having different margins and padding, badges having different border radius, etc. I think I need to go back to fix certain font sizes to make them match what's in production. I'll do a styling audit on the differences

But if any other differences stand out and/or seem like they may be blockers for merging, let me know!

mmeigs:
> But if any other differences stand out and/or seem like they may be blockers for merging, let me know!

Gotcha! None of these are blockers for this PR imo. The only other one to mention is the highlight of the tab in the Unified Nav. But you seem to already be aware of all of them. I was just curious!

Dockerfile
```diff
@@ -79,10 +90,32 @@ RUN git clone --depth 1 https://github.com/mongodb/devhub.git snooty-devhub
     && cd snooty-devhub                                                                            \
     && npm install --production
 
+# install redoc fork
+RUN git clone -b ${REDOC_CLI_VERSION} --depth 1 https://github.com/mongodb-forks/redoc.git redoc \
```
seungpark, line 94: minor, should we put this URL as a var as well

modules/oas-page-builder/src/services/database.ts
```diff
@@ -1,15 +1,15 @@
 import { Db, MongoClient } from 'mongodb';
 import { OASFile, OASFileGitHash } from './models/OASFile';
 
-const DB_NAME = process.env.DB_NAME || 'snooty_dotcomprd';
+const DB_NAME = 'snooty_dotcomprd';
```
seungpark, line 4: minor, dont forget to revert here

src/job/stagingJobHandler.ts
```diff
@@ -57,6 +57,8 @@ export class StagingJobHandler extends JobHandler {
       this.currJob.buildCommands[this.currJob.buildCommands.length - 1] = 'make next-gen-html';
       if (this.currJob.payload.repoName === 'devhub-content-integration') {
         this.currJob.buildCommands[this.currJob.buildCommands.length - 1] += ` STRAPI_PUBLICATION_STATE=preview`;
+      } else {
```
seungpark, line 60: qq: is this to preview staging jobs with the client side build for openapi?

seungpark (approved):
LGTM. minor comments and questions below but overall looks good!

rayangler:
> the highlight of the tab in the Unified Nav.

Good catch! It seems like it doesn't highlight at all in the S3 links, but seems to work fine if the [staging link](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/atlas/app-services/admin/api/v3/) is actually part of the docs-qa site. I think the consistent nav has a regex that determines highlighting, and our S3 site just isn't included

Dockerfile
rayangler: We could. The URL is a lot less likely to change than the Redoc CLI version though. I was also just following conventions for our other git repo clones haha

modules/oas-page-builder/src/services/database.ts
rayangler: This change is intentional. `DB_NAME` in the Autobuilder seems to be tied to either `pool` or `pool_test`, but we need a `snooty_*` database since this is where the OAS-related data is currently located. For the sake of simplicity, I think it makes sense to tie it down to the database and collection we actually need for now.

(Although in the future/long-term, we should probably move this OAS-related collection to be in `pool` or `pool_test`, or some other db entirely, as OpenAPI support evolves)

src/job/stagingJobHandler.ts
rayangler: Yup, the build commands here are for staging jobs. Although the `oas-page-build` command here would be to also build things statically, as opposed to client-side. The only page to be built client-side should be the preview page for `docs-landing`

modules/oas-page-builder/src/services/database.ts
seungpark: wouldn't this result in test data getting into prod docs? also did not see the DB_NAME in env files, idk where `pool` would be coming from..

rayangler: `snooty_dotcomprd` is the same database that the frontend pulls OpenAPI data from to build the Cloud Admin API on the client-side. There shouldn't be any test data present in that database.

As for the `DB_NAME`, I think it was being declared/passed down from the [ECS container definitions](https://github.com/mongodb/docs-worker-pool/blob/8aab9e6afd765fff1af0e5743e3077434a0a4bad/infrastructure/ecs-main/ecs_service.yml#L113-L114).

rayangler: Just a heads up that I added the logic to use the Autobuilder's `SNOOTY_ENV` env var in order to allow us to access either `snooty_dotcomstg.oas_files` or `snooty_dotcomprd.oas_files`. 👍

casthewiz (approved):

Dockerfile
```diff
@@ -79,10 +90,32 @@ RUN git clone --depth 1 https://github.com/mongodb/devhub.git snooty-devhub
...
 COPY --from=ts-compiler /home/docsworker-xlarge/config config/
 COPY --from=ts-compiler /home/docsworker-xlarge/build ./
 RUN npm install
+
+# OAS Page Builder module copy
+# Create directory and add permissions to allow node module installation
+RUN mkdir -p modules/oas-page-builder && chmod 755 modules/oas-page-builder
+COPY --from=ts-compiler /home/docsworker-xlarge/modules/oas-page-builder/package*.json ./modules/oas-page-builder/
+COPY --from=ts-compiler /home/docsworker-xlarge/modules/oas-page-builder/dist ./modules/oas-page-builder/
+RUN cd ./modules/oas-page-builder/ && npm install
```
seungpark, line 113: is the installation necessary again? see that we are installing in line 13-15?

also see from [OAS readme](https://github.com/mongodb/docs-worker-pool/blob/master/modules/oas-page-builder/README.md#running) that we should be using npm commands

seungpark: oh i see, previously installing and building in original directory, then transferring from repo dir to home dir? do we have to install the first time around then?

rayangler: > also see from [OAS readme](https://github.com/mongodb/docs-worker-pool/blob/master/modules/oas-page-builder/README.md#running) that we should be using npm commands

I'm a little confused by which npm commands you're referring to. Running with `npm start` isn't really applicable in this case with building.

> previously installing and building in original directory, then transferring from repo dir to home dir?

More or less. We're building the module in the `ts-compiler` stage, and then copying only the needed files over from the modules directory of that stage to the modules directory our current/final stage.

> do we have to install the first time around then?

I'm actually unsure about this. The first installation ensures that anything needed to successfully build the module is installed. Testing locally, it seems like it might be okay to leave out the first installation since we don't really have anything special about the build step, and `tsc` should already be installed. Will push up a new commit later

seungpark: thanks @rayangler !

PR #725, opened 2022-12-19
DOP-3414: Add back button support for OAS Page Builder

rayangler:
### Ticket

[DOP-3414](https://jira.mongodb.org/browse/DOP-3414)

### Notes
* Adds new required option to OAS Page Builder module build command for site's URL. This will be used as the Redoc fork's `backNavigationPath` option
* Introduces logic for pulling the site's title from the bundle/build's metadata document and passes that to the Redoc fork as the `siteTitle` option
* Refactors the `commandExecutor` service function to be the `RedocExecutor` class to help avoid repeated operations
* Related [Redoc fork PR for back button](https://github.com/mongodb-forks/redoc/pull/3), for reference


casthewiz (approved):
Nit - it feels a little awkward to create a `new RedocExecutor` per entry in the `buildOpenAPIPages` function, since the `siteUrl, siteTitle` is the same per iteration. 

If we're doing a class based approach for the `RedocExecutor` to manage state, it might make more sense to just mutate the `redocPath` property.

rayangler:
@casthewiz Thanks for catching that! The `redocExecutor` should have been declared as a constant outside of the `for` loop when building the pages. Does the latest change look good?

casthewiz (approved):
@rayangler It looks good!

PR #726, opened 2022-12-20
DOP-3420

seungpark:
DOP-3420: configure github actions to run for respective file changes

- Scope all existing tests (jest config) and linter (eslint config) to be scoped to worker pool (ie. exclude modules/**)
- add [github action filter](https://github.com/mongodb/docs-worker-pool/actions/runs/3842190930/jobs/6543242295) to run test and linter on changes to anything other than `/modules` directory. the purpose here is to run the subsequent jobs of action if the filter passes. see action link and note
```
Results:
Filter autobuilder = true
  Matching files:
  .eslintrc.js [modified]
  .github/workflows/test-persistence.yml [added]
  .github/workflows/test.yml [modified]
```
- update lint and jest configs for persistence module, invoked by changes to `/modules/persistence` directory, using similar filter as above


.github/workflows/test.yml
```diff
@@ -17,7 +35,7 @@ jobs:
         node-version: '14.x'
     - name: Install dependencies
       run: npm install --dev
-    - name: Lint
+    - name: Lint # TODO: set linter to be scoped to not modules
```
rayangler, line 38: Is this TODO still needed? I see the ignore pattern was added in the `.eslintrc`

rayangler (approved):
I think usage of the paths filter looks good to me!

We could potentially combine the CI workflows to just be separate jobs in the same CI workflow, to avoid needing separate `changes` jobs. But separation might be better for modularity in workflows for now 🤔

.github/workflows/test.yml
seungpark: good catch! removed the comment tyty

seungpark:
> I think usage of the paths filter looks good to me!
> 
> We could potentially combine the tests in the different CI workflows to just be separate jobs in the same CI workflow, to avoid needing separate `changes` jobs. But separation might be better for modularity in workflows for now 🤔

i like the modularity and repeated patterns for now.... maybe we can wrap the workflows with a `per module subdirectory, run install, lint, format, test`, and we would just have to conform to those npm script commands 👍

PR #733, opened 2023-01-05
DOP-3439: use checkout for merge changes

seungpark:
cc @rayangler the master branch is failing at CI steps due to no checkout being invoked.
we'll need to merge and test this since merge and PR actions are separate


rayangler (approved):
Thanks for catching! I forgot that behavior might be different between PRs and merges

PR #734, opened 2023-01-06
DOP-3412: Multiple metadata files with Urlified ToCNodes

casthewiz:


modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -109,26 +121,27 @@ const getAssociatedProducts = async (umbrellaMetadata) => {
...
     // This query matches on projects in associated_products for our given metadata that have a build_id and that do not have merged tocs
+    // then sorts by build_id for all matches
     // then groups per branch and per project from those matches
-    // and gets the most recent doc entry (by build_id), with the toctree and toctreeOrder fields.
-    const tocs = snooty.collection('metadata').aggregate([
+    // and gets the first metadata entry for each group
+    // since we are sorted by build_id, we get the most recent entry.
+    const associatedMetadataEntries = snooty.collection('metadata').aggregate([
       {
         $match: { project: { $in: associatedProductNames }, build_id: { $exists: true }, is_merged_toc: { $ne: true } },
```
seungpark, line 130: `build_id: $exists: true`

qq: meaning that both umbrella + associated products have to be persisted at least once to have the merged ToC available

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -147,18 +160,45 @@ export const mergeAssociatedToCs = async (metadata) => {
...
+    repoBranchesMap
+  );
 
-  return mergedMetadataEntry;
+  // We need to have copies of the main umbrella product's ToC here, to handle multiple metadata entry support
+  const umbrellaToCs = {
+    original: copyToCTree(umbrellaMetadata.toctree),
+    urlified: copyToCTree(
+      umbrellaMetadata.toctree,
+      umbrellaMetadata.url,
```
seungpark, line 181: passing nonexistant url here instead of `project`

modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -30,15 +38,16 @@ const isInsertionCandidateNode = (node: ToC, associated_products: AssociatedProd
...
-const mergeNode = (node: ToC, tocs: ToCInsertions) => {
+const mergeNode = (node: ToC, tocs: ToCInsertions, currentProject) => {
   // Options might be undefined, so safely cast to {} if nullish
   node.options = node.options ?? {};
+  const needsUrlifiedToC = currentProject !== node?.options?.project;
 
-  const project = tocs[node?.options?.project];
-  const branches = Object.keys(project);
+  const associatedProject = tocs[node?.options?.project];
+  const branches = Object.keys(associatedProject);
```
seungpark, line 47: when testing with an umbrella entry that does not have its associated metadata entry persisted yet, getting error for `VM506:1 Uncaught TypeError: Cannot convert undefined or null to object`

seungpark (changes requested):
minor changes requested below. still going through the logic, and testing cases ie.

- umbrella / associated metadata doesnt exist, persisting umbrella metadata
- umbrella / associated metadata doesnt exist, persisting associated metadata
- umbrella metadata exists, persisting umbrella
- umbrella metadata exists, persisting associated
- same for associated

modules/persistence/src/services/metadata/associated_products/index.ts
casthewiz: Yep - otherwise the module doesn't have any data to source.

modules/persistence/src/services/metadata/ToC/index.ts
casthewiz: Good catch here - I should just short circuit here in general and raise a log if we can't find a branches object.

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -147,18 +160,45 @@ export const mergeAssociatedToCs = async (metadata) => {
...
+  const umbrellaRepoBranchesEntry = await getRepoBranchesEntry(umbrellaMetadata.project, umbrellaMetadata.branch);
+  if (!umbrellaRepoBranchesEntry)
+    throw `No repoBranches entry available for umbrella metadata with project: ${umbrellaMetadata.project}, branch: ${umbrellaMetadata.branch}`;
+
   const repoBranchesEntries = await getAllAssociatedRepoBranchesEntries(umbrellaMetadata);
   const repoBranchesMap = mapRepoBranches(repoBranchesEntries);
-  const tocsCursor = await getAssociatedProducts(umbrellaMetadata);
-  const { tocInsertions, tocOrderInsertions } = await shapeToCsCursor(tocsCursor, repoBranchesMap);
-  const mergedMetadataEntry = traverseAndMerge(umbrellaMetadata, tocInsertions, tocOrderInsertions);
+  const metadataCursor = await getAssociatedProducts(umbrellaMetadata);
```
seungpark, line 169: minor nitpick. combination of `getAssociatedProducts` and `shapeToCsCursor` can be cleaned up here a bit.

`getAssociatedProducts` can iterate over the cursor and return an object in the same format as mapRepoBranches {project: {branch: metadata}}, and `shapeToCsCursor` can read from this format and be agnostic of keys, ie. `most_recent`

modules/persistence/src/services/metadata/ToC/utils/prefixFromEnvironment.ts
```diff
@@ -0,0 +1,9 @@
+import { BranchEntry } from '../../associated_products';
+
+export const prefixFromEnvironment = (repoBranchEntry: BranchEntry) => {
```
seungpark, line 3: `BranchEntry` does not have specified keys `url` and `prefix`

also passing in a repoBranchEntry from db, which also does not have the two attributes.

referring to this comment, we should separate logic for pulling url and prefix info, apart from {project-name: {branch-name: repoBranch}}

casthewiz: Hmmm, I've made the mistake of conflating the `repos_branches.branches` object vs. the parent `repos_branches` entry - I'll need to adjust the typing here - I'll think about if we need to generalize pulling info for the `url` and `prefix`, I'm mostly convinced but need to reason about when/where that responsibility should be

seungpark: sorry, linked the previous comment (forgot the link before)

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -52,54 +53,65 @@ const getAllAssociatedRepoBranchesEntries = async (metadata: SharedMetadata) =>
 const mapRepoBranches = (repoBranches: ReposBranchesDocument[]) =>
   Object.fromEntries(
     repoBranches.map((entry) => {
-      const branches = Object.fromEntries(entry.branches.map((branch) => [branch.gitBranchName, branch]));
-      return [entry.project, branches];
+      const { url, prefix } = entry;
+      const branches = Object.fromEntries(entry.branches.map((branch) => [branch.gitBranchName, { ...branch }]));
+      return [entry.project, { ...branches, url, prefix }];
```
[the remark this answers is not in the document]
casthewiz: Resolving since we talked about this via Zoom - fine to iterate through here as is.

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -52,54 +61,65 @@ const getAllAssociatedRepoBranchesEntries = async (metadata: SharedMetadata) =>
...
     // Initialize to empty object if we haven't already, for a given project.
     if (!tocInsertions[doc._id.project]) tocInsertions[doc._id.project] = {};
     if (!tocOrderInsertions[doc._id.project]) tocOrderInsertions[doc._id.project] = {};
 
+    const repoBranchesEntry = repoBranchesMap?.[doc._id.project]?.[doc._id.branch];
     // TODO: If we want staging builds with embedded versions, it needs to be added here
-    if (repoBranchesMap?.[doc._id.project]?.[doc._id.branch]) {
-      tocInsertions[doc._id.project][doc._id.branch] = doc.most_recent.toctree;
+    if (repoBranchesEntry) {
+      const { url, prefix } = prefixFromEnvironment(repoBranchesEntry);
```
seungpark, line 111: discussed offline, leaving here for note. passing in `BranchEntry` here when expecting `ReposBranchesDocument`. can be optimized to `projectInfo` vs `branchInfo`

modules/persistence/src/services/metadata/ToC/utils/prefixFromEnvironment.ts
```diff
@@ -0,0 +1,9 @@
+import { ReposBranchesDocument } from '../../associated_products';
+
+export const prefixFromEnvironment = (repoBranchEntry: ReposBranchesDocument) => {
+  const env = process.env.SNOOTY_ENV ?? 'dotcomprd';
```
seungpark, line 4: minor question. why is dotcomprd default prefix setting?

modules/persistence/src/services/metadata/ToC/utils/convertSlugToUrl.ts
```diff
@@ -0,0 +1,11 @@
+export const convertSlugToUrl = (slug, projectSlug, prefix, url) => {
+  return ensureTrailingSlash(url) + ensureTrailingSlash(prefix) + projectSlug + ensureLeadingSlash(slug);
```
seungpark, line 2: does the prefix not already contain the projectSlug?

ie. 
image

(also on minor note above looks off? prefix.dotcomstg for pool.repos_branches )

seungpark (changes requested):

modules/persistence/src/services/metadata/associated_products/index.ts
casthewiz: I've tinkered with this a bit and `shapeToCsCursor` can be agnostic of `most_recent` and `_id`, but the overall format of the object from `getAssociatedProducts` shouldn't match `mapRepoBranches` output, since we basically just want to match on entries that have matching keys for the `project` and `branch` values for each entry in the cursor. Making `project` and `branch` be the actual keys makes this harder - (and we'll still need to iterate over each entry from the aggregation anyways)

If we move to `{project: {branch: metadata}}`, we can pull the keys out by invoking `Object.keys` twice, but this seems redundant - or I guess we could do some fancy `...` logic, but I'm having a hard time coming up with what that would look like off the top of my head (and thus suspect that it would be non-obvious to other contributors what the codeblock is doing)

I've instead opted to destructure the bits we care about out of the `doc` and I've added a new interface to buff up the typing here a bit.

If I'm misunderstanding the suggestion, please flag so I can understand better 😄

modules/persistence/src/services/metadata/ToC/utils/convertSlugToUrl.ts
casthewiz: 🤦‍♂️ very good catch

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -109,56 +130,89 @@ const getAssociatedProducts = async (umbrellaMetadata) => {
...
+    const repoBranchesMap = mapRepoBranches(repoBranchesEntries);
+    const metadataCursor = await getAssociatedProducts(umbrellaMetadata);
+
+    const { tocInsertions, tocOrderInsertions, associatedMetadataEntries } = await shapeToCsCursor(
+      metadataCursor,
+      repoBranchesMap
+    );
+
+    // We need to have copies of the main umbrella product's ToC here, to handle multiple metadata entry support
+    const umbrellaToCs = {
```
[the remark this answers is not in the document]
casthewiz: 🤦‍♂️ very good catch

modules/persistence/src/services/metadata/associated_products/index.ts
seungpark: discussed offline. can optimize in future but not necessary ATM

modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -49,33 +59,60 @@ const mergeNode = (node: ToC, tocs: ToCInsertions) => {
...
+// Create a deep copy of a ToC, converting all slugs present to absolute urls if project, prefix and url is provided.
+// Copy logic should be tightly coupled to the urlification logic here - we DON'T want to mutate the base ToC objects.
+export const copyToCTree = (toBeCopied: ToC, prefix?: string, url?: string): ToC => {
+  const toctree = JSON.parse(JSON.stringify(toBeCopied));
+  if (!prefix || !url) return toctree;
+  let queue = [toctree];
+  while (queue?.length) {
+    const next = queue.shift();
+    if (next && next.slug) {
+      next.url = convertSlugToUrl(next.slug, project, prefix, url);
```
seungpark, line 108: remove project here as well

casthewiz: classic vscode save conflict, I've rectified and pushed 👍

casthewiz:
Just need to fix some of the failing tests but this should be in a good spot for another look.

modules/persistence/src/services/metadata/ToC/utils/prefixFromEnvironment.ts
casthewiz: Safest failure option in the event that a production job run encounters a nullish `SNOOTY_ENV` value.

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -52,54 +69,68 @@ const getAllAssociatedRepoBranchesEntries = async (metadata: SharedMetadata) =>
...
+    if (!tocOrderInsertions[project]) tocOrderInsertions[project] = {};
 
+    const projectEntry = repoBranchesMap?.[project];
+    const branchesEntry = projectEntry?.[branch];
     // TODO: If we want staging builds with embedded versions, it needs to be added here
-    if (repoBranchesMap?.[doc._id.project]?.[doc._id.branch]) {
-      tocInsertions[doc._id.project][doc._id.branch] = doc.most_recent.toctree;
-      tocOrderInsertions[doc._id.project][doc._id.branch] = doc.most_recent.toctreeOrder;
+    if (branchesEntry) {
+      const { url, prefix } = prefixFromEnvironment(branchesEntry);
```
seungpark, line 122: minor. projectEntry here

seungpark (approved):
this LGTM. but there exists the error of duplicate key from metadata collections. will be tackling this issue as https://jira.mongodb.org/browse/DOP-3467

modules/persistence/src/services/metadata/associated_products/index.ts
casthewiz: I think var names just need to be improved here - we have [project][branch], which is confusing since we've discussed `repoBranches` really being a "project" entry


but the [project][branch] object from the map has the url and prefix on it, from line 74

going to rename these and then merge

PR #735, opened 2023-01-08
DOP-2246: Remove publishedbranches from meta branch

i80and:
Corresponding commit to remove publishedbranches from main branch to follow, and both should be handled in lockstep.


makefiles/Makefile.docs-csfle-merge
```diff
@@ -14,8 +14,4 @@ REPO_DIR=$(shell pwd)
 SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
 SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
 
-include ~/shared.mk
-
-
-
-
+include ~/shared.mk
```
schmalliso, line 17: Do we need a newline at the end of a makefile?

schmalliso (commented):
One question but otherwise this all looks good to me!

makefiles/Makefile.docs-csfle-merge
i80and: oh, funky! It's probably fine, but I'd rather have one. Good catch!

schmalliso (approved):

PR #737, opened 2023-01-09
DOP-3377: install and run persistence module in staging instances

seungpark:
**goal:** PR is designed to install and call the persistence module from the autobuilder

**prereq:** while deploying on preprod, found a front end bug dealing with env vars<env vars are stored as strings but evaluated as boolean>. fix is [here](https://github.com/mongodb/snooty/pull/751).

**process**: 
- install persistence module to Docker image
- invoke said module as a fail-safe option in [makefiles](https://github.com/mongodb/docs-worker-pool/pull/738/files), post-parse and pre-front end build steps.
- fix env variables for persistence module according to autobuilder
- add front end feature flag to prevent front end from consuming persisted data

**tests:** tested with /test-deploy after deploy-stg workflows. these can be found in Atlas under `snooty_dotcomstg
.metadata` with the same project ie. `{project: 'landing'}` cc @casthewiz are we not sharing build ids between artifacts and build logs?
- [docs landing deployed with persistence module. errored on duplicated key in Atlas, continued to deploy](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63c1bc44c5cd20791001d2eb)
- [docs tutorials deployed](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63c1bf315af6b7eae85399b7)
- [realm docs deployed, failure not terminal due to ignore errors flag in makefile](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63c1bc44c5cd20791001d2ec)

TODO:
- [x] revert github workflow to build current branch on preprod
- [x] revert build step reference to test makefile branch
- [x] revert Dockerfile reference to test makefile branch

PR #738, opened 2023-01-09
DOP-3377: update makefiles to invoke persistence module

seungpark:
Invoke persistence module [installed via Docker images](https://github.com/mongodb/docs-worker-pool/pull/737) as a step after parser.

See test build logs in above PR.
[Errors are ignored and build steps continue after non terminal error from persistence module](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63c1bc44c5cd20791001d2ec)

### DIFF
--- a/makefiles/shared.mk
+++ b/makefiles/shared.mk
@@ -3,7 +3,6 @@ SNOOTY_ENV = $(shell printenv SNOOTY_ENV)
 REGRESSION = $(shell printenv REGRESSION)
 BUCKET = $(shell printenv BUCKET)
 URL = $(shell printenv URL)
-USE_PERSISTENCE = $(shell printenv USE_PERSISTENCE)
 
 # "PATCH_ID" related shell commands to manage commitless builds
 PATCH_FILE="myPatch.patch"
--- a/makefiles/shared.mk
+++ b/makefiles/shared.mk
@@ -52,7 +52,7 @@ next-gen-parse:
 next-gen-html: next-gen-parse
 	# persistence module - add bundle zip to Atlas documents
 	# ignore errors "-" flag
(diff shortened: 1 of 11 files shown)


makefiles/shared.mk
```diff
@@ -81,4 +81,8 @@ oas-page-build:
 	node ${OAS_MODULE_PATH} --bundle ${BUNDLE_PATH} --output ${REPO_DIR}/public --redoc ${REDOC_PATH} --repo ${REPO_DIR} --site-url ${URL}/${MUT_PREFIX}
 
 persist-data:
-	node ${PERSISTENCE_MODULE_PATH} --path ${BUNDLE_PATH}
\ No newline at end of file
+	ifeq ($(USE_PERSISTENCE), true)
+		-node ${PERSISTENCE_MODULE_PATH} --path ${BUNDLE_PATH}
```
rayangler, line 85: Should this be `-node` or `node`?

seungpark: this was for [ignoring errors ](https://www.gnu.org/software/make/manual/html_node/Errors.html#:~:text=To%20ignore%20errors%20in%20a,to%20the%20shell%20for%20execution.&text=This%20causes%20make%20to%20continue,unable%20to%20remove%20a%20file) but will double check for execution

seungpark: @rayangler  https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63bdba5eb9c1638e2c81c235

this is from a bad commit, but the build was still pushed as the last step

rayangler (approved):
Makes sense 👍

casthewiz (approved):
This looks good.

PR #739, opened 2023-01-17
DOP-3388: Add PREVIEW_BUILD_ENABLED flag for upcoming writer builds without consistent-nav

mmeigs:
### Ticket

[DOP-3388](https://jira.mongodb.org/browse/DOP-3388)

### Notes

The objective is to have writer builds (on `prd`) build faster. The first thrust of this is to remove the very large `consistent-nav` dependency. This will omit the header and footer for docs pages but keep the dimensions in place for an accurate viewing experience. 

Added support for env var flag `PREVIEW_BUILD_ENABLED` that will enable upcoming change to **Snooty** builds when in `prd` environments - the docs writers' staging environment as seen [here](https://wiki.corp.mongodb.com/display/DE/Environments). 

The six environment variables (one for each environment) are in our **AWS Parameter Store**. Each has the value of `false` except for the `prd` environment, which has the value of `true`.


schmalliso:
@mmeigs OH NO! Writer builds don't build on `dotcomstg` they build on `prd`; `dotcomstg` is DOP's preprd environment which should continue to have the topnav.  (Environment doc is here: https://wiki.corp.mongodb.com/display/DE/Environments).

mmeigs:
> 
@schmalliso 
Ok! Wires crossed in getting this info! So, am I correct that the only reliable flag available is the `jobType` itself as a `githubPush`? 

And oh, I see. So, `dotcomprd` is real production. Whereas `prd` is writers staging. Is `prd` only EVER staging?

schmalliso:
Eventual name is a goal we were shooting for some many months ago, but there are no *actual* plans to achieve said goals :(. `prd` is only used for writer staging so you can confidently use that as the flag we care about here.

seungpark (approved):
this LGTM. verified only writers-staging is set to true in AWS param store 👍 and clean use of env vars @mmeigs !

PR #740, opened 2023-01-17
DOP-3447: delete stale metadata

seungpark:
[DOP-3447](https://jira.mongodb.org/browse/DOP-3447)

**Goal:** remove stale metadata from `snooty_<env>.metadata` collection

**Description:**
- add a db utility function that deletes from collection, using _ids as reference
- add a metadata function that finds the stale metadata for current `project-branch` combination and extract the _ids of said documents
- call above function as a last step of the persistence module app
- add unit tests to metadata module

**note:** it is a non-goal to clear the metadata collections. we can do this as a one-off task to clear any testing data, if need be


modules/persistence/src/services/metadata/index.ts
```diff
@@ -1,23 +1,34 @@
...
+}
 // Service responsible for memoization of metadata entries.
 // Any extraneous logic performed on metadata entries as part of upload should be added here
 // or within subfolders of this module
 const metadataFromZip = (zip: AdmZip) => {
   const zipEntries = zip.getEntries();
-  return zipEntries.filter((entry) => entry.entryName === 'site.bson').map((entry) => deserialize(entry.getData()));
+  return zipEntries
+    .filter((entry) => entry.entryName === 'site.bson')
+    .map((entry) => deserialize(entry.getData()))[0] as Metadata;
```
seungpark, line 25: minor change to return the metadata file itself, not a list

seungpark: see unit test added below

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -1,21 +1,15 @@
...
 
-export interface Metadata {
-  project: string;
-  branch: string;
-  associated_products?: AssociatedProduct[];
-  toctree: ToC;
-  toctreeOrder: any[];
-  [key: string]: any;
-}
+export type SharedMetadata = Metadata;
```
casthewiz, line 12: Good catch on the type duplication here!

casthewiz (approved):
This looks good!

PR #746, opened 2023-01-23
update concurrency and remove deploy dev on push

seungpark:
**Goal:** 
- update concurrency groups so that dev/stg/prd workflows can run in parallel
- remove dev deploy (unsure of purpose...) <- should we remove this completely? it hasn't been required ever, or successful in 4 months...


mmeigs (approved):
LGTM. This will be so soothing.

PR #748, opened 2023-01-25
DOP-3384: Use snooty parser binary builds

i80and:
* `apt install unzip` so we can extract the needful
* Remove the `FLIT_ROOT_INSTALL` environment variable, since we're no longer invoking `flit`
* Move the parser install logic to before the user is switched, so that we can install the parser to `/opt/snooty/`
* Add `/opt/snooty` to the system PATH so that `/opt/snooty/snooty` can be resolved
* Instead of git cloning and flit installing and all that jazz, just curl the salient parser release for linux x86_64, and unzip it in `/opt/`

Building the image succeeds, and confirmed `/opt/snooty/snooty` is accessible in the PATH.

Staged the server manual:
* https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63d41e68cbd7d810f293c565
* https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/stable/

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,23 +23,21 @@ RUN cd ./modules/oas-page-builder \
 FROM ubuntu:20.04
 ARG SNOOTY_PARSER_VERSION=0.13.16
 ARG SNOOTY_FRONTEND_VERSION=0.13.36
-# The Redoc CLI branch will most likely stay static. Updates to the branch should 
+# The Redoc CLI branch will most likely stay static. Updates to the branch should
 # be limited to CLI bug fixes and Redoc dependency version bumps
 ARG REDOC_CLI_VERSION=0.1.1
 ARG FLIT_VERSION=3.0.0
 ARG NPM_BASE_64_AUTH
 ARG NPM_EMAIL
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -23,23 +23,21 @@ RUN cd ./modules/oas-page-builder \
 FROM ubuntu:20.04
 ARG SNOOTY_PARSER_VERSION=0.13.16
 ARG SNOOTY_FRONTEND_VERSION=0.13.36
-# The Redoc CLI branch will most likely stay static. Updates to the branch should 
+# The Redoc CLI branch will most likely stay static. Updates to the branch should
 # be limited to CLI bug fixes and Redoc dependency version bumps
 ARG REDOC_CLI_VERSION=0.1.1
 ARG FLIT_VERSION=3.0.0
```
rayangler, line 29: Is this arg safe to remove as well since we're no longer using flit?

rayangler (approved):
LGTM! Just one non-blocking question about removing another flit arg

Dockerfile
i80and: OOOH yes, good call. Thanks for the catch!

casthewiz (approved):

PR #755, opened 2023-01-31
DOP-3480: unnest merged ToC children, and add node.options.urls

seungpark:
**Goal:** 
1. when merging ToC nodes, we were previously including the root node of the associated product, which lead to a double nested ToC. **this PR unnests the children of the root node so that the associated product's root node's children are directly injected into the umbrella project's target node** 
2. given an associated product's root ToC can lead to two links (ie. https://www.mongodb.com/docs/atlas/cli/v1.3/ vs https://www.mongodb.com/docs/atlas/cli/stable), we need to be able to change the href of the ToC on the front end. **this PR adds an option to a ToC node. if it has node.options.versions: str[], it also has node.options.urls: {branchName: urlString}**

**Note:** 
Due to lack of clean up, ToC nodes front end ToC node's priorities will have to be updated so that a node's href is determined in order of:
- node.options.urls[currentVersion]
- node.options.url
- node.options.slug

### DIFF
--- a/modules/persistence/src/services/metadata/ToC/index.ts
+++ b/modules/persistence/src/services/metadata/ToC/index.ts
@@ -47,15 +47,24 @@ const mergeNode = (node: ToC, tocs: ToCInsertions, currentProject) => {
   if (!associatedProject) return node;
   const branches = Object.keys(associatedProject);
   node.options.versions = branches;
-  node.children = branches.map((branch) => {
-    const child = needsUrlifiedToC ? associatedProject[branch].urlified : associatedProject[branch].original;
-    const options = {
-      ...child.options,
-      version: branch,
-    };
-    child.options = options;
-    return child;
-  });
+  node.options.urls = node.options.urls || {};
+  for (const branch of branches) {
+    node.options.urls[branch] = associatedProject[branch]['urlified']?.url;
+  }
+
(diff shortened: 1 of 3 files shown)


modules/persistence/src/services/metadata/ToC/index.ts
```diff
@@ -48,15 +48,35 @@ const mergeNode = (node: ToC, tocs: ToCInsertions, currentProject) => {
...
-    };
-    child.options = options;
-    return child;
-  });
+  node.options.urls = node.options.urls || {};
+  for (const branch of branches) {
+    node.options.urls[branch] = associatedProject[branch]['urlified']?.url;
+  }
+
+  if (node.options?.project === currentProject) {
```
seungpark, line 56: @casthewiz this was to flip the injection - merged node for associated project was originally from umbrella, but should be the index slug ("/")

if merged node is for umbrella project (else), we should delete the slug here and resort to the options.urls as first priority on the front end. (with fallbacks to slug / url as already implemented)

seungpark: results from above change:
image

image

casthewiz: Makes a lot of sense. Thanks for explaining and showing the results here!

casthewiz (approved):

PR #760, opened 2023-02-02
Dop 3489

casthewiz:
### The Summary

Changes secret reference to fully eliminate potential occurrences of values leaking through re-runs of deleted github actions. (This enables us to delete the leaked value and secret).

Removes failing dev-deploy as a ridealong (no need to have it run on every PR, we should revisit dev experience for this repo in general).

### The Specifics

Even though I removed the actions which std-out'd NPM_BASE_64_AUTH, they can still be re-run, which will result in any subsequently updated values to that secret being logged.

Thus, I've deleted NPM_BASE_64_AUTH and moved it to NPM_BASE64_AUTH, preserving the variable's semantic name while removing the possibility of older actions referencing the new value.

As a ridealong, I've deleted `dev-deploy`, since it typically failed and it's good to remove vestigial actions which reference our secrets.

I've also deleted our installation of snooty devhub, since our build times are getting onerous and it's safe to remove.


.github/workflows/deploy-stg-ecs.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-3489"
```
```suggestion
```

This needs to be applied before merge.

rayangler (approved):

schmalliso (approved):

seungpark (approved):

PR #762, opened 2023-02-08
DOP-3499: concatenate associated children when merging node

seungpark:
the source of the bug for missing ToC was that we were not returning the merged children from each version as the merging was happening, and hence was just returning the last associated branch's ToC children.

screenshot below shows document with merged ToC nodes from different versions already existing in the DB:
image

(see document with ObjectId 63e3d944a3bbd4e497ecf964 in snooty_dev.metadata)


casthewiz (approved):
Makes sense - I totally overlooked this in the previous PR. Good catch!

PR #764, opened 2023-02-14
DOP-3518: get umbrella metadata with one from repos_branches

seungpark:
**Goal:** reduce chance of bug in production where associated products query for the wrong umbrella product. was able to reproduce this error locally when attempting to persist an associated product on a testing branch

Persistence module (on master branch) would query `snooty_<env>.metadata` for an entry that has `associated_products` with one of the associated products. This could potentially lead to a test branch, and persistence module will exit quickly due to error `No repoBranches entry available for umbrella metadata with project: cloud-docs, branch: test-branch`.

this change adds a double trip query to first check for umbrella metadata, then its verified branches, then its metadata within the verified branches. 

**note:** should the umbrella metadata only be from the default (first of repo_branches.branches) or can it be one of any umbrella product versions. ie. atlas has v1 and v1.2, both have associated cli versions v1.3 and v1.4, which should be the target umbrella metadata


modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -74,10 +82,25 @@ export const hasAssociations = (metadata) => !!metadata.associated_products?.len
 const umbrellaMetadataEntry = async (project: string): Promise<Metadata> => {
   try {
     const snooty = await db();
+
+    // first find any umbrella
+    const umbrella = await snooty.collection('metadata').findOne({
+      'associated_products.name': project,
+      is_merged_toc: false,
+    });
```
casthewiz, line 90: nit for the future: this should probably return most recent result, but not worth worrying about yet.

modules/persistence/src/services/metadata/associated_products/index.ts
```diff
@@ -74,10 +82,25 @@ export const hasAssociations = (metadata) => !!metadata.associated_products?.len
...
+
+    if (!umbrella) {
+      return null as unknown as Metadata;
+    }
+
+    const repoDoc = await getRepoBranchesEntry(project);
```
casthewiz, line 96: Did we want to use the `branch` from umbrella as an argument here?

casthewiz (commented):
This LGTM structurally, just want to check on some small things

> note: should the umbrella metadata only be from the default (first of repo_branches.branches) or can it be one of any umbrella product versions. ie. atlas has v1 and v1.2, both have associated cli versions v1.3 and v1.4, which should be the target umbrella metadata

It should be from any umbrella product version. 👍

casthewiz:
I spent a good amount of time trying to think about whether we should batch all the repoBranches document fetching at once - still haven't really thought about a good way to do it, but worth considering.

modules/persistence/src/services/metadata/associated_products/index.ts
seungpark: i think this would re-introduce the error here. i purposely left out the branch here

ie. `writers staging branch (which includes associated products) is persisted the latest` ->  no repo_branches entry for this branch

modules/persistence/src/services/metadata/associated_products/index.ts
seungpark: that's a good catch. adding sort in the findOptions

seungpark:
> I spent a good amount of time trying to think about whether we should batch all the repoBranches document fetching at once - still haven't really thought about a good way to do it, but worth considering.

i agree. this can be cached in memory with `project` options to minimize storage. we can take this as a performance improvement along with the early returns and queue shifts

casthewiz (approved):
LGTM!

PR #766, opened 2023-02-15
DOP-2565: Create cron job to fail stuck jobs

rayangler:
### Ticket

[DOP-2565](https://jira.mongodb.org/browse/DOP-2565)

### Notes

* Adds a new cron job `FailStuckJobs` to update `inProgress` jobs that started 8+ hours ago to `timedOut` in the database.
* Adds a new `taskId` field to the job's document. This is used for programmatically stopping failed ECS tasks with said `taskId`.
* Tested by setting cron schedule and database query to 2 minutes instead of 8 hours and then test deploying jobs on preprd. If needed, let me know if you'd like to test, so we can take over preprd.
* Created a related [wiki page](https://wiki.corp.mongodb.com/display/DE/Stopping+Stuck+Autobuilder+Jobs).
* Example of [failed job](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=63efbaa9519cf19d637d3c3f) running for longer than 2 minutes.


src/services/containerServices.ts
```diff
@@ -87,4 +87,21 @@ export class ECSContainer implements IContainerServices {
...
+        cluster: clusterName,
+        task: taskId,
+      });
+    } catch (error) {
+      // If task not found, then the task should have already been stopped properly
+      if (error.code === 'TaskNotFoundException') {
+        this._logger.error('stopZombieECSTask', error.message);
+      } else {
+        throw error;
+      }
```
rayangler, line 104: Need to circle back and double-check the correct exception. Pretty sure both the exception and the syntax are incorrect

api/controllers/v1/jobs.ts
```diff
@@ -62,13 +62,22 @@ export const HandleJobs = async (event: any = {}): Promise<any> => {
...
             queueUrl = c.get('jobUpdatesQueueUrl');
             await NotifyBuildProgress(jobId);
             break;
           case JobStatus[JobStatus.failed]:
+            await NotifyBuildSummary(jobId);
+            const taskId = body['taskId'];
+            if (taskId) {
+              await stopECSTask(taskId, consoleLogger);
+            }
+            break;
```
rayangler, line 80: I think alternatively, we can use a new job status (like `timedOut`) if we want to prevent conflating the `failed` status. Wasn't sure how precise we wanted to be though

src/services/containerServices.ts
```diff
@@ -87,4 +87,24 @@ export class ECSContainer implements IContainerServices {
...
+        cluster: clusterName,
+        tasks: [taskId],
+      });
+
+      // Only stop ECS task if it's still running
+      if (res.tasks?.[0].lastStatus === 'RUNNING') {
+        await this._client.stopTask({
+          cluster: clusterName,
+          task: taskId,
+        });
```
rayangler, line 104: This was recently updated to ensure that `stopTask` was only called if the described task instance is still running. 

BEFORE MERGING: I should double-check that this logic does what it should and the proper IAM privileges are set

api/controllers/v1/jobs.ts
schmalliso: I have a slight preference for `timedOut` but it's *very* slight.

src/services/containerServices.ts
```diff
@@ -87,4 +87,24 @@ export class ECSContainer implements IContainerServices {
     }
     return '';
   }
+
+  async stopZombieECSTask(taskId: string) {
```
schmalliso, line 91: 10/10 naming here

schmalliso (approved):
This looks good to me! Ideally I'd like @casthewiz to also have a look since autobuilder changes are scary. Thanks for the wiki page, it's delightfully thorough.

api/controllers/v1/jobs.ts
rayangler: Created new `timedOut` status to ensure we're handling separately

src/services/containerServices.ts
```diff
@@ -87,4 +87,24 @@ export class ECSContainer implements IContainerServices {
...
+    const clusterName = this._config.get<string>('taskDefinitionFamily');
+    try {
+      const res = await this._client.describeTasks({
+        cluster: clusterName,
+        tasks: [taskId],
+      });
+
+      // Only stop ECS task if it's still running
+      if (res.tasks?.[0].lastStatus === 'RUNNING') {
+        await this._client.stopTask({
```
casthewiz, line 101: Nit/Non-blocking suggestion:

It might be valuable in the future to attempt to log the `task` object on the response if it exists - just for the sake of being able to diagnose if the task can't be stopped, or diagnosing why the task became hanging. This can definitely be addressed in a future ticket though!

https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-ecs/interfaces/stoptaskcommandoutput.html

casthewiz (approved):
LGTM! One minor suggestion, nonblocking to merge.

src/services/containerServices.ts
rayangler: Good idea! Let's address this in a future ticket so we can test output, formatting, etc. separately.

PR #767, opened 2023-02-16
Ignore modules when formatting

rayangler:
`prettier` would catch unformatted files nested in the `modules/` directory, leading to potential errors in the GH Actions when running `npm run format` on the repo. 

This PR does the following:
* Uses `prettier` to format unformatted metadata json file in the persistence module. (Perhaps unnecessary given the bullet point below.)
* Ignores files nested in the `modules/` directory when formatting. This is consistent with the behavior for the root `.eslintrc.js` file.


seungpark (approved):
good catch Raymund! thanks for disabling the modules subfolder linting

PR #771, opened 2023-02-23
DOP-3511

seungpark:
**Goal:** emit a warning from the persistence module if associated products are of wrong values

**Process:** 
- when parsing the original `metadata` zip file passed to persistence module, mutate the parsed result if associated product does not exist in `repos_branches` collection
- above process will parse an error when `associated_products` do not exist: 
```
No branches found for associated project(s) [atlas-clisss]. Removing such associated_products
```
- **note:** since we are making multiple trips to `repos_branches` from beginning to end -> do the TODO to separate it out into its own module and cache the db results to reduce number of trips to db
- tested with valid data as well on snooty_dev (build id 63f79947c640ca9828ce55d5) : 
image


modules/persistence/src/services/metadata/repos_branches/index.ts
```diff
@@ -0,0 +1,98 @@
...
+  prefix: EnvKeyedObject;
+  [key: string]: any;
+}
+
+const internals: { [key: project]: ReposBranchesDocument } = {};
+
+// Queries pool*.repos_branches for all entries for associated_products in a shared metadata entry
+export const getAllAssociatedRepoBranchesEntries = async (metadata: Metadata) => {
+  const { associated_products } = metadata;
+  if (!associated_products || !associated_products.length) return [];
```
casthewiz, line 33: nit: we can do something like `const { associated_products = [] }` since the destructure is for only one property here

casthewiz (approved):
Broadly looks good, and I like splitting out the `repos_branches` code and adding the cache!

Some nitpicks on my end, but structure seems sound to me.

modules/persistence/src/services/metadata/index.ts
```diff
@@ -18,26 +19,55 @@ export interface Metadata {
...
+      return metadata;
+    }
+    const invalidNames: project[] = [];
+    const promises = metadata['associated_products'].map(async (ap) => {
+      const branchEntry = await getRepoBranchesEntry(ap.name);
+      if (!branchEntry) {
+        invalidNames.push(ap.name);
+      }
+    });
+    await Promise.all(promises);
```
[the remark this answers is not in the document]
seungpark: we actually need this here as the `promises` const is a list of promises that won't be awaited for.

here is an example of this in codepen https://codepen.io/seungpark222/pen/JjaaLbx

(calling asnycFunc awaits all promises before logging in the final line, whereas asyncList logs the final line before promises resolve)

modules/persistence/src/services/metadata/repos_branches/index.ts
seungpark: good catch! i also was not using the lists of `res` and `fetch` as intended, and fixed in the latest commit.

PR #772, opened 2023-02-27
DOP-3437: Remove DevHub code from Autobuilder

rayangler:
### Ticket

[DOP-3437](https://jira.mongodb.org/browse/DOP-3437)

### Notes

* Removes DevHub-related code paths.
* Removes DevHub-related configuration from yml files. 
   * **Open question:** Do we want to remove the corresponding values from Parameter Store after validating that preprd and prd environments are unaffected by config removal? Just want to make sure that those values are not needed for long-term reference.
* Feel free to let me know if there's anything else I could/should delete that is missing from this PR!


api/controllers/v1/dochub.ts
```diff
@@ -1,10 +1,6 @@
 import { FastlyConnector } from '../../../src/services/cdn';
 import { ConsoleLogger } from '../../../src/services/logger';
 import { CDNCreds } from '../../../src/entities/creds';
-import * as c from 'config';
```
schmalliso, line 4: THANK YOUUUUUU

schmalliso (approved):

PR #774, opened 2023-03-01
[DOP-3506]: Autobuilder generates pages using version data

branberry:
## Description:
This PR addresses generating OAS pages for each major and minor version (if present)

[DOP-3506](https://jira.mongodb.org/browse/DOP-3506)


modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -15,21 +15,36 @@ const fetchTextData = async (url: string, errMsg: string) => {
...
   // but it's possible to have more in the future with varying processes.
   const keywords = ['cloud'];
   if (!keywords.includes(apiKeyword)) {
     throw new Error(`${apiKeyword} is not a supported API for building.`);
   }
 
+  let versionExtension = '';
+
+  if (apiVersion) {
+    versionExtension += `-${apiVersion}`;
```
rayangler, line 35: We might need to do something like:

```
versionExtension += `-v${parseInt(apiVersion)}`
```

Major API versions provided by APIx are stringified floats, but the S3 URLs expect `v{int}`

rayangler (commented):
Still reviewing when I can, but noticed this one potential issue when fetching S3 URLs.

modules/oas-page-builder/src/services/pageBuilder.ts
branberry: Ah gotcha, makes sense! I do have a question, does the OAS data such as the spec name, api version and resource version come from Atlas, or will this be invoked directly from APIx when it is called here? If it is the former, I believe there should be logic already to remove the .0 at the end. Either way, I think if I add the logic to make the version number into an int that it won't cause any issues.

rayangler: The module receives the metadata from the parser, implemented in https://github.com/mongodb/snooty-parser/pull/449

branberry: Gotcha! Thank you for the clarification, updated this to strip out the trailing .0

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -15,21 +15,36 @@ const fetchTextData = async (url: string, errMsg: string) => {
...
   // but it's possible to have more in the future with varying processes.
   const keywords = ['cloud'];
   if (!keywords.includes(apiKeyword)) {
     throw new Error(`${apiKeyword} is not a supported API for building.`);
   }
 
+  let versionExtension = '';
+
+  if (apiVersion) {
+    versionExtension += `-${apiVersion.split('.')[0]}`;
```
rayangler, line 35: Thanks for stripping away the decimal! Seems like the `v` is still missing from the version extension though. Should be: `-v${apiVersion.split('.')[0]}`.

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -50,39 +65,76 @@ const getAtlasSpecUrl = async (apiKeyword: string) => {
...
+      spec = localFilePath;
+    } else if (sourceType === 'atlas') {
+      spec = await getAtlasSpecUrl({ apiKeyword: source, apiVersion, resourceVersion });
+      // Ignore "incompatible types" warnings for Atlas Admin API/cloud-docs
+      buildOptions['ignoreIncompatibleTypes'] = true;
+    } else {
+      throw new Error(`Unsupported source type "${sourceType}" for ${pageSlug}`);
+    }
+
+    const finalFilename = normalizePath(`${output}/${pageSlug}/index.html`);
```
rayangler, line 106: For versioned OpenAPI pages, we should have the filename output path include the resource version. For example, if "2023-01-01" is a resource version for v2, the final filename should be something like:

`/reference/api-resources-spec/v2/2023-01-01/index.html`

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -50,39 +65,76 @@ const getAtlasSpecUrl = async (apiKeyword: string) => {
...
+      );
       continue;
     }
+
+    if (resourceVersions) {
+      // if a resource versions array is provided, then we can loop through the resourceVersions array and call the getOASpec
+      // for each minor version
+      for (const resourceVersion of resourceVersions) {
+        await getOASpec({ source, sourceType, output, pageSlug, redocExecutor, repoPath, apiVersion, resourceVersion });
+      }
```
rayangler, line 134: Note that if there are resource versions for an API version, we should also create a page for the general API version. This page will have all of the resource versions bundled together into one, while the individual resource version pages are specific to that version. The tables in [this section](https://docs.google.com/document/d/1F2tmPnFM1Kj_gFZDAyWJUAO0M5L1Bd6xQ4pHTcmaI9M/edit#heading=h.k7miji4vp0u2) have example file outputs!

modules/oas-page-builder/tests/unit/services/pageBuilder.test.ts
```diff
@@ -67,8 +67,7 @@ describe('pageBuilder', () => {
     ];
 
     await buildOpenAPIPages(testEntries, testOptions);
-    // const mockSoundPlayerInstance = SoundPlayer.mock.instances[0];
-    // const mockRedocExecutorInstance = RedocExecutor.mock.instances[0];
```
rayangler, line 71: Whoops, thanks for catching and removing!

modules/oas-page-builder/src/services/pageBuilder.ts
branberry: oops!! will add that

modules/oas-page-builder/src/services/pageBuilder.ts
branberry: Gotcha! Apologies for missing this, will update

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -15,21 +15,36 @@ const fetchTextData = async (url: string, errMsg: string) => {
...
   // Currently, the only expected API fetched programmatically is the Cloud Admin API,
   // but it's possible to have more in the future with varying processes.
   const keywords = ['cloud'];
   if (!keywords.includes(apiKeyword)) {
     throw new Error(`${apiKeyword} is not a supported API for building.`);
   }
 
+  let versionExtension = '';
+
+  if (apiVersion) {
```
YusufAzam, line 34: const versionExtension = apiVersion && resourceVersion ? versionExtension += `-${resourceVersion}` : versionExtension += `-v${apiVersion.split('.')[0]}`

branberry: Hey Yusuf! Thank you for the comment, I made a slight modification to what you provided, but I like the idea 👍

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -50,39 +59,79 @@ const getAtlasSpecUrl = async (apiKeyword: string) => {
...
+      spec = await getAtlasSpecUrl({ apiKeyword: source, apiVersion, resourceVersion });
+      // Ignore "incompatible types" warnings for Atlas Admin API/cloud-docs
+
+      buildOptions['ignoreIncompatibleTypes'] = true;
+    } else {
+      throw new Error(`Unsupported source type "${sourceType}" for ${pageSlug}`);
+    }
+
+    const filePathExtension = `${apiVersion ? `/v${apiVersion.split('.')[0]}` : ''}${
+      resourceVersion ? `/${resourceVersion}` : ''
```
mmeigs, line 101: Should the last conditional here (like in `versionExtension` above) have two truthy checks - for `resourceVersion` and `apiVersion`? This would just ensure that we wouldn't get half of a correct filePathExtension (ie '-2020-10-02' with no apiVersion) in an odd edge case. We'd either want a fully valid filePath or an empty string, I think.

branberry: Good catch! Thank you for that, will update accordingly.

modules/oas-page-builder/tests/unit/services/pageBuilder.test.ts
```diff
@@ -21,7 +21,10 @@ jest.mock('../../../src/services/database', () => ({
 }));
 
 // Helper function for concatendated output path
-const getExpectedOutputPath = (destination: string, pageSlug: string) => `${destination}/${pageSlug}/index.html`;
+const getExpectedOutputPath = (destination: string, pageSlug: string, apiVersion?: string, resourceVersion?: string) =>
+  `${destination}/${pageSlug}${apiVersion ? `${apiVersion ? `/v${apiVersion.split('.')[0]}` : ''}` : ''}${
+    resourceVersion ? `/${resourceVersion}` : ''
```
mmeigs, line 26: This shouldn't be an issue, but for consistency maybe add the extra truthy check here as well!

mmeigs (approved):
Looks good! I put one teeny nitpick. Once you update the branch and lint/format, I think we're good to go!

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -50,39 +59,80 @@ const getAtlasSpecUrl = async (apiKeyword: string) => {
...
+    } else if (sourceType === 'atlas') {
+      spec = await getAtlasSpecUrl({ apiKeyword: source, apiVersion, resourceVersion });
+      // Ignore "incompatible types" warnings for Atlas Admin API/cloud-docs
+
+      buildOptions['ignoreIncompatibleTypes'] = true;
+    } else {
+      throw new Error(`Unsupported source type "${sourceType}" for ${pageSlug}`);
+    }
+
+    const filePathExtension = `${apiVersion ? `/v${apiVersion.split('.')[0]}` : ''}${
```
rayangler, line 100: When attempting to build separate v1 and v2 pages locally, the resulting page includes the API version twice. Example:

```
/reference/api-resources-spec/v2/v2/index.html
```

I'm assuming it's because we're concatenating the API version to the file path here, but I'm wondering if anyone else is running into this or if this is just user error on my part 😅.

mmeigs: I'm not having that problem when running solely the oas-page-builder module on cloud-docs. Let's sync about where you're seeing this.

rayangler (approved):

mmeigs (approved):

PR #776, opened 2023-03-02
[DOP-3507]: Autobuilder passes version data to CLI

branberry:
## Description:
This PR provides the redoc cli with the apiVersion and resourceVersion arguments.

[DOP-3507](https://jira.mongodb.org/browse/DOP-3507)


mmeigs (commented):
The data contract with the Redoc CLI actually is a bit more complicated! You can see a great example written up in the [Redoc CLI section](https://docs.google.com/document/d/1F2tmPnFM1Kj_gFZDAyWJUAO0M5L1Bd6xQ4pHTcmaI9M/edit#heading=h.7aqgyy2zxh6i) of the Tech Spec. Make sure to read through the Autobuilder section and the Redoc CLI section to really understand how these two pieces fit together. I'll look at it more tomorrow too!

branberry:
@mmeigs Gotcha! My apologies, I will convert this to a draft, and read up on these sections more closely before moving forward. Thanks!

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -148,7 +152,40 @@ async function getOASpec({
 
     const path = `${output}/${pageSlug}${filePathExtension}/index.html`;
     const finalFilename = normalizePath(path);
-    await redocExecutor.execute(spec, finalFilename, buildOptions);
+
+    let versionOptions: RedocVersionOptions | undefined;
+
+    if (resourceVersions && apiVersion) {
+      const rootUrl = `${siteUrl}/${pageSlug}`;
```
mmeigs, line 159: I think keeping this forward slash is ok just in case, but we should run this through the util `normalizeUrl` because right now it has two forward slashes in a row

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -148,7 +152,40 @@ async function getOASpec({
 
     const path = `${output}/${pageSlug}${filePathExtension}/index.html`;
     const finalFilename = normalizePath(path);
-    await redocExecutor.execute(spec, finalFilename, buildOptions);
+
+    let versionOptions: RedocVersionOptions | undefined;
+
+    if (resourceVersions && apiVersion) {
```
mmeigs, line 158: There's an edge case I'm seeing where resourceVersions might "exist" but might be an empty array. I think we might want to account for that here!

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -148,7 +152,40 @@ async function getOASpec({
...
+      const rootUrl = `${siteUrl}/${pageSlug}`;
+
+      // if there is no resource version provided, but there is a resourceVersions array present,
+      // get the latest resource version from the array, and assign it to the active resource version
+      if (!resourceVersion) {
+        const latestResourceVersion = resourceVersions.sort((a, b) => new Date(a).getTime() - new Date(b).getTime())[
```
mmeigs, line 164: I did NOT realize how much faster this comparison is than just subtracting two Dates!! That is a wild thing I learned today.

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -186,21 +223,39 @@ export const buildOpenAPIPages = async (
...
           resourceVersion,
+          siteUrl,
+          resourceVersions,
         });
+
         if (!isSuccessfulBuild) totalSuccess = false;
       }
+
+      // provide no resource version in this context for the base version
+      await getOASpec({
```
mmeigs, line 234: I might be making a mistake at the end of the day, but I believe we don't need to have this called twice here. I think the way it was before handled this! Because no matter what, we will want to call getOASpec once for the "base" API Version spec.

mmeigs (commented):
Left a few small comments!! And I think the rest of the testing will occur when we finally have the entire pipeline created in a testing environment to show us any other edge cases!

modules/oas-page-builder/src/services/pageBuilder.ts
branberry: I added this second call to handle the "base" case, where there exists a resource version array, but no resource version. In this case, we will pull the latest resource version. 

In the else statement, we do not provide the `resourceVersions` array. In this case, we do not provide the version options to the redoc cli.

With that said, I think I can refactor this code in such away where I don't have 3 API calls. I'll push that change up shortly!!

mmeigs:
Thanks @branberry, this is fantastic!! Working my way through releases to get us to a point where we can spin up a full environment.

mmeigs (approved):

PR #795, opened 2023-04-12
[DOP-3586]: Cancel archive job for staging

branberry:
Ticket
[DOP-3586](https://jira.mongodb.org/browse/DOP-3586)

### DIFF
--- a/api/controllers/v1/jobs.ts
+++ b/api/controllers/v1/jobs.ts
@@ -296,11 +296,13 @@ function getMongoClient(config: IConfig): mongodb.MongoClient {
   return new mongodb.MongoClient(url);
 }
 
+const STAGING_ENVS = ['stg', 'prd'];
+
 async function SubmitArchiveJob(jobId: string) {
   const consoleLogger = new ConsoleLogger();
   const environment: string = c.get('env');
 
-  if (environment === 'staging') {
+  if (STAGING_ENVS.includes(environment)) {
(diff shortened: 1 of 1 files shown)


api/controllers/v1/jobs.ts
```diff
@@ -299,6 +299,12 @@ function getMongoClient(config: IConfig): mongodb.MongoClient {
 async function SubmitArchiveJob(jobId: string) {
   const consoleLogger = new ConsoleLogger();
   const environment: string = c.get('env');
+
+  if (environment === 'staging') {
```
rayangler, line 303: I believe in this case, `env` should correspond to one of our several platform environments (`dotcomprd`, `prd`, `dotcomstg`, `stg`, etc.). We might want to check for either `stg` or `prd`, since those are directly associated with content staging jobs at the moment.

Alternatively, we could check to see if the job payload is of type `githubPush`, if we want to make this agnostic of environment

rayangler: In theory, no. But in current practice, yes!

Each env has an associated [webhook](https://us-east-2.console.aws.amazon.com/apigateway/main/apis?region=us-east-2) that can be used to trigger a build on GitHub push. We share only the `prd` webhook with the docs team so content staging jobs are limited to that env.

However, nothing is stopping anyone from performing a `githubPush` job to `dotcomstg` if they have their repo configured with the `dotcomstg` webhook by accident or intentionally

rayangler (approved):

PR #796, opened 2023-04-13
[DOP-3637]: Create AWS CDK Project

branberry:
## Ticket 
[DOP-3637](https://jira.mongodb.org/browse/DOP-3637)

## Notes

This project was created using the `cdk` cli tool. Right now, the stack does not create anything if deployed. The project was also bootstrapped. This means I had to execute the command `cdk bootstrap`, which creates the resources the CDK project needs to use to manage the CloudFormation stack, and provision resources. For example, lambda code , static assets, or ECR images are stored within resources that are specific for the CDK project. The boostrap process also creates the necessary IAM roles to provision resources that we define in our stacks. 

The stack can be found [here](https://us-east-2.console.aws.amazon.com/cloudformation/home?region=us-east-2#/stacks/stackinfo?filteringText=&filteringStatus=active&viewNested=true&stackId=arn%3Aaws%3Acloudformation%3Aus-east-2%3A216656347858%3Astack%2FCDKToolkit%2F694936b0-da2e-11ed-a356-0a3ded2b2be4).

More information on the boostrapping process can be found [here](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html).

This is a necessary, one time step for the project. It will not need to be run again moving forward.


caesarbell:
Overall pretty solid. I have some small questions and Github beta seems to have provided some warnings you might want to checkout around type `any` and a variable being defined but not used.

rayangler (approved):
Looks good! Thanks for the docs page on bootstrapping!

PR #800, opened 2023-04-18
[DOP-3638]: Transpile API handlers from TS to JS

branberry:
## Ticket
[DOP-3638](https://jira.mongodb.org/secure/RapidBoard.jspa?rapidView=953&projectKey=DOP&view=detail&selectedIssue=DOP-3638)

## Notes

For the AWK CDK project, we now need to manually bundle the TypeScript files that exist within the `api` directory to JavaScript. I have taken the additional step to zip these files, so that we reduce the bundle size when we deploy the lambda functions. 

Given that the API files have TypeScript errors, I decided to resolve these in this PR. This is because we will run into build failures in our CI/CD pipeline when we perform the build command.

## Dependency changes
I added a few additional dependencies for typing purposes. Here is a list of the new dependencies:

```json
    "aws-lambda": "^1.0.7",
    "@octokit/webhooks-types": "^6.11.0",
    "@types/aws-lambda": "^8.10.114"
```


api/controllers/v1/jobs.ts
```diff
@@ -181,9 +222,6 @@ async function NotifyBuildSummary(jobId: string): Promise<any> {
     ),
     entitlement['slack_user_id']
   );
-  return {
```
branberry, line 184: removed this return value as it is never used anywhere

api/controllers/v1/jobs.ts
```diff
@@ -284,9 +331,6 @@ async function NotifyBuildProgress(jobId: string): Promise<any> {
...
```
branberry, line 287: removed this return value as it is never used anywhere

src/entities/job.ts
```diff
@@ -25,48 +25,48 @@ export type Payload = {
...
-  logs: string[] | null | undefined;
-  priority: number | null | undefined;
-  result: any | null | undefined;
-  startTime: Date;
+  endTime?: Date | null;
+  error?: any;
+  logs?: string[] | null;
+  priority?: number;
+  result?: any;
+  startTime: Date | null;
```
branberry, line 58: I made this null because in the `prepGithubPushPayload` function, we assign null to this value. This then gets upserted into the collection, so it appears that we have been providing a null value for this one previously.

api/controllers/v1/jobs.ts
```diff
@@ -134,11 +140,36 @@ async function stopECSTask(taskId: string, consoleLogger: ConsoleLogger) {
...
+    if (typeof val === 'string') {
+      return !isNaN(parseInt(val));
+    }
+    return false;
+  }
+
+  return true;
+}
+
+async function retry(message: JobQueueMessage, consoleLogger: ConsoleLogger, url: string): Promise<void> {
```
branberry, line 161: This isn't used, but I am not sure if there is value in keeping it around still. Any thoughts on keeping this function around?

api/controllers/v1/github.ts
```diff
@@ -27,7 +40,7 @@ async function prepGithubPushPayload(githubEvent: any, branchRepository: BranchR
     title: githubEvent.repository.full_name,
     user: githubEvent.pusher.name,
     email: githubEvent.pusher.email,
-    status: 'inQueue',
+    status: JobStatus.inQueue,
```
branberry, line 43: The `JobStatus.inQueue` enum has `'inQueue'` as the string value.

api/controllers/v1/jobs.ts
schmalliso: I think there is little value in keeping it - we used to try all failed builds three times (I don't recall why) but it was very annoying and provided no value. I'm pro removing unused code.

src/entities/job.ts
```diff
@@ -25,48 +25,48 @@ export type Payload = {
   branchName: string;
   isFork: boolean;
   private: boolean;
-  isXlarge: boolean | null | undefined;
+  isXlarge?: boolean | null;
```
schmalliso, line 28: It's probably out of scope for this, but we should really remove this at some point since we don't have different sizes of builders anymore and *everything* is Xlarge.

schmalliso (approved):
This looks great and also I want another set of eyes that aren't mine before you merge 😁

src/entities/job.ts
branberry: I'll remove, it's actually a small change. I did a quick search, and we only ever set this value to true when we assign it a value at all.

api/controllers/v1/jobs.ts
```diff
@@ -11,13 +11,25 @@ import { Job, JobStatus } from '../../../src/entities/job';
 import { ECSContainer } from '../../../src/services/containerServices';
 import { SQSConnector } from '../../../src/services/queue';
 import { Batch } from '../../../src/services/batch';
+import { APIGatewayEvent, APIGatewayProxyResult, SQSEvent, SQSRecord } from 'aws-lambda';
 
-export const TriggerLocalBuild = async (event: any = {}, context: any = {}): Promise<any> => {
+export const TriggerLocalBuild = async (event: APIGatewayEvent): Promise<APIGatewayProxyResult> => {
   const client = new mongodb.MongoClient(c.get('dbUrl'));
```
mmeigs, line 17: Is this connection ever used? (I know this doesn't have to do with your PR, but great time to clean this if so)

mmeigs (approved):
All this typing work is just so appreciated. LGTM. One possible additional deletion below, but all is good.

api/controllers/v1/jobs.ts
branberry: Good catch! It is not used. I will remove.

PR #804, opened 2023-04-21
DOP-3592: Add legacy dockerfile

i80and:
```
$ hadolint-Linux-x86_64 ./Dockerfile.legacy 
./Dockerfile.legacy:9 DL3008 warning: Pin versions in apt get install. Instead of `apt-get install <package>` use `apt-get install <package>=<version>`
./Dockerfile.legacy:38 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
./Dockerfile.legacy:39 DL3059 info: Multiple consecutive `RUN` instructions. Consider consolidation.
```

I think these warnings are okay


casthewiz (approved):
Those warnings are indeed okay!

PR #807, opened 2023-04-26
DOP-3634: make sure tests cases are sorted

seungpark:
I was running into instances where the snapshots were coming in mixed order. The test change below in associated_products.test.ts ensures same order each time, and thus the snapshot change.

Also, was running into errors for `DuplicateKey` indicating test data was being inserted multiple times. To prevent this, ordered the `clear collection` logic before `populate collection` logic in each test suite.


seungpark:
closing this PR as a different solution is provided [here](https://github.com/mongodb/docs-worker-pool/pull/810)

PR #808, opened 2023-05-01
DOP-3593: Add legacy queue infrastructure

i80and:
This just creates the environments: it does not wire anything up.

If this looks good, I can push it into testing!

### DIFF
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-3593"
 concurrency:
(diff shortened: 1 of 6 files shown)


casthewiz (approved):
This looks good - I think that we might need to move the second docker file build to another action (so it can run in parallel with the deploy step) in the future. But, we can wait and see how the build + deploy timing is impacted first.

casthewiz:
@i80and - to formally respond here, I think this should be good to merge and test within `master` branch 😄

.github/workflows/deploy-stg-ecs.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-3593"
```
```suggestion
      - "DOP-3593"
```

PR #810, opened 2023-05-02
DOP-3634: add test util to make new db for each test suite

seungpark:
PR creates a separate database for each suite, so it mitigates errors of duplicate keys as seen before.

This was caused by jest *randomly* choosing to either run the test suites sequentially or in a bunch, observed by myself and @rayangler unless we specified the option [--runInBand](https://jestjs.io/docs/29.3/cli#--runinband) which would be sequential.

Each caller should be generated a new db name, and thus no collision on Duplicate Error key


rayangler (approved):
LGTM!

PR #812, opened 2023-05-03
[HOTFIX]: Use Node 18 for staging builds

branberry:


seungpark (commented):
do we not need to change the Dockerfile?

PR #814, opened 2023-05-03
DOP-3633: Use bulkWrites for asset upsertion

rayangler:
### Ticket

DOP-3633

### Notes

* Replaces `upsert` function to be `bulkUpsert`. Core logic upserts entire document object by ID, similar to previous `upsert` logic. Biggest difference is that `bulkWrite` is utilized instead of individual `updateOne` operations.
* Observations from local runs that might be interesting to see in production: 
   * Connection count per build seems to be reduced with this change (observed from [cluster monitoring](https://cloud.mongodb.com/v2/5bad1d3d96e82129f16c5df3#/host/replicaSet/5d40b1bccf09a2026cbad969) based on timing with local builds). From console logs, I noticed that there'd be multiple attempts for the db client to connect with the previous iteration (probably through race conditions from promises).
   * Bulk write implementation seems to save about 30 seconds for the server manual locally, but could be dependent on machine.

### DIFF
--- a/modules/persistence/src/services/assets/index.ts
+++ b/modules/persistence/src/services/assets/index.ts
@@ -1,5 +1,5 @@
 import AdmZip from 'adm-zip';
-import { upsert } from '../connector';
+import { bulkUpsert } from '../connector';
 
 const COLLECTION_NAME = 'assets';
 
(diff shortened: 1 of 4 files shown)


seungpark (approved):
this LGTM 👍 sweet update on the tests too

PR #819, opened 2023-05-17
[DOP-3714]: Modify App to Listen to SQS Queue

branberry:
### Story

[DOP-3714](https://jira.mongodb.org/browse/DOP-3714)

### Notes

This story creates a separate app and job manager to handle messages from the queue directly. The code is derived from the `app.ts` directory, but with some small modifications. Primarily, we want the app to be running as long as it has not received a message. This is like the original app, but with a small modification where the job does not continue to listen to additional messages after it has processed one. When the job is done processing, the application ends, also ending the task.


cdk-infra/.gitignore
```diff
@@ -8,4 +8,6 @@ node_modules
 cdk.out
 
 # Build artifacts
-dist
\ No newline at end of file
+dist
+
+cdk.context.json
```
branberry, line 13: Not relevant to the current PR, but I'm adding it now since I want to make sure I don't accidentally commit this file.

src/enhancedApp.ts
```diff
@@ -0,0 +1,87 @@
...
+    fileSystemServices,
+    hybridJobLogger,
+    repoBranchesRepo,
+    cleanUp
+  );
+
+  try {
+    jobManager.start();
+  } catch (err) {
+    consoleLogger.error('App', `ERROR: ${err}`);
```
rayangler, line 83: Just wondering if we'd ever want to make this `enhancedApp` to differentiate between any caught errors here for the regular app.

src/job/enhancedJobManager.ts
```diff
@@ -0,0 +1,53 @@
...
+      fileSystemServices,
+      logger,
+      repoBranchesRepo
+    );
+    this._cleanUp = cleanUp;
+  }
+
+  override async work() {
+    while (true) {
+      const job = await this.getQueuedJob();
```
rayangler, line 44: [getQueuedJob](https://github.com/mongodb/docs-worker-pool/blob/303798fec002e9286bf1bb7d6b482dbf15a78c81/src/job/jobManager.ts#L131) seems to query MongoDB and uses it as its "queue" instead of polling for messages from the SQS. Is this intentional?

branberry: Ah, good catch. I made an assumption from the `app.ts` code that this is reading from the SQS queue, but you're right, it's not. We can discuss in standup or later about how we want to go about this. It might be a smaller change to follow this pattern as opposed to having the task listening to the queue.

src/enhanced/types/job-types.ts
```diff
@@ -0,0 +1,26 @@
+export type JobType = 'githubPush' | 'manifestGeneration' | 'productionDeploy' | 'regression';
```
branberry, line 1: NOTE: These properties will probably change in the near future, but I am keeping this for now as this is what we can potentially expect from the queue. This was added before I realized that we need to make changes with how we interact with the MongoDB collection.

src/enhanced/utils/queue/listen-to-job-queue.ts
```diff
@@ -0,0 +1,48 @@
+import { SQS } from '@aws-sdk/client-sqs';
```
branberry, line 1: NOTE: Main change here

src/enhanced/utils/queue/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const client = new SQS({ region });
+
+  // We want to loop indefinitely so that we continue to poll the queue.
+  while (true) {
+    const receiveMessage = { QueueUrl: queueUrl, MaxNumberOfMessages: 1, WaitTimeSeconds: 2 };
+
+    const res = await client.receiveMessage(receiveMessage);
+
+    if (res.Messages && res.Messages.length > 0) {
```
seungpark, line 22: minor nitpick. would continue early if no messages here

src/enhanced/types/utils/type-guards.ts
```diff
@@ -0,0 +1,53 @@
...
+    typeof isFork === 'boolean' &&
+    typeof repoOwner === 'string' &&
+    (typeof stable === 'undefined' || typeof stable === 'string') &&
+    (typeof primaryAlias === 'undefined' || typeof primaryAlias === 'boolean') &&
+    (typeof project === 'undefined' || typeof project === 'string') &&
+    (typeof prefix === 'undefined' || typeof prefix === 'string') &&
+    (typeof urlSlug === 'undefined' || typeof urlSlug === 'string') &&
+    (typeof aliased === 'undefined' || typeof aliased === 'boolean') &&
+    (typeof newHead === 'undefined' || typeof newHead === 'string') &&
+    (typeof unsafePayload.private === 'undefined' || typeof unsafePayload.private === 'string') // can't object destructure private since it is a reserved word :S
```
rayangler, line 51: It can be destructured if it's renamed. Figured this was worth sharing in case it was something you wanted to do!

```
const { private: isPrivate } = unsafePayload;
```

src/enhanced/utils/queue/index.ts
```diff
@@ -0,0 +1,48 @@
...
+      if (!message.Body) {
+        console.error(`ERROR! Received message from queue without body. Message ID is: ${message.MessageId}`);
+        continue;
+      }
+
+      const payload = JSON.parse(message.Body);
+
+      // Use type guard here to validate payload we have received from the queue.
+      // This ensures that the `payload` object will be of type `JobQueuePayload` after the if statement.
+      if (!isJobQueuePayload(payload)) {
```
rayangler, line 34: Currently, we send messages through SQS with a minimal body ([class](https://github.com/mongodb/docs-worker-pool/blob/4feb61fef683a48c736422ff2977d5777688d93a/src/entities/queueMessage.ts#LL3C1-L3C1)), leveraging the job's ID for db queries as a source of truth.

Are we expecting to change the message's body in a future PR?

src/enhanced/enhancedApp.ts
```diff
@@ -0,0 +1,7 @@
+import { listenToJobQueue } from './utils/queue';
+
+async function app() {
+  const job = await listenToJobQueue();
```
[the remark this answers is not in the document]
branberry: Yep, that's correct!

src/enhanced/utils/queue/index.ts
```diff
@@ -0,0 +1,48 @@
...
+          `ERROR! Invalid payload data received from message ID: ${
+            message.MessageId
+          }. Payload received: ${JSON.stringify(payload)}`
+        );
+        continue;
+      }
+
+      // Great! we received a proper message from the queue. Return this object as we will no longer
+      // want to poll for more messages.
+      return payload;
```
seungpark, line 45: QQ: i'm not sure if this would be the right place, but would we want to delete message after it has been received, or does this happen at the end of the task? did not see anywhere where we are deleting queue messages, perhaps the message is processed (and marked for deletion?)

according to SQS docs on `visibility`, `If the consumer fails to process and delete the message before the visibility timeout expires, the message becomes visible to other consumers.`

src/enhanced/utils/queue/index.ts
branberry: Yes, we will change the payload moving forward. If my understanding is correct, we use the jobId to query the the queue collection in Atlas. Since we are no longer using that collection I am thinking we provide all of the data needed to complete the job as a part of the message.

seungpark (approved):
payload validation and `live listen` "warm job" LGTM. minor nitpick on early returns, and a question about message digestion below

src/enhanced/utils/queue/index.ts
branberry: Ah, good catch. I assumed the message would be removed from the queue if it has been received, but that isn't correct. I'll update my code to handle deletion of a message tomorrow👍

src/enhanced/utils/queue/index.ts
rayangler: Gotcha. Passing in the data the Autobuilder would need to complete the job makes sense.

> Since we are no longer using that collection

Do you mean we're no longer using that collection for fetching jobs, or in general? I imagine we'd still be using the collection to save relevant information for auditing and debugging purposes

i80and (approved):
No objections from me!

src/enhanced/utils/queue/index.ts
branberry: I think it's how SQS and Lambda interact. When a message is read from an SQS queue, and the lambda is triggered, I believe that handles the message deletion on its own. We are using the client here which requires us to do a bit more work I believe to achieve the same behavior.

src/enhanced/utils/queue/index.ts
branberry: Cassidy and I discussed, and we believe that the logging to CloudWatch should handle cover us in this regard. If you do think there is more value in keeping it around though, I'd be more than happy to discuss.

src/enhanced/utils/queue/index.ts
```diff
@@ -0,0 +1,51 @@
...
+    const receiveMessage = { QueueUrl: queueUrl, MaxNumberOfMessages: 1, WaitTimeSeconds: 2 };
+
+    const res = await client.receiveMessage(receiveMessage);
+
+    if (!res.Messages || res.Messages.length === 0) continue;
+
+    const message = res.Messages[0];
+
+    // We have the message body, now we can delete it from the queue.
+    client.deleteMessage({ QueueUrl: queueUrl, ReceiptHandle: message.ReceiptHandle });
```
seungpark, line 27: this returns a promise. would we want to await for this. and as for any awaits, should wrap in try/catch block to catch any client failures. besides this LGTM! thanks for adding the message delete!

branberry: Ah, thank you for pointing this out. Rookie mistake! lol

src/enhanced/utils/queue/index.ts
rayangler: Sure, that makes sense. I was mostly asking for my own understanding of all the changes coming through. I would argue that having a collection of jobs is still beneficial for having easy-to-see/query objects for reference, but if CloudWatch can achieve something similar, then no complaints.

I think the collection's only other uses are for sharing build logs to users, and for tracking metrics via our charts, but I think both of those can be achievable with CloudWatch/AWS magic.

Either way, this may not be relevant to the current PR so I'm down to pivot! Thanks for clearing things up for me and answering my questions!

branberry: Gotcha! That all makes sense to me. CloudWatch should support these metrics, but you're right it might be more work, and I'm definitely flexible with how we go about this. Right now, we are trying to implement this without having to use the collection because it is a bit more difficult to test, and we are also looking to strip down the build process to the bare bones, and that requires not using that collection for now.

Anyhoo, we can definitely revisit this as this enhanced worker gets fleshed out more.

rayangler (approved):
LGTM!

PR #820, opened 2023-05-17
DOP-3718: Gate page upserts behind feature flag

rayangler:
### Ticket

DOP-3718

### (Long) Notes

This PR reintroduces the reverted commit for upserting page documents. Additional code has been added for observability and to make rollbacks easier in case something goes awry in production.

The main suspect for slow build times is the database write speed. When testing locally, page diffing was a reasonable speed (milliseconds—seconds at most—rather than minutes).

Sample [build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=646687e0df0f8173e9d0e344) with index and unordered writes. (Scroll down to see time outputs, or just search for something like "page document update writes".) The job was performed with a clean branch that would expect all pages to be uploaded.

**Notably, there are a few new additions:**
* A feature flag, ~~which should be added to the serverless configuration and Parameter Store prior to merging this PR~~ (DONE. Feel free to double-check values in Parameter Store). The flag is responsible for determining if the persistence module should also upsert page documents after identifying diffs. We'll only keep it enabled in pre-production for now.
* Timers for some of the steps of the persistence module to help with debugging, and for general auditing. We can remove these later along with the feature flag if unnecessary.
* Bulk writes and insertions are now `unordered`. According to [the docs](https://www.mongodb.com/docs/manual/reference/method/Bulk/#unordered-operations), this can sometimes help with performance as operations are reordered for optimization or even handled in parallel.

**Outside of this PR:**
* A unique index was added for `page_id` in the `updated_documents` collection. This should hopefully be the biggest carry in performance.

**Next steps:**
* Enable the feature flag in both pre-production AND production. Enabling in production will allow us to observe build times with more practical use cases and real-world data.
* If there are once again issues with build times, we'll disable the feature through Parameter Store and rerun the Autobuilder deployment to prd. This will allow the code to exist in preprd without needing to revert the code entirely again.
* If things look good in production, a separate PR (perhaps closer to when Gatsby Cloud integration is ready to go) will be created to remove feature flags from the code.


modules/persistence/src/services/connector/index.ts
```diff
@@ -39,11 +39,16 @@ export const db = async () => {
...
   const insertSession = await db();
   try {
-    return await insertSession
-      .collection(collection)
-      .insertMany(docs.map((d) => ({ ...d, build_id: buildId, created_at: buildId.getTimestamp() })));
+    const res = await insertSession.collection(collection).insertMany(
+      docs.map((d) => ({ ...d, build_id: buildId, created_at: buildId.getTimestamp() })),
+      { ordered: false }
+    );
+    console.timeEnd(timerLabel);
```
seungpark, line 50: this will only end on success. should we add these to `finally` block

modules/persistence/src/services/pages/index.ts
```diff
@@ -184,24 +184,42 @@ const updatePages = async (pages: Document[], collection: string) => {
...
+  const diffsTimerLabel = 'finding page differences';
+  console.time(diffsTimerLabel);
   const updatedPagesManager = new UpdatedPagesManager(prevPageDocsMapping, prevPageIds, pages);
   const operations = updatedPagesManager.getOperations();
+  console.timeEnd(diffsTimerLabel);
 
   if (operations.length > 0) {
+    const bulkWriteTimerLabel = 'page document update writes';
+    console.time(bulkWriteTimerLabel);
     await bulkWrite(operations, collection);
```
seungpark, line 205: similar here. can we wrap these in a try/catch/finally block to catch the timer? seems like errors from bulkWrite will also be uncaught

seungpark (changes requested):
the env setup LGTM 👍 minor request on the timers (finishing in `finally` blocks)

modules/persistence/src/services/connector/index.ts
rayangler: Added this to `finally` block. Also added other similar timers to end in `finally` blocks in case db errors are thrown

modules/persistence/src/services/pages/index.ts
rayangler: Wrapped in a try/finally block. Ended up wrapping the main body of the `updatePages` function in a try/catch/finally to catch errors at a function-wide level

seungpark (approved):
LGTM. thanks for the verbose logging !

casthewiz (approved):

PR #821, opened 2023-05-23
[DOP-3640]: Define the Autobuilder Core Construct

branberry:
## Ticket
[DOP-3640](https://jira.mongodb.org/browse/DOP-3640)

## Description

This PR adds the core Autobuilder AWS infrastructure into the stack. This includes the ECS task, service, buckets, and environment variables. Comments have been left throughout by me to call out why certain decisions were made.

These changes also have the enhanced Autobuilder changes in mind. These changes include long-running tasks, the `HandleJobs` function only listening to the `jobUpdatesQueue`, and to have tasks pick up messages from the queue directly. I've also created an ECS service to manage the creation of new tasks if one ends automatically.

To view the resources that were deployed with the latest changes, they can be found [here](https://us-east-2.console.aws.amazon.com/cloudformation/home?region=us-east-2#/stacks/resources?filteringText=&filteringStatus=active&viewNested=true&stackId=arn%3Aaws%3Acloudformation%3Aus-east-2%3A216656347858%3Astack%2Fauto-builder-stack-enhancedApp%2Fed07b830-048d-11ee-b17e-02811f5204f3).

## Tips

One approach I used throughout this PR when recreating the current AWS infrastructure is to reference the CloudFormation templates and serverless.yml files. So, when going through my changes, it might help to also check these files as well. 

## Testing
To verify the CloudFormation is being generated successfully, you can use the `cdk synth` command to generate the CloudFormation. In the `/cdk-infra` directory, execute the following command:

```zsh
npm run cdk synth -- -c enhanced=true -c featureName=enhancedApp > cdk.out/template.yaml     
```

Make sure to update your `~/.aws/credentials` file.

### DIFF
--- a/src/job/jobManager.ts
+++ b/src/job/jobManager.ts
@@ -116,7 +116,7 @@ export class JobManager {
     try {
       this._jobHandler = null;
       if (job?.payload) {
-        const excludeRepoFromBenchmarks = ['mms-docs'].includes(job.payload.repoName);
+        const excludeRepoFromBenchmarks = ['mms-docs', 'docs-k8s-operator'].includes(job.payload.repoName);
         // Can easily rollback with commenting out this flag.
         job.useWithBenchmark = !excludeRepoFromBenchmarks;
         await this.createHandlerAndExecute(job);


Dockerfile.enhanced
```diff
@@ -0,0 +1,120 @@
...
+COPY --from=ts-compiler --chown=docsworker-xlarge /home/docsworker-xlarge/modules/oas-page-builder/dist ./modules/oas-page-builder/
+RUN cd ./modules/oas-page-builder/ && npm install
+
+# Needed for OAS Page Builder module in shared.mk
+ENV REDOC_PATH=${WORK_DIRECTORY}/redoc/cli/index.js
+ENV OAS_MODULE_PATH=${WORK_DIRECTORY}/modules/oas-page-builder/index.js
+
+RUN mkdir repos && chmod 755 repos
+EXPOSE 3000
+CMD ["node", "enhanced/enhancedApp.js"]
```
branberry, line 120: Only difference (for now) between the regular Dockerfile is this

cdk-infra/bin/cdk-infra.ts
```diff
@@ -2,17 +2,52 @@
...
+    /* Uncomment the next line to specialize this stack for the AWS Account
+     * and Region that are implied by the current CLI configuration. */
+    // env: { account: process.env.CDK_DEFAULT_ACCOUNT, region: process.env.CDK_DEFAULT_REGION },
+    /* Uncomment the next line if you know exactly what Account and Region you
+     * want to deploy the stack to. */
+    env: { account: process.env.CDK_DEFAULT_ACCOUNT, region: process.env.CDK_DEFAULT_REGION },
+    workerSecureStrings,
+    webhookSecureStrings,
+    tags: {
+      stackName,
```
branberry, line 47: Added for tracking purposes

cdk-infra/lib/constructs/api/webhook-api-construct.ts
```diff
@@ -130,6 +137,15 @@ export class AutoBuilderApiConstruct extends Construct {
     jobsQueue.grantSendMessages(triggerLocalBuildLambda);
 
     // grant permission for lambdas to enqueue messages to the job updates queue
+    jobUpdatesQueue.grantSendMessages(slackTriggerLambda);
```
branberry, line 140: Added this for the enhanced app changes. Basically, we need to send a message to the `jobUpdatesQueue` for the `inQueue` status. Previously, we would send a message to the `jobsQueue` which would handle this state in the `HandleJobs` function, but since we want long-running tasks, the ECS tasks will be listening to the `jobsQueue` instead.

cdk-infra/utils/ssm.ts
```diff
@@ -0,0 +1,128 @@
+import { GetParameterCommand, SSMClient } from '@aws-sdk/client-ssm';
+
+export function getSsmPathPrefix(env: string): string {
```
branberry, line 3: This code is to work around how SSM handles secure strings

cdk-infra/lib/constructs/worker/buckets-construct.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    docsBucketNames.forEach((bucketName) => {
+      let websiteRoutingRules: RoutingRule[] | undefined;
+
+      if (bucketName === 'docs-mongodb-org') {
+        const hostName = getHostUrl(env);
+        const prefixUrl = getPrefixUrl(env);
+        // docs-mongodb-org has specific routing roles that the rest of the buckets do not have
+        websiteRoutingRules = [
+          {
```
branberry, line 25: These rules are based on what is defined in the `buckets.yaml`

cdk-infra/lib/constructs/worker/worker-construct.ts
```diff
@@ -0,0 +1,113 @@
...
+        S3: { service: GatewayVpcEndpointAwsService.S3 },
+      },
+    });
+
+    vpc.addInterfaceEndpoint('EcrDockerEndpoint', {
+      service: InterfaceVpcEndpointAwsService.ECR_DOCKER,
+    });
+
+    vpc.addInterfaceEndpoint('EcrApiEndpoint', {
+      service: InterfaceVpcEndpointAwsService.ECR,
```
branberry, line 40: This isn't a super important point, but I find it interesting and worth sharing nonetheless; these VPC endpoints actually cut back on our current biggest cost with our current AWS infrastructure. 

Basically, it allows us to pull from ECR through a private connection as opposed to going out to the public internet. Since the images are so large, and the costs for going from private to public internet so high, most of what we pay each month is on this.

Also, this change (at least with CDK) was really simple, and the behavior is exactly the same in terms of how we pull from ECR.

cdk-infra/utils/buckets.ts
```diff
@@ -0,0 +1,12 @@
+export const docsBucketNames = [
+  'docs-mongodb-org',
+  'docs-atlas',
+  'docs-atlas-osb',
```
branberry, line 4: I moved this to a separate file to cut back on the size of the `buckets-construct.ts` file, but looking at it now, I could probably just move it to the `buckets-construct.ts` since this array isn't that big. Thoughts?

cdk-infra/utils/ssm.ts
```diff
@@ -0,0 +1,128 @@
...
+  secureStrings: readonly string[],
+  paramToEnvMap: Map<string, string>,
+  resourceName: string
+) {
+  const ssmClient = new SSMClient({ region: process.env.CDK_DEFAULT_REGION });
+
+  const secureStringsMap: Record<string, string> = {};
+
+  await Promise.all(
+    secureStrings.map(async (paramName: string) => {
```
branberry, line 23: I'm not explicitly returning anything, but I want to use map here still so I can use `Promise.all` on the resulting promises that are returned from the map. I don't believe `Promise.all` would work on a `forEach` since `forEach` doesn't return anything, even if the callback function is async.

cdk-infra/utils/ssm.ts
```diff
@@ -0,0 +1,128 @@
...
+  '/fastly/docs/opsmanager/service_id',
+  '/cdn/client/id',
+  '/cdn/client/secret',
+] as const;
+
+type WorkerSecureString = typeof workerSecureStrings[number];
+
+const workerParamPathToEnvName = new Map<WorkerSecureString, string>();
+
+workerParamPathToEnvName.set('/npm/auth', 'NPM_BASE_64_AUTH');
```
branberry, line 78: In hindsight, I am not entirely sure why I went with a `Map` object as opposed to a regular object, but oh well haha!

cdk-infra/lib/constructs/worker/worker-construct.ts
```diff
@@ -0,0 +1,112 @@
...
+    });
+
+    const executionRole = new Role(this, 'fargateExecutionRole', {
+      assumedBy: new ServicePrincipal('ecs-tasks.amazonaws.com'),
+    });
+
+    executionRole.addToPolicy(executionRoleSsmPolicy);
+
+    const containerProps: AssetImageProps = {
+      file: isEnhanced ? 'Dockerfile.enhanced' : undefined,
```
rayangler, line 78: (Asking for context) Wondering if we would need to consider the `Dockerfile.legacy` file as well for legacy builds (either now or in the future), or if this is just unnecessary altogether?

cdk-infra/lib/constructs/worker/worker-construct.ts
```diff
@@ -0,0 +1,112 @@
...
+      buildArgs: {
+        NPM_BASE_64_AUTH: environment.NPM_BASE_64_AUTH,
+        NPM_EMAIL: environment.NPM_EMAIL,
+      },
+    };
+
+    const taskDefLogGroup = new LogGroup(this, 'workerLogGroup');
+    const taskDefinition = new FargateTaskDefinition(this, 'workerTaskDefinition', {
+      cpu: 2048,
+      memoryLimitMiB: 8192,
```
rayangler, line 88: We have different cpu and memory limits based on prd vs. stg/pre-prd env. Is this something we need to consider now? 

(I'm assuming this has all been built with just ensuring this can be deployed in pre-prd only for now, but wanted to make sure regardless!)

cdk-infra/utils/buckets.ts
rayangler: I think it's fine to keep as-is for the sake of organization, but no strong preferences

cdk-infra/README.md
```diff
@@ -1,4 +1,20 @@
-# Welcome to your CDK TypeScript project
+# Docs Platform
+
+## Tips
+
+My approach when recreating the current AWS infrastructure is to reference the CloudFormation templates and serverless.yml files. So, when going through my changes, it might help to also check these files as well.
```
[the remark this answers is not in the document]
branberry: No you're right, lol. I did initially copy and paste from the PR description, and made some small modifications. I'll clean it up some more though

cdk-infra/bin/cdk-infra.ts
```diff
@@ -2,17 +2,52 @@
...
+
+  let stackName = 'auto-builder-stack';
+
+  // If we want to create a specific feature, we will use this name.
+  // NOTE: This value will take precedence over the feature branch name so that
+  // we can deploy and update the same stack for a specific feature between branches.
+  const customFeatureName = app.node.tryGetContext('featureName');
+
+  // If this is a feature branch i.e., it's not master, use this name.
+  const isFeatureBranch = app.node.tryGetContext('isFeature');
```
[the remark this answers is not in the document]
branberry: That is correct. I should clarify/modify these names. I should rename maybe `isFeature` to `isFeatureBranch` and feature name to `customName` or something. The `isFeature` context variable is for when we have feature branches and want to update the stack name to be the branch name. The `featureName` I am using to provision new infrastructure e.g. the enhanced autobuilder.

cdk-infra/lib/constructs/worker/buckets-construct.ts
```diff
@@ -0,0 +1,151 @@
...
+            replaceKey: {
+              prefixWithKey: `${prefixUrl}/atlas/cli/v1.1`,
+            },
+          },
+        ];
+      }
+
+      const featureName = this.node.tryGetContext('featureName');
+
+      const stackBucketName = `${featureName}-${bucketName}-${env}`.toLowerCase();
```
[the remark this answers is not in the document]
branberry: Yes! I should check that.

cdk-infra/lib/constructs/worker/worker-construct.ts
branberry: Hmm, I think that could be something we might need to take into consideration, but I am not familiar enough with the legacy build to know for sure. Will bring this up to others.

cdk-infra/lib/constructs/worker/worker-construct.ts
branberry: Ya! I think for now we are going to keep it as static values since it will currently only be for pre-prd, but it is worth considering making these values parameters in the future if we need to change it depending on the environment.

cdk-infra/bin/cdk-infra.ts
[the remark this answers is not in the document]
rayangler: Oh, gotcha. This makes sense now. Thanks for clarifying for me!

rayangler (approved):
LGTM!

PR #822, opened 2023-05-23
DOP-3710: Staging Prefix

mmeigs:
### Ticket

DOP-3710

### Notes

This enables writers staging to have correct paths for the oas module. 
This PR goes hand-in-hand with [next PR to meta](https://github.com/mongodb/docs-worker-pool/pull/823) to use this data.

Neither will break the process without the other.


rayangler (approved):
LGTM, as long as variable names remain the same (`BRANCH_NAME` vs. `GIT_BRANCH`, from the `meta` PR)

src/job/stagingJobHandler.ts
```diff
@@ -55,7 +55,9 @@ export class StagingJobHandler extends JobHandler {
   prepStageSpecificNextGenCommands(): void {
     if (this.currJob.buildCommands) {
       this.currJob.buildCommands[this.currJob.buildCommands.length - 1] = 'make next-gen-html';
-      this.currJob.buildCommands.push(`make oas-page-build MUT_PREFIX=${this.currJob.payload.mutPrefix}`);
+      this.currJob.buildCommands.push(
```
i80and, line 58: these give me the hives on grounds of potential command line injection concerns, but that's how the code is structured, so *chews fingernails nervously* it's okay for now

i80and (approved):

src/job/stagingJobHandler.ts
mmeigs: This is such a good point. I can create a check on PROJECT because it should only be cloud-docs that's ever needed... for now. Branch name is harder. I can add a regex check to make sure it can only have the characters allowed in a github branch name, a-zA-Z0-9_/-.

mmeigs: @i80and - the cloud-docs check might be a little silly, but these additions make my mind more at ease. Lmk what you think!

i80and: I'm good with this! The real solution is to not invoke a shell, but this validation is much better than nothing!

PR #823, opened 2023-05-23
Api version meta

mmeigs:
Companion PR to [3710 PR](https://github.com/mongodb/docs-worker-pool/pull/822).

This utilizes the data passed through to replace MUT_PREFIX with correct staging link when MUT_PREFIX is not available.

In writers staging, this was unavailable at this point in the build. 

We have recreated the correct path which is as follows: Project/User/GitBranchName

### DIFF
--- a/makefiles/shared.mk
+++ b/makefiles/shared.mk
@@ -86,5 +86,5 @@ endif
 # Staging: https://github.com/mongodb/docs-worker-pool/blob/42bd36b1f52e49c646a79474c12d299f33d774cb/src/job/stagingJobHandler.ts#L57
 # Production: https://github.com/mongodb/docs-worker-pool/blob/1a482242fa6f1463abb059884cddb2c56ba9fad9/src/job/productionJobHandler.ts#L68
 oas-page-build:
-	echo "oas page build - MUT_PREFIX = ${MUT_PREFIX}, SITE_URL = ${SITE_URL}"
+	echo "oas page build - MUT_PREFIX = ${MUT_PREFIX}, Project/User/Branch = ${PROJECT}/${USER}/${BRANCH_NAME}"
 	node ${OAS_MODULE_PATH} --bundle ${BUNDLE_PATH} --output ${REPO_DIR}/public --redoc ${REDOC_PATH} --repo ${REPO_DIR} --site-url ${SITE_URL}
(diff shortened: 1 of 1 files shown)


makefiles/shared.mk
```diff
@@ -77,9 +77,14 @@ next-gen-stage: ## Host online for review
 	fi
 endif
 
+ifeq ($(MUT_PREFIX),)
+SITE_URL=${URL}/${PROJECT}/${USER}/${BRANCH_NAME}
```
rayangler, line 81: (Nit; non-blocking) Should we consider naming `BRANCH_NAME` -> `GIT_BRANCH` to help consolidate variable names? Or would doing this be a bad idea?

rayangler (approved):
LGTM! I like that this change makes it easy to set the site url as needed

makefiles/shared.mk
```diff
@@ -77,9 +77,14 @@ next-gen-stage: ## Host online for review
...
+else
+SITE_URL=${URL}/${MUT_PREFIX}
+endif
 # Intended to be called by the autobuilder as a build command after frontend build, but before mut upload
 # Staging: https://github.com/mongodb/docs-worker-pool/blob/42bd36b1f52e49c646a79474c12d299f33d774cb/src/job/stagingJobHandler.ts#L57
 # Production: https://github.com/mongodb/docs-worker-pool/blob/1a482242fa6f1463abb059884cddb2c56ba9fad9/src/job/productionJobHandler.ts#L68
 oas-page-build:
-	node ${OAS_MODULE_PATH} --bundle ${BUNDLE_PATH} --output ${REPO_DIR}/public --redoc ${REDOC_PATH} --repo ${REPO_DIR} --site-url ${URL}/${MUT_PREFIX}
-
+	echo "oas page build - MUT_PREFIX = ${MUT_PREFIX}, Project/User/Branch = ${PROJECT}/${USER}/${BRANCH_NAME}"
```
i80and, line 89: It's usually unidiomatic to have this sort of echo in a makefile, since make prints each command as it's run. That said, if we're really sure we want to keep this, the line shoud start with an `@` character to prevent double logging

makefiles/shared.mk
```diff
@@ -77,9 +77,14 @@ next-gen-stage: ## Host online for review
 	fi
 endif
 
+ifeq ($(MUT_PREFIX),)
```
i80and, line 80: Could we have a comment explaining the purpose of this condition?

makefiles/shared.mk
mmeigs: This is a legitimate question. I didn't want to overlap with anything because of the speed of this fix needed and, like MUT_PREFIX, I didn't know if GIT_BRANCH would be available at time of the oas build either.

I can start experimenting with this maybe in the future?

makefiles/shared.mk
mmeigs: Absolutely! Thanks

makefiles/shared.mk
rayangler: Sure, sounds good.

makefiles/shared.mk
i80and: Yep, that would be completely reasonable!

i80and (approved):
Perfect!

PR #825, opened 2023-05-26
Meta cb 2023 timestamp

caesarbell:


makefiles/Makefile.docs
```diff
@@ -53,7 +53,7 @@ next-gen-parse: examples
 		fi \
 	fi
 
-next-gen-html:  get-build-dependencies next-gen-parse 
+next-gen-html:  get-build-dependencies
```
rayangler, line 56: Is `get-build-dependencies` something we need to split out rn as well? Or is it okay to keep as-is based on your changes to the Autobuilder?

caesarbell: @rayangler actually we could split this out rn as well. Because with the auto builder changes it will end up calling it twice because it is caused in the auto builder. But that has me thinking, was that always the case?

rayangler: Might've always been the case then 🤔

caesarbell: It got me thinking as well.

seungpark (approved):

makefiles/Makefile.docs-landing
```diff
@@ -20,6 +20,24 @@ DOCS_TOOLS_THEME=docs-tools/themes/mongodb/static
...
+		if [ $$? -eq 1 ]; then \
+			exit 1; \
+		else \
+			exit 0; \
+		fi \
+	fi
+
 next-gen-html:
 	# snooty parse and then build-front-end
 	snooty build "${REPO_DIR}" --output "${BUNDLE_PATH}" ${RSTSPEC_FLAG}; \
```
seungpark, line 43:
```suggestion

```

can remove this line since we're defining above

caesarbell: @seungpark I am not against this request, but let's just get this out first and make sure it works. I can create a separate branch and remove this line and run some sanity checks.

PR #826, opened 2023-05-30
DOP-3598

caesarbell:
Fired off by Slack

[workerpoolstaging-qgeyp.mongodbstitch.com](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=647e20b1eeea0b3e8073f379)

image

### DIFF
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,7 +3,6 @@ on:
     branches:
       - "master"
       - "integration"
-      - "DOP-3598-ext"
 concurrency:
   group: environment-stg-${{ github.ref }}
   cancel-in-progress: true


src/job/jobHandler.ts
```diff
@@ -234,21 +234,94 @@ export abstract class JobHandler {
...
+    if (resp.status != 'success') {
+      const error = new AutoBuilderError(resp.error, 'BuildError');
+      await this.logError(error);
+      throw error;
+    }
+  }
+
+  // call this method when we want benchmarks and uses cwd option to call command outside of a one liner.
+  private async callWithBenchmark(command: string, stage: string): Promise<CommandExecutorResponse> {
+    const start = performance.now();
```
mmeigs, line 252: whoaaa, I'd never heard of this! Very cool.

src/services/commandExecutor.ts
```diff
@@ -43,11 +43,17 @@ export interface IGithubCommandExecutor {
...
 
 export class ShellCommandExecutor implements ICommandExecutor {
-  async execute(commands: string[]): Promise<CommandExecutorResponse> {
+  async execute(commands: string[], cwd?: string): Promise<CommandExecutorResponse> {
     const exec = promisify(cp.exec);
     const resp = new CommandExecutorResponse();
     try {
-      const { stdout, stderr } = await exec(commands.join(' && '), { maxBuffer: c.get('MAX_STDOUT_BUFFER_SIZE') });
+      const options: { maxBuffer: number; cwd?: string } = {
+        maxBuffer: c.get('MAX_STDOUT_BUFFER_SIZE'),
```
mmeigs, line 51: dumb nitpick: if you wanted, you could add cwd within the object declaration.
`...(cwd ? { cwd }: {}),` or `...(!!cwd && { cwd })`
But up to you, because your code might be more clear in the long run.

src/job/jobHandler.ts
```diff
@@ -234,21 +234,94 @@ export abstract class JobHandler {
...
+
+  private async exeBuildModified(): Promise<void> {
+    const stages = {
+      ['get-build-dependencies']: 'nextGenBuildExe',
+      ['next-gen-parse']: 'nextGenParserExe',
+      ['next-gen-html']: 'nextGenHTMLExe',
+      ['oas-page-build']: 'nextGenStageExe',
+    };
+
+    const prerequisiteCommands = this.currJob.buildCommands.slice(0, 3);
```
seungpark, line 275: are buildCommands constant? CMIIW, but it seems like they can be different per job and should be found by indexOf

src/job/jobHandler.ts
```diff
@@ -234,21 +234,94 @@ export abstract class JobHandler {
     }
   }
 
+  private async loggingMessage(resp: CommandExecutorResponse): Promise<void> {
+    await this._logger.save(this.currJob._id, `${'(BUILD)'.padEnd(15)}Finished Build`);
```
seungpark, line 238: do we need to input `finished build` for each time we log? seeing this multiple times on the sample log in description. i think we can log these in the exeBuildModified after all jobs have executed

src/repositories/jobRepository.ts
```diff
@@ -93,6 +93,29 @@ export class JobRepository extends BaseRepository {
...
+    const options = { sort: { priority: -1, createdTime: 1 }, returnNewDocument: true };
+    const response = await this.findOneAndUpdate(
+      query,
+      update,
+      options,
+      `Mongo Timeout Error: Timed out while retrieving job`
+    );
+
+    if (!response) {
+      throw new InvalidJobError('JobRepository:getOneQueuedJobAndUpdate unable to update job execution time');
```
mmeigs, line 110: Question: will this end the build? Do we want an error on saving execution time to be a build-ending error?

mmeigs (commented):
So exciting! Mainly just questions!

src/job/jobHandler.ts
seungpark, line 252: do we need to import performance? ie. `import { performance } from 'perf_hooks';`

seungpark (changes requested):
this is neat! TIL about `performance` module 😎  left a few comments below! 

i'm not sure if the posted [queue](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64766b872d51dfe3162fa884) is most recent, but some of the `\n` seem to be not producing new lines. might have to call a new log function(?)]

also, not sure if we want to separate the persistence module call in the [next-gen-html](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L51C1-L55) call in order for us to isolate gatsby builds from MongoDB performance

src/job/jobHandler.ts
```diff
@@ -234,21 +234,94 @@ export abstract class JobHandler {
...
+    await this._logger.save(
+      this.currJob._id,
+      `${'(BUILD)'.padEnd(15)}worker.sh run details:\n\n${resp.output}\n---\n${resp.error}`
+    );
+    if (resp.status != 'success') {
+      const error = new AutoBuilderError(resp.error, 'BuildError');
+      await this.logError(error);
+      throw error;
+    }
```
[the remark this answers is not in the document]
caesarbell: @mmeigs good callout. I definitely can. Going to make these changes.

src/repositories/jobRepository.ts
caesarbell: @mmeigs good question, I believe this will end the build, which seems to handle it [here](https://github.com/mongodb/docs-worker-pool/blob/d14e3a702bdd4436369389716bb6ed3be4be36af/src/job/jobHandler.ts#L495-L496). That is a good question. I can ask the channel but, I wouldn't see a reason for this to be a build-ending error.

src/services/commandExecutor.ts
```diff
@@ -43,11 +43,17 @@ export interface IGithubCommandExecutor {
 }
 
 export class ShellCommandExecutor implements ICommandExecutor {
-  async execute(commands: string[]): Promise<CommandExecutorResponse> {
+  async execute(commands: string[], cwd?: string): Promise<CommandExecutorResponse> {
```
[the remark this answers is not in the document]
caesarbell: > this is neat! TIL about `performance` module 😎 left a few comments below!
> 
> i'm not sure if the posted [queue](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64766b872d51dfe3162fa884) is most recent, but some of the `\n` seem to be not producing new lines. might have to call a new log function(?)]
> 
> also, not sure if we want to separate the persistence module call in the [next-gen-html](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L51C1-L55) call in order for us to isolate gatsby builds from MongoDB performance

Not sure if I understand correctly or missing something, but let me give it short. For the first query. The newlines are appearing you can tell by the `---` (dashes). Below is a copy-paste example 

```"(BUILD)        Finished Build",
"(BUILD)        worker.sh run details:


---
",
"(COMMAND)       make get-build-dependencies run details in cloud-docs",
"(JOB VALUE AFTER UPDATE) {"_id":"64766b872d51dfe3162fa884","title":"Slack deploy: 10gen/cloud-docs/master, by caesarbell","user":"caesarbell","email":"caesar.bell@mongodb.com","status":"inProgress","createdTime":"2023-05-30T21:32:55.169Z","startTime":"2023-05-30T21:34:31.361Z","endTime":null,"priority":1,"nextGenBuildExeStartTime":0,"nextGenBuildExeEndTime":0,"nextGenParserExeStartTime":0,"nextGenParserExeEndTime":0,"nextGenHTMLExeStartTime":0,"nextGenHTMLExeEndTime":0,"nextGenStageExeStartTime":0,"nextGenStageExeEndTime":0,"error":{},"result":null,"payload":{"jobType":"productionDeploy","source":"github","action":"push","repoName":"cloud-docs","branchName":"master","project":"cloud-docs","prefix":"docs-qa/atlas","aliased":false,"urlSlug":"master","isFork":true,"private":true,"isXlarge":true,"repoOwner":"10gen","url":"https://github.com/10gen/cloud-docs","newHead":null,"primaryAlias":false,"stable":true},"logs":["* Starting Job with ID: 64766b872d51dfe3162fa884 and type: productionDeploy","(GIT)          Cloning repository","(GIT)          running fetch","(GIT)          Finished git clone","(BUILD)        Running Build","(BUILD)        running worker.sh","'(PREREQUISITE COMMANDS)'. /venv/bin/activate && cd repos/cloud-docs && rm -f makefile","'(MAKE COMMANDS)'make get-build-dependencies && make next-gen-parse && make next-gen-html && make oas-page-build MUT_PREFIX=docs-qa/atlas/master","(BUILD)        Finished Build","(BUILD)        worker.sh run details:\n\n\n---\n","(COMMAND)       make get-build-dependencies run details in cloud-docs"],"taskId":"85e6d3b0b1794143b6efb9e681af581d"}",
"(BUILD)        Finished Build",
"(BUILD)        worker.sh run details:


---
```

After `"(BUILD)        worker.sh run details:` you can see the dashes and spaces, which indicate the newlines after the `resp.output`.

caesarbell:
> this is neat! TIL about `performance` module 😎 left a few comments below!
> 
> i'm not sure if the posted [queue](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64766b872d51dfe3162fa884) is most recent, but some of the `\n` seem to be not producing new lines. might have to call a new log function(?)]
> 
> also, not sure if we want to separate the persistence module call in the [next-gen-html](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L51C1-L55) call in order for us to isolate gatsby builds from MongoDB performance

> this is neat! TIL about `performance` module 😎 left a few comments below!
> 
> i'm not sure if the posted [queue](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64766b872d51dfe3162fa884) is most recent, but some of the `\n` seem to be not producing new lines. might have to call a new log function(?)]
> 
> also, not sure if we want to separate the persistence module call in the [next-gen-html](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L51C1-L55) call in order for us to isolate gatsby builds from MongoDB performance

Not sure if I understand correctly or missing something, but let me give it short. For the first query. The newlines are appearing you can tell by the `---` (dashes). Below is a copy-paste example 

```"(BUILD)        Finished Build",
"(BUILD)        worker.sh run details:


---
",
"(COMMAND)       make get-build-dependencies run details in cloud-docs",
"(JOB VALUE AFTER UPDATE) {"_id":"64766b872d51dfe3162fa884","title":"Slack deploy: 10gen/cloud-docs/master, by caesarbell","user":"caesarbell","email":"caesar.bell@mongodb.com","status":"inProgress","createdTime":"2023-05-30T21:32:55.169Z","startTime":"2023-05-30T21:34:31.361Z","endTime":null,"priority":1,"nextGenBuildExeStartTime":0,"nextGenBuildExeEndTime":0,"nextGenParserExeStartTime":0,"nextGenParserExeEndTime":0,"nextGenHTMLExeStartTime":0,"nextGenHTMLExeEndTime":0,"nextGenStageExeStartTime":0,"nextGenStageExeEndTime":0,"error":{},"result":null,"payload":{"jobType":"productionDeploy","source":"github","action":"push","repoName":"cloud-docs","branchName":"master","project":"cloud-docs","prefix":"docs-qa/atlas","aliased":false,"urlSlug":"master","isFork":true,"private":true,"isXlarge":true,"repoOwner":"10gen","url":"https://github.com/10gen/cloud-docs","newHead":null,"primaryAlias":false,"stable":true},"logs":["* Starting Job with ID: 64766b872d51dfe3162fa884 and type: productionDeploy","(GIT)          Cloning repository","(GIT)          running fetch","(GIT)          Finished git clone","(BUILD)        Running Build","(BUILD)        running worker.sh","'(PREREQUISITE COMMANDS)'. /venv/bin/activate && cd repos/cloud-docs && rm -f makefile","'(MAKE COMMANDS)'make get-build-dependencies && make next-gen-parse && make next-gen-html && make oas-page-build MUT_PREFIX=docs-qa/atlas/master","(BUILD)        Finished Build","(BUILD)        worker.sh run details:\n\n\n---\n","(COMMAND)       make get-build-dependencies run details in cloud-docs"],"taskId":"85e6d3b0b1794143b6efb9e681af581d"}",
"(BUILD)        Finished Build",
"(BUILD)        worker.sh run details:


---
```

After `"(BUILD)        worker.sh run details:` you can see the dashes and spaces, which indicate the newlines after the `resp.output`.

seungpark:
> > also, not sure if we want to separate the persistence module call in the next-gen-html call in order for us to isolate gatsby builds from MongoDB performance
> 
> For this, I believe we needed to separate out the parsing and building. Unless I understood the relationship between `next-gen-parse` and `next-gen-html` wrong.

ah i'm mentioning this because `next-gen-html` also contains the [persistence command](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L55). thinking we should separate this to a separate make command and call before next-gen-html(?)

src/services/commandExecutor.ts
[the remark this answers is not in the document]
caesarbell: @seungpark I think you can find it being passed or being set [here](https://github.com/mongodb/docs-worker-pool/blob/d14e3a702bdd4436369389716bb6ed3be4be36af/src/job/jobHandler.ts#L253-L254).

```
const resp = await this._commandExecutor.execute([command], `repos/${this.currJob.payload.repoName}`);
```

and yes that is correct, it is for setting the current working directory.

seungpark:
> > this is neat! TIL about `performance` module 😎 left a few comments below!
> > i'm not sure if the posted [queue](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64766b872d51dfe3162fa884) is most recent, but some of the `\n` seem to be not producing new lines. might have to call a new log function(?)]
> > also, not sure if we want to separate the persistence module call in the [next-gen-html](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L51C1-L55) call in order for us to isolate gatsby builds from MongoDB performance
> 
> > this is neat! TIL about `performance` module 😎 left a few comments below!
> > i'm not sure if the posted [queue](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64766b872d51dfe3162fa884) is most recent, but some of the `\n` seem to be not producing new lines. might have to call a new log function(?)]
> > also, not sure if we want to separate the persistence module call in the [next-gen-html](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L51C1-L55) call in order for us to isolate gatsby builds from MongoDB performance
> 
> Not sure if I understand correctly or missing something, but let me give it short. For the first query. The newlines are appearing you can tell by the `---` (dashes). Below is a copy-paste example
> 
> ```
> "(BUILD)        worker.sh run details:
> 
> 
> ---
> ",
> "(COMMAND)       make get-build-dependencies run details in cloud-docs",
> "(JOB VALUE AFTER UPDATE) {"_id":"64766b872d51dfe3162fa884","title":"Slack deploy: 10gen/cloud-docs/master, by caesarbell","user":"caesarbell","email":"caesar.bell@mongodb.com","status":"inProgress","createdTime":"2023-05-30T21:32:55.169Z","startTime":"2023-05-30T21:34:31.361Z","endTime":null,"priority":1,"nextGenBuildExeStartTime":0,"nextGenBuildExeEndTime":0,"nextGenParserExeStartTime":0,"nextGenParserExeEndTime":0,"nextGenHTMLExeStartTime":0,"nextGenHTMLExeEndTime":0,"nextGenStageExeStartTime":0,"nextGenStageExeEndTime":0,"error":{},"result":null,"payload":{"jobType":"productionDeploy","source":"github","action":"push","repoName":"cloud-docs","branchName":"master","project":"cloud-docs","prefix":"docs-qa/atlas","aliased":false,"urlSlug":"master","isFork":true,"private":true,"isXlarge":true,"repoOwner":"10gen","url":"https://github.com/10gen/cloud-docs","newHead":null,"primaryAlias":false,"stable":true},"logs":["* Starting Job with ID: 64766b872d51dfe3162fa884 and type: productionDeploy","(GIT)          Cloning repository","(GIT)          running fetch","(GIT)          Finished git clone","(BUILD)        Running Build","(BUILD)        running worker.sh","'(PREREQUISITE COMMANDS)'. /venv/bin/activate && cd repos/cloud-docs && rm -f makefile","'(MAKE COMMANDS)'make get-build-dependencies && make next-gen-parse && make next-gen-html && make oas-page-build MUT_PREFIX=docs-qa/atlas/master","(BUILD)        Finished Build","(BUILD)        worker.sh run details:\n\n\n---\n","(COMMAND)       make get-build-dependencies run details in cloud-docs"],"taskId":"85e6d3b0b1794143b6efb9e681af581d"}",
> "(BUILD)        Finished Build",
> "(BUILD)        worker.sh run details:
> 
> 
> ---
> ```
> 
> After `"(BUILD) worker.sh run details:` you can see the dashes and spaces, which indicate the newlines after the `resp.output`.

the line that starts with `"(JOB VALUE AFTER UPDATE)` has a few `\n`. thinking we can actually remove this since its the queue object. wdyt?

src/job/jobHandler.ts
caesarbell: The `buildCommadns` seem to be constant in order, i.e. prerequisites come first and keep the order following the [preBuildCommands](https://github.com/mongodb/docs-worker-pool/blob/d14e3a702bdd4436369389716bb6ed3be4be36af/src/job/jobHandler.ts#L399). The `make html` is replaced with the first make command, it is treated as the placeholder if I understood the code correctly.

We could use indexOf if the order isn't consistent and could change in the future.

src/job/jobHandler.ts
caesarbell: @seungpark we could, good call out. I can move this out so it isn't logged each time a build step is done.

src/job/jobHandler.ts
caesarbell: @seungpark actually we don't. `performance` is native to node and browser.  We can use it just like `fetch` or `console.log`. Good question though.

src/services/commandExecutor.ts
caesarbell: After looking at the node docs for [exe](https://nodejs.org/api/child_process.html#child_processexeccommand-options-callback) I thnk it might be a cleaner and safer approach to declare it outside of the object, because when no `cwd` passed in as a param then the option will just be the `maxBuffer`, leaving the `cwd` in a default state which is the current working directory `.` The `cwd` option can either be a string or a URL. Let me know if that makes sense.

src/job/jobHandler.ts
seungpark: seems like we push commands depending on stage/production? https://github.com/mongodb/docs-worker-pool/blob/d14e3a702bdd4436369389716bb6ed3be4be36af/src/job/productionJobHandler.ts#L65 and https://github.com/mongodb/docs-worker-pool/blob/d14e3a702bdd4436369389716bb6ed3be4be36af/src/job/stagingJobHandler.ts#L55

seungpark:
> 



> > > > also, not sure if we want to separate the persistence module call in the next-gen-html call in order for us to isolate gatsby builds from MongoDB performance
> > > 
> > > 
> > > For this, I believe we needed to separate out the parsing and building. Unless I understood the relationship between `next-gen-parse` and `next-gen-html` wrong.
> > 
> > 
> > ah i'm mentioning this because `next-gen-html` also contains the [persistence command](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L55). thinking we should separate this to a separate make command and call before next-gen-html(?)
> 
> I think there might have been miscommunication on my end. When developing this it was pointing to this [make file](https://github.com/mongodb/docs-worker-pool/blob/meta-cb-2023-timestamp/makefiles/shared.mk#L52). There is an open PR for that one as well to merge in. Sorry for the miscommunication. I also just added you as a reviewer for that one too for transparency.

this is still calling persistence module within the [next-gen-html](https://github.com/mongodb/docs-worker-pool/pull/825/files#diff-dd33e2da23dc9e9d3cba9d8ec8f81cb303a5ecb9182d22f601c8579f4fbf79acR59), which was my original concern. also a confusion: if we are removing `next-gen-parse` as a prereq, do we have to explicitly have it as a command? did not see it within the changes CMIIW

src/job/jobHandler.ts
caesarbell: That is correct. Now the make commands are diff based on the environment as you have shown above, and for the make commands we are using indexOf which you can find [here](https://github.com/mongodb/docs-worker-pool/blob/60f301993dbfa3f6a80eeb20be99c09cc16d8b80/src/job/jobHandler.ts#L309-L310). We just don't use them for the prerequisite commands, because that order doesn't change.

caesarbell:
> > 
> 
> > > > > also, not sure if we want to separate the persistence module call in the next-gen-html call in order for us to isolate gatsby builds from MongoDB performance
> > > > 
> > > > 
> > > > For this, I believe we needed to separate out the parsing and building. Unless I understood the relationship between `next-gen-parse` and `next-gen-html` wrong.
> > > 
> > > 
> > > ah i'm mentioning this because `next-gen-html` also contains the [persistence command](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L55). thinking we should separate this to a separate make command and call before next-gen-html(?)
> > 
> > 
> > I think there might have been miscommunication on my end. When developing this it was pointing to this [make file](https://github.com/mongodb/docs-worker-pool/blob/meta-cb-2023-timestamp/makefiles/shared.mk#L52). There is an open PR for that one as well to merge in. Sorry for the miscommunication. I also just added you as a reviewer for that one too for transparency.
> 
> this is still calling persistence module within the [next-gen-html](https://github.com/mongodb/docs-worker-pool/pull/825/files#diff-dd33e2da23dc9e9d3cba9d8ec8f81cb303a5ecb9182d22f601c8579f4fbf79acR59), which was my original concern. also a confusion: if we are removing `next-gen-parse` as a prereq, do we have to explicitly have it as a command? did not see it within the changes CMIIW

Yes, so we still need persistence and still calling it. it was added to the make commands found [here](https://github.com/mongodb/docs-worker-pool/blob/25bbe26f1efc7f26a45b7196c4e72b5a2b613cfc/src/job/jobHandler.ts#L311-L312). I hope this clears things up. So to recap, we decoupled it from the make files and now called it its only make file within the make file commands. Originally we would call the `make next-gen-html` which that rule would call the `next-gen-parse` first then run the commands within itself. Now we separated them out and call the `make next-gen-parse` first separately and then next called the `next-gen-html`. We separated the commands to be able to benchmark each one. 

Let me know if this makes sense.

src/services/commandExecutor.ts
mmeigs: I think either way works!! No worries on this at all!

src/repositories/jobRepository.ts
```diff
@@ -93,6 +93,29 @@ export class JobRepository extends BaseRepository {
...
+    } else if (response.value) {
+      const job: Job = response.value;
+      await this.notify(job._id, c.get('jobUpdatesQueueUrl'), JobStatus.inProgress, 0);
```
rayangler, line 113: I don't think we need to have this notification. This current job should already be in progress

src/repositories/jobRepository.ts
mmeigs: Great! Yeah, I think we're all on the same page with this. So, this could just be an error log rather than a throw.

src/repositories/jobRepository.ts
```diff
@@ -93,6 +93,29 @@ export class JobRepository extends BaseRepository {
...
+      return job;
```
rayangler, line 114: Do we need to return a `job` here? It doesn't seem like we're doing anything with it when this function is called

src/job/jobHandler.ts
```diff
@@ -234,21 +234,94 @@ export abstract class JobHandler {
...
+    await this._jobRepository.findOneAndUpdateExecutionTime(this.currJob._id, update);
+    return resp;
+  }
+
+  private async exeBuildModified(): Promise<void> {
+    const stages = {
+      ['get-build-dependencies']: 'nextGenBuildExe',
+      ['next-gen-parse']: 'nextGenParserExe',
+      ['next-gen-html']: 'nextGenHTMLExe',
+      ['oas-page-build']: 'nextGenStageExe',
```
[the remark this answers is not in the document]
caesarbell: @rayangler good callout. I am not 100% sold on the `nextgen` convention. I can update them to rather say

- buildDepsExe
- parseExe
- htmlExe
- oasPageBuildExe

Thoughts?

src/repositories/jobRepository.ts
caesarbell: @rayangler good callout. I will update this.

src/repositories/jobRepository.ts
caesarbell: @rayangler good catch, this was based on the original logic, where I was logging the job.

seungpark:
> https://github.com/mongodb/docs-worker-pool/pull/826#issuecomment-1570632125

discussed offline. will follow up with an issue to decouple persistence module to a separate make command and time these separately

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
...
+    };
+
+    // get the prerequisite commands which should be all commands up to `rm -f makefile`
+    const endOfPrerequisiteCommands = this.currJob.buildCommands.indexOf('rm -f makefile');
+    const index = endOfPrerequisiteCommands + 1;
+    const prerequisiteCommands = this.currJob.buildCommands.slice(0, index);
+    await this._logger.save(this.currJob._id, `'(PREREQUISITE COMMANDS)'${prerequisiteCommands.join(' && ')}`);
+
+    const makeCommands = this.currJob.buildCommands.slice(index);
+    await this._logger.save(this.currJob._id, `'(MAKE COMMANDS)'${makeCommands.join(' && ')}`);
```
seungpark, line 281:
```suggestion
```

can remove these commands logs before merge (totally fine with leaving them on for transparency as well)

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
...
+
+    // call prerequisite commands
+    const prerequisiteResp = await this._commandExecutor.execute(prerequisiteCommands);
+    await this.loggingMessage(prerequisiteResp);
```
seungpark, line 285: is this required? seems like these are
```
      `. /venv/bin/activate`,
      `cd repos/${this.currJob.payload.repoName}`,
      `rm -f makefile`,
```
and not sure if we want to log this as well as including them with the log message `worker.sh run details: ` as in `loggingMessage`

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
     }
   }
 
+  private async loggingMessage(resp: CommandExecutorResponse): Promise<void> {
```
seungpark, line 237: minor. would rename to `logBuildDetails`

src/job/jobHandler.ts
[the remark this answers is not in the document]
caesarbell: @rayangler good call out. It is better to keep things consistent for sure. I will make the changes to this.

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
...
-      await this._logger.save(this.currJob._id, `${'(BUILD)'.padEnd(15)}Finished Build`);
-      await this._logger.save(
-        this.currJob._id,
-        `${'(BUILD)'.padEnd(15)}worker.sh run details:\n\n${resp.output}\n---\n${resp.error}`
-      );
-      if (resp.status != 'success') {
-        const error = new AutoBuilderError(resp.error, 'BuildError');
-        await this.logError(error);
-        throw error;
+      // Using indexOf to handle between the two environments.
```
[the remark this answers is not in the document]
caesarbell: Oh sorry. I will make this more clear, but I was referring to the stagingJobHandler and productionJobHandler's `prepStageSpecificNextGenCommands` method.

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
...
+      const insertionPoint = this.currJob.buildCommands.indexOf('make next-gen-html');
+      if (insertionPoint !== -1) {
+        this.currJob.buildCommands.splice(insertionPoint, 0, 'make next-gen-parse');
```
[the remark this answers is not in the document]
caesarbell: @rayangler my thought was to keep the changes within one file or method. This way if the `prepStageSpecificNextGenCommands` order needed to change the `executeBuild` would be flexible enough to adapt if need be.

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
...
+    await this._logger.save(
+      this.currJob._id,
+      `${'(COMMAND)'.padEnd(15)} ${command} run details in ${this.currJob.payload.repoName}`
+    );
+    const end = performance.now();
+    const update = {
+      [`${stage}StartTime`]: start,
+      [`${stage}EndTime`]: end,
+    };
+    await this._jobRepository.findOneAndUpdateExecutionTime(this.currJob._id, update);
```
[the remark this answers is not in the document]
caesarbell: @seungpark good callout. I will look into how [1.5s of awaiting](https://github.com/mongodb/docs-worker-pool/blob/8f12a3220057dcdec30241fd9493c9fdbb7b0a8b/src/repositories/baseRepository.ts#L19-L26) is used within the app and follow that pattern.

src/job/jobHandler.ts
```diff
@@ -234,21 +234,90 @@ export abstract class JobHandler {
...
+      ['oas-page-build']: 'oasPageBuildExe',
+    };
+
+    // get the prerequisite commands which should be all commands up to `rm -f makefile`
+    const endOfPrerequisiteCommands = this.currJob.buildCommands.indexOf('rm -f makefile');
+    const index = endOfPrerequisiteCommands + 1;
+    const prerequisiteCommands = this.currJob.buildCommands.slice(0, index);
+    await this._logger.save(this.currJob._id, `'(PREREQUISITE COMMANDS)'${prerequisiteCommands.join(' && ')}`);
+
+    const makeCommands = this.currJob.buildCommands.slice(index);
```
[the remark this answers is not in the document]
caesarbell: @rayangler that would actually be a pretty cool approach. Would this be something we would want to address in another ticket? Or would it be worth implementing in this ticket?

src/job/jobHandler.ts
caesarbell: I thought it would be good visibility on the separation. Out of curiosity what would be the harm in keeping those comments in the log?

src/job/jobHandler.ts
caesarbell: @seungpark I am not sure what you are asking here, sorry.

As for the logging, I guess we don't need to log those as there will not be an output. I can no longer pass the response to the logger message method.

src/job/jobHandler.ts
caesarbell: I will make the change to the method name.

src/job/jobHandler.ts
seungpark: no harm! could keep in log if necessary as well. seemed more of debug logs to list out the commands

src/job/jobHandler.ts
seungpark: the `prerequisiteResp` are the ones listed above, and i suspect this will be logged as 

```
"(BUILD)        worker.sh run details:

<result of cd and rm>        <------------------- is this relevant?
---
"(BUILD)        worker.sh run details:

<result of make commands>

```

src/job/jobHandler.ts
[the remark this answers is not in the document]
rayangler: That makes sense. I'm not too sure how flexible keeping this logic in `executeBuild` would be though. If in the future we want to split out the call to the persistence module as a separate command, we would definitely need to add a new build command to accommodate that, and this logic would still need to change. My preference in terms of readability would be to have the commands listed out together, rather than having them be inserted elsewhere if they don't need to be.

If it makes more sense to separate it out like this, a short comment explaining why the commands are split like this would probably be appreciated by future us!

src/job/jobHandler.ts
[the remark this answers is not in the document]
rayangler: I'm okay with either, but this probably isn't something that needs to block the PR. We can keep things as-is

src/job/jobHandler.ts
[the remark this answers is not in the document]
caesarbell: @rayangler taking a closer look. I see for other commands they are using single quotes. Would we want to keep consistent with that? However, I did notice I should add the `padEnd` to keep the padding at the end consistent.

src/job/jobHandler.ts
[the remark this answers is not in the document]
caesarbell: @rayangler good callout. I will keep the commands together and move them into the `prepStageSpecificNextGenCommands` methods.

src/job/jobHandler.ts
caesarbell: @seungpark that is a fair assessment, I can see how they would be or could be more used for debugging logs. But wouldn't it be useful to have these debugging commands if we ever need to debug or develop around those commands?

src/job/jobHandler.ts
caesarbell: @seungpark that is fair, probably no value in logging those. I will not pass the response from the prerequisite commands. Good callout.

src/job/jobHandler.ts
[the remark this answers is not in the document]
caesarbell:
```"(PREREQUISITE COMMANDS) . /venv/bin/activate && cd repos/cloud-docs && rm -f makefile",
"(MAKE COMMANDS) make get-build-dependencies && make next-gen-parse && make next-gen-html && make oas-page-build MUT_PREFIX=docs-qa/atlas/master",

It is more consistent now, thanks again for that callout.

File: src/repositories/jobRepository.ts [hunk 18]
```diff
@@ -93,6 +93,15 @@ export class JobRepository extends BaseRepository {
     await this._queueConnector.sendMessage(new JobQueueMessage(jobId, status, 0, taskId), url, delay);
   }
 
+  async findOneAndUpdateExecutionTime(id: string, setValues: { [key: string]: number }): Promise<void> {
+    const query = {
+      _id: new objectId(id),
+    };
+    const update = { $set: setValues };
+    const options = { sort: { priority: -1, createdTime: 1 }, returnNewDocument: true };
+    this.findOneAndUpdate(query, update, options, `Mongo Timeout Error: Timed out while retrieving job`);
```
rayangler, line 102: Are we sure we don't need to add a `.catch()` method in case `findOneAndUpdate` throws an error? Or will an error from a promise automatically be ignored?

src/job/jobHandler.ts
seungpark: i am fine with leaving as is :)

rayangler (commented):
One last question just to make sure potential errors are a non-issue from the promise. But otherwise LGTM!

src/repositories/jobRepository.ts
caesarbell: @rayangler good question, from what I understand about Promises with error handling is that if an error occurs it will jump to the nearest catch in our cases it should be handle within this [try/catch](https://github.com/mongodb/docs-worker-pool/blob/85c7b8005894f16d3bf2bfa607394bb9e7b57701/src/repositories/baseRepository.ts#L131).

So nope, it shouldn't be ignored. Instead the handling the error in a catch within the `findOneAndUpdate` method in the `jobRepository` file, it will just use the catch inside the `findOneAndUpdate` inside the `baseRepository` file.

rayangler: It looks like `findOneAndUpdate` catches the error then throws it again. Wouldn't that additional `throw` propagate to a different try/catch block? Just want confirmation that an error here won't cause further errors elsewhere

seungpark: i think this will lead to an `unhandled rejection error`. sorry for the shortsighted suggestion @caesarbell 

thinking of a way we can avoid awaiting but also catch errors? this would require some separate queue/service. think for now,
we can return findOneAndUpdate here

and when we call findOneAndUpdateExecutionTime, we can wrap with 
```
try {
  findOneAndUpdateExecutionTime
} catch(e) {
  logger.save()
  // but continue the code
}
```

rayangler: If we're planning to not `await`, might need to do something like:

```
findOneAndUpdateExecutionTime().catch((e) => {
   logger()
   // but continue the code
})
```

I don't think `try`/`catch` can catch non-awaited errors. 

Or, we can do it within the `findOneAndUpdateExecutionTime` function itself so that the function is always guaranteed to be non-blocking if we ever need to use it in more than one place in the future.

```
this.findOneAndUpdate(query, update, options, `Mongo Timeout Error: Timed out while retrieving job`).catch((e) => {
   logger()
   // but continue the code
})
```

caesarbell: @rayangler @seungpark within the `callWithBenchmark` can we just handle the error here? Example like this 

```
try {
      await this._jobRepository.findOneAndUpdateExecutionTime(this.currJob._id, update);
    } catch (error) {
      this._logger.error(this._currJob._id, error.message);
    }
```

seungpark: > If we're planning to not `await`, might need to do something like:
> 
> ```
> findOneAndUpdateExecutionTime().catch((e) => {
>    logger()
>    // but continue the code
> })
> ```
> 
> I don't think `try`/`catch` can catch non-awaited errors.
> 
> Or, we can do it within the `findOneAndUpdateExecutionTime` function itself so that the function is always guaranteed to be non-blocking if we ever need to use it in more than one place in the future.
> 
> ```
> this.findOneAndUpdate(query, update, options, `Mongo Timeout Error: Timed out while retrieving job`).catch((e) => {
>    logger()
>    // but continue the code
> })
> ```

catching and returning within `findOneAndUpdateExecutionTime` sounds ideal. it can be non blocking for rest of logic besides logging the times 👍

caesarbell: We will have the following 

```
this._jobRepository.findOneAndUpdateExecutionTime(
      query,
      update,
      options,
      `Mongo Timeout Error: Timed out while updating job with timestamps for ${this.currJob._id}`
    ).catch(error => {
      this._logger.save(this.currJob._id, error.message)
    })
``` 
and 

```
async findOneAndUpdateExecutionTime<Query, UpdateValues, Options>(query: Query, update: UpdateValues, options: Options, errorMsg: string): Promise<void> {
    this.findOneAndUpdate(query, update, options, errorMsg);
  }
```

rayangler (approved):
LGTM! I don't think we need to await the updating of execution time, as discussed previously. But also not a blocker if we keep as-is

seungpark (approved):
changes LGTM. Thanks a ton @caesarbell !
note: we'll need the changes to `meta` branch before merge

mmeigs (approved):

PR #827, opened 2023-06-02
DOP-3750: add makefile for docs-relational-migrator

seungpark:
adds makefile and yaml file for build dependencies as per [instructions](https://wiki.corp.mongodb.com/display/DE/Set+up+a+new+repo+to+build)


casthewiz (commented):
No need to add a published branches entry or the get-build-dependencies step - they are relics of a mostly bygone era 😄

rayangler:
> No need to add a published branches entry or the get-build-dependencies step - they are relics of a mostly bygone era 😄

Our [jobHandlers](https://github.com/mongodb/docs-worker-pool/blob/8f12a3220057dcdec30241fd9493c9fdbb7b0a8b/src/job/productionJobHandler.ts#L67) still call `make get-build-dependencies`. Would excluding it result in an Autobuilder error?

casthewiz:
> > No need to add a published branches entry or the get-build-dependencies step - they are relics of a mostly bygone era 😄
> 
> Our [jobHandlers](https://github.com/mongodb/docs-worker-pool/blob/8f12a3220057dcdec30241fd9493c9fdbb7b0a8b/src/job/productionJobHandler.ts#L67) still call `make get-build-dependencies`. Would excluding it result in an Autobuilder error?

Ah, I thought we had a base `get-build-dependencies` in the shared.mk, but I am mistaken 😬 

I'm 💯 that we don't need to have a published branches yaml entry though - everything should flow off of the `published_branches` collection

Things are kinda weird in some of our other makefiles to get around this. for example, https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/Makefile.docs-atlas-cli has a `get-build-dependencies` but gets `mongocli`'s yaml.

@seungpark and @rayangler maybe the best way to handle this is to remove the published branches and the curl, and instead do something like `@echo "No Dependencies to fetch"`? I am not the best at makefiles, so not 💯 on if this will be sufficient

seungpark:
> No need to add a published branches entry or the get-build-dependencies step - they are relics of a mostly bygone era 😄



> 



> > > No need to add a published branches entry or the get-build-dependencies step - they are relics of a mostly bygone era 😄
> > 
> > 
> > Our [jobHandlers](https://github.com/mongodb/docs-worker-pool/blob/8f12a3220057dcdec30241fd9493c9fdbb7b0a8b/src/job/productionJobHandler.ts#L67) still call `make get-build-dependencies`. Would excluding it result in an Autobuilder error?
> 
> Ah, I thought we had a base `get-build-dependencies` in the shared.mk, but I am mistaken 😬
> 
> I'm 💯 that we don't need to have a published branches yaml entry though - everything should flow off of the `published_branches` collection
> 
> Things are kinda weird in some of our other makefiles to get around this. for example, https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/Makefile.docs-atlas-cli has a `get-build-dependencies` but gets `mongocli`'s yaml.
> 
> @seungpark and @rayangler maybe the best way to handle this is to remove the published branches and the curl, and instead do something like `@echo "No Dependencies to fetch"`? I am not the best at makefiles, so not 💯 on if this will be sufficient

perfect. instead of curling false info 👍

rayangler:
An echo sounds good to me. I can file a ticket to actually remove that command from the Autobuilder if we're sure no Makefiles need them

makefiles/Makefile.docs-relational-migrator
```diff
@@ -0,0 +1,21 @@
...
+REPO_DIR=$(shell pwd)
+
+SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
+SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
+
+include ~/shared.mk
+
+
+get-build-dependencies: 
+	@echo "No Dependencies to fetch
```
casthewiz, line 21: small suggestion since my original example probably wasn't the best 😅 
```suggestion
get-build-dependencies: 
	$(info    No dependencies to fetch)
```

casthewiz:
> An echo sounds good to me. I can file a ticket to actually remove that command from the Autobuilder if we're sure no Makefiles need them

There are some makefiles like manual that curl other dependencies beyond the published branches yaml, but it might be good to have a stub command in the shared.mk that just info's out that there's nothing to fetch.

makefiles/Makefile.docs-relational-migrator
seungpark: thats on me for not checking... thanks for the catch

seungpark: i opted to use echo instead of `$( )` as found in https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/Makefile.docs-csharp

schmalliso:
The docs-atlas-cli fetching unneeded YAML was definitely for expediency rather than actually making sense, since we were also finishing the removal of the published branches files from the tests in the autobuilder at the time and I wasn't sure if we needed to have *something* in there.

PR #831, opened 2023-06-07
[DOP-3726]: Update Autobuilder to process messages from queue for enhanced infrastructure

branberry:
## Ticket
[DOP-3726](https://jira.mongodb.org/browse/DOP-3726)

## Notes
This PR incorporates the necessary changes to process jobs using the enhanced infrastructure. To test, please refer to the following [documentation](https://wiki.corp.mongodb.com/display/DE/How-To%3A+Use+Snooty%27s+Autobuilder+to+Build+Your+Content). For the webhook URL, use the following address:

https://90ajwbnu7h.execute-api.us-east-2.amazonaws.com/prod/webhook/githubEndpoint/trigger/build

 Feel free to use any docs site repository that you choose. 
 
 To verify your results, the `pool_test.dotcom_queue` contains the run results of each job. For additional logging, visit the following [ECS Service logs](https://us-east-2.console.aws.amazon.com/ecs/v2/clusters/auto-builder-stack-enhancedApp-workerclusterEBE39B1B-ysZJ1PjROkJq/services/auto-builder-stack-enhancedApp-workerfargateServiceAF0FA7E1-VXY7L71kgUve/logs?region=us-east-2).


cdk-infra/lib/constructs/worker/worker-env-construct.ts
```diff
@@ -55,11 +56,12 @@ export class WorkerEnvConstruct extends Construct {
       PREVIEW_BUILD_ENABLED: previewBuildEnabled,
       USER_ENTITLEMENT_COL_NAME: entitlementCollection,
       NPM_EMAIL: npmEmail,
-      REPO_BRANCHES_COLLECTION: repoBranchesCollection,
+      REPO_BRANCHES_COL_NAME: repoBranchesCollection,
```
branberry, line 59: this was the wrong env name 😱

cdk-infra/lib/constructs/worker/worker-construct.ts
```diff
@@ -84,7 +84,7 @@ export class WorkerConstruct extends Construct {
 
     const taskDefLogGroup = new LogGroup(this, 'workerLogGroup');
     const taskDefinition = new FargateTaskDefinition(this, 'workerTaskDefinition', {
-      cpu: 2048,
+      cpu: 4096,
```
branberry, line 87: increased the CPU amount because it appeared that at 2 vCPUs (which is what a `cpu` value of 2048 represents) that the tasks were maxing out their CPU usage.

src/enhanced/enhancedApp.ts
```diff
@@ -1,8 +1,15 @@
...
 async function app() {
   console.log('starting application');
-  await listenToJobQueue();
+  const { jobId } = await listenToJobQueue();
+
+  await handleJob(jobId);
+
+  console.log('process completed');
+
+  process.exit(0);
```
branberry, line 12: Not entirely sure if this is necessary, but I think I noticed that some jobs weren't restarting, so throwing this to ensure the process does end.

seungpark (commented):
thanks for putting this together and giving me the run down for the stack!

one minor note while i was testing this yesterday:

seems like we set AWS auth in the [Dockerfile](https://github.com/mongodb/docs-worker-pool/blob/master/.github/workflows/deploy-stg-ecs.yml#L20-L27)

thus, getting [error](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=648a2c4767a273b816465986) for no staging auth: 

```
# stagel local jobs
if [ -n "" -a "cloud-docs" = "cloud-docs" ]; then \
	mut-publish public enhancedApp-docs-atlas-dotcomstg --prefix="37dbd6b4//cloud-docs" --stage ; \
	echo "Hosted at https://mongodbcom-cdn.website.staging.corp.mongodb.com//37dbd6b4//cloud-docs/docsworker-xlarge/DOP-3726/"; \
# stagel commit jobs and regular git push jobs\
else \
	mut-publish public enhancedApp-docs-atlas-dotcomstg --prefix="cloud-docs" --stage ; \
	echo "Hosted at https://mongodbcom-cdn.website.staging.corp.mongodb.com//cloud-docs/docsworker-xlarge/DOP-3726/"; \
fi
No staging authentication found. Create a file at /home/docsworker-xlarge/.config/giza-aws-authentication.conf with contents like the following:

[authentication]
accesskey=<AWS access key>
secretkey=<AWS secret key>

```

however, this does not result in the build error'ing or having any error messages, just shows the echo'ed message

api/controllers/v2/jobs.ts
```diff
@@ -65,20 +65,15 @@ export const HandleJobs = async (event: SQSEvent): Promise<void> => {
...
             await NotifyBuildProgress(jobId);
             break;
           case JobStatus[JobStatus.timedOut]:
             await NotifyBuildSummary(jobId);
             const taskId = body['taskId'];
+            // for the enhanced application, the taskId will never be defined
+            // as we are not saving it at this time
             if (taskId) {
               await stopECSTask(taskId, consoleLogger);
             }
```
seungpark, line 79: completely off topic, but wonder why we don't handle `failed` jobStatus here

src/enhanced/job/enhancedJobHandlerFactory.ts
```diff
@@ -0,0 +1,57 @@
...
+    fileSystemServices: IFileSystemServices,
+    commandExecutor: IJobCommandExecutor,
+    cdnConnector: ICDNConnector,
+    repoConnector: IRepoConnector,
+    logger: IJobRepoLogger,
+    validator: IJobValidator,
+    repoBranchesRepo: RepoBranchesRepository
+  ): JobHandler {
+    const jt = job.payload?.jobType;
+    if (jt in enhancedJobHandlerMap) {
```
seungpark, line 41: minor nitpick. would throw if not found first and return after

api/controllers/v2/jobs.ts
branberry: We do handle the `failed` jobStatus [here](https://github.com/mongodb/docs-worker-pool/blob/086075234892f14a2ccc07bd003b29b03404bbad/api/controllers/v2/jobs.ts#L81)

src/enhanced/job/enhancedJobHandlers.ts
```diff
@@ -0,0 +1,56 @@
...
+import { ProductionJobHandler } from '../../job/productionJobHandler';
+import { RegressionJobHandler } from '../../job/regressionJobHandler';
+import { StagingJobHandler } from '../../job/stagingJobHandler';
+
+/**
+ * This method overrides the JobHandler's current setEnvironmentVariablesMethod so that it points to
+ * the custom buckets that are used in the enhanced app.
+ * @param this reference to current object
+ */
+async function setEnvironmentVariablesEnhanced(this: JobHandler) {
```
seungpark, line 12: minor nitpick. would rename `this` parameter to something like `jobHandler`

api/controllers/v2/slack.ts
```diff
@@ -99,6 +100,13 @@ export const DisplayRepoOptions = async (event: APIGatewayEvent): Promise<APIGat
 async function deployRepo(deployable: Array<any>, logger: ILogger, jobRepository: JobRepository, jobQueueUrl) {
   try {
     await jobRepository.insertBulkJobs(deployable, jobQueueUrl);
+
+    // we need to also notify the jobUpdatesQueue as well so that slack users get notified
+    await Promise.all(
+      deployable.map(async ({ jobId }) => {
+        await jobRepository.notify(jobId, c.get('jobUpdatesQueueUrl'), JobStatus.inQueue, 0);
```
seungpark, line 107: i see we have some different places for `jobRepository.notify` vs sqs.sendMessage directly. will this be consolidated. ie. can queue collection and sqs events be simultaneous or does it have to be split up ie. [job controller TriggerLocalBuild](https://github.com/mongodb/docs-worker-pool/pull/831/files#diff-c45958abd9415b796244840efc442d191113b562eed4712968ef25fd51429d72R33)

cdk-infra/lib/constructs/api/webhook-env-construct.ts
```diff
@@ -19,15 +19,16 @@ export class WebhookEnvConstruct extends Construct {
     const dbName = StringParameter.valueFromLookup(this, `${ssmPrefix}/atlas/dbname`);
     const dbUsername = StringParameter.valueFromLookup(this, `${ssmPrefix}/atlas/username`);
     const dbHost = StringParameter.valueFromLookup(this, `${ssmPrefix}/atlas/host`);
+    const jobCollection = StringParameter.valueFromLookup(this, `${ssmPrefix}/atlas/collections/job/queue`);
 
-    const dbPassword = secureStrings[`${ssmPrefix}/atlas/password`];
+    const dbPassword = secureStrings['MONGO_ATLAS_PASSWORD'];
     this.environment = {
       ...secureStrings,
```
seungpark, line 26: we have a few places where [env vars are set via Dockerfile.](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk#L58-L66) might be a good place to start centralizing env files

seungpark (commented):
expanding on my previous review!

src/enhanced/job/enhancedJobHandlers.ts
branberry: So, this is really weird, but I think the param name needs to be `this` to access properties of the object that are private or protected for that class. As soon as I change the parameter name as something other than `this` it says it can't access the properties of `_config` for example since that is a protected attribute.

api/controllers/v2/slack.ts
branberry: I am not entirely sure why in certain cases the logic is separated out, but I am not going to consolidate them at this point. Right now, I am trying to as closely mimic the original logic so that I don't introduce unintended bugs or behaviors. Hopefully, this can be addressed at another point.

api/controllers/v2/jobs.ts
branberry: For the failed case, it'll still execute the `NotifyBuildSummary` and `SubmitArchiveJob` I believe

cdk-infra/lib/constructs/api/webhook-env-construct.ts
branberry: I think this is a good idea, but I am worried about how this might impact the current infrastructure if we stop setting environment variables in the makefiles. We might need to think about how we can eventually migrate all env vars over to when we define them for the infra.

src/enhanced/utils/job/index.ts
```diff
@@ -0,0 +1,74 @@
...
+let repoBranchesRepo: RepoBranchesRepository;
+let ssmConnector: ParameterStoreConnector;
+let ssoConnector: ISSOConnector;
+
+export async function handleJob(jobId: string) {
+  const atlasURL = `mongodb+srv://${c.get('dbUsername')}:${c.get('dbPassword')}@${c.get(
+    'dbHost'
+  )}/?retryWrites=true&w=majority`;
+  const client = new mongodb.MongoClient(atlasURL);
+  await client.connect();
```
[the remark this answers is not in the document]
branberry: Now closing connection on `SIGTERM` below.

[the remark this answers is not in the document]
seungpark: we can move that outside handleJob fn as in the [old app sigterm](https://github.com/mongodb/docs-worker-pool/blob/master/src/app.ts#L72)

Dockerfile.enhanced
```diff
@@ -4,19 +4,19 @@ WORKDIR /home/docsworker-xlarge
 COPY  config config/
 COPY package*.json ./
 COPY tsconfig*.json ./
-RUN npm install
+RUN npm ci --legacy-peer-deps
```
seungpark, line 7: minor. this is only required for snooty and should be mitigated by @rayangler 's [PR for npmrc ](https://github.com/mongodb/snooty/pull/842)

seungpark (approved):
minor comment on legacy peer deps. otherwise the stack LGTM 👍  would want another set of eyes as infra is not my forte

with that said, we still need an [update to mut](https://github.com/mongodb/mut/pull/60) to allow S3 publishing

src/entities/job.ts
```diff
@@ -46,6 +46,35 @@ export type Payload = {
...
 };
 
+export type EnhancedPayload = {
+  jobType: string;
+  source: string;
+  action: string;
+  repoName: string;
+  branchName: string;
+  isFork: boolean;
+  private: boolean;
```
rayangler, line 56: (Nit) We can remove this field, as it's been removed in a recent commit

rayangler (approved):
Overall lgtm!

branberry:
@rayangler For the timeout, I set it to an arbitrarily high duration. Previously, the timeout duration was only 3 seconds, so the fact that it was timing out before I don't think is a concern.

PR #834, opened 2023-06-13
Add exclusion array to jobManager.ts

casthewiz:
Adds an exclusion array to fix an issue where `mms-docs` was incompatible with deployment benchmark changes

### DIFF
--- a/src/job/jobManager.ts
+++ b/src/job/jobManager.ts
@@ -116,7 +116,7 @@ export class JobManager {
     try {
       this._jobHandler = null;
       if (job?.payload) {
-        const excludeRepoFromBenchmarks = ['mms-docs'].includes(payload.repoName);
+        const excludeRepoFromBenchmarks = ['mms-docs'].includes(job.payload.repoName);
         // Can easily rollback with commenting out this flag.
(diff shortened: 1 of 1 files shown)


caesarbell (approved):
LGTM

seungpark (approved):

src/job/jobManager.ts
```diff
@@ -116,8 +116,9 @@ export class JobManager {
     try {
       this._jobHandler = null;
       if (job?.payload) {
+        const excludeRepoFromBenchmarks = ['mms-docs'].includes(job.payload.repoName);
```
rayangler, line 119: Do we know if this is limited to `mms-docs`, or should this be for any old-gen properties?

casthewiz: I'm not certain on if `ruby` or `php` driver docs will fail in the same way - I think those are the only old-gen exceptions, now that server manual isn't actively deploying docs built with old-gen?

rayangler: Gotcha. Down to keep it as-is and alter the array if needed

rayangler (approved):

PR #836, opened 2023-06-15
[fix] Add docs-k8s-operator to exclusion array

casthewiz:
Noticed an issue with docs-k8s while trying to investigate a separate issue involving 200 responses but no slack message.

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=648b398f67a273b816827da1

^^ Example job log of failure for docs-k8s-operator.

Performing mitigation in the same vein as mms-docs.


caesarbell (approved):
LGTM - its good we have this ignore array.

PR #839, opened 2023-06-20
DOP-3665: Remove private field from payload

rayangler:
### Ticket

DOP-3665

### Notes

* Removes `private` field from job payload object. I couldn't find anywhere that actually used this field.
* Leaving some of the GH actions warnings as-is for now as the issues are pre-existing and aren't necessarily related to the `private` field. (I also wanted to avoid breaking anything that didn't need to change)


schmalliso (approved):
Yay!

PR #840, opened 2023-06-22
Add explicit client closure to V1 Lambdas

casthewiz:
Adds explicit `client.close()` statements to the jobHandler to mitigate potential thread overuse from dangling connections.

Successful deploy log here:

https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=6494b9aa0379b4804917fa5e

### DIFF
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "close-lambda-mdb-clients"
 concurrency:
   group: environment-stg-${{ github.ref }}
   cancel-in-progress: true
--- a/api/controllers/v1/jobs.ts
+++ b/api/controllers/v1/jobs.ts
@@ -23,12 +23,14 @@ export const TriggerLocalBuild = async (event: any = {}, context: any = {}): Pro
     consoleLogger.info(body.jobId, 'enqueuing Job');
     await sqs.sendMessage(new JobQueueMessage(body.jobId, JobStatus.inQueue), c.get('jobUpdatesQueueUrl'), 0);
     consoleLogger.info(body.jobId, 'Job Queued Job');
+    client.close();
     return {
       statusCode: 202,
       headers: { 'Content-Type': 'text/plain' },
       body: body.jobId,
     };
   } catch (err) {
+    client.close();
     consoleLogger.error('TriggerLocalBuild', err);
     return {
       statusCode: 500,
@@ -106,7 +108,9 @@ export const FailStuckJobs = async () => {
   try {
     const hours = 8;
     await jobRepository.failStuckJobs(hours);
+    client.close();
   } catch (err) {
+    client.close();
     consoleLogger.error('FailStuckJobs', err);
   }
 };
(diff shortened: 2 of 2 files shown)


.github/workflows/deploy-stg-ecs.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "close-lambda-mdb-clients"
```
```suggestion
```

api/controllers/v1/jobs.ts
```diff
@@ -23,12 +23,14 @@ export const TriggerLocalBuild = async (event: any = {}, context: any = {}): Pro
     consoleLogger.info(body.jobId, 'enqueuing Job');
     await sqs.sendMessage(new JobQueueMessage(body.jobId, JobStatus.inQueue), c.get('jobUpdatesQueueUrl'), 0);
     consoleLogger.info(body.jobId, 'Job Queued Job');
+    client.close();
```
seungpark, line 26: do we explicitly need to call `client.connect()` in these cases? we are just getting db info and the promise calls to insert can throw namespace errors later if they don't exist. CMIIW

i've been noticing on [node driver examples](https://www.mongodb.com/docs/drivers/node/v5.1/usage-examples/insertOne/) that they do not explicitly call `client.connect()` and the node source code shows that there is a check for connection with [async executions](https://github.com/mongodb/node-mongodb-native/blob/v5.1.0/src/operations/execute_operation.ts#L82)

that said, above suggestion will require some more testing that nothing breaks, so i am fine with at least explicitly closing connection

seungpark (approved):
thanks for closing our connections according to the [HELP ticket](https://jira.mongodb.org/plugins/servlet/desk/portal/13/HELP-47047)!

minor comment below, which might be reconsideration for calling connect in the first place!

api/controllers/v1/jobs.ts
```diff
@@ -106,7 +108,9 @@ export const FailStuckJobs = async () => {
   try {
     const hours = 8;
     await jobRepository.failStuckJobs(hours);
+    client.close();
   } catch (err) {
+    client.close();
```
rayangler, line 113: (Nit suggestion; non-blocking) Since we're intending to close in both cases, it might be a good practice to wrap this in a `finally` block. This can also be applied in a few other places

api/controllers/v1/jobs.ts
rayangler, line 26: Just curious if the uses of `client.close()` should be awaited, or if it's okay to keep as promise? Wasn't sure if the promise would be resolved before process exited/lambdas returned (or however lambdas finish)

rayangler (commented):
Looks good overall, but just wanted to ask 1 thing before fully lgtm

api/controllers/v1/jobs.ts
casthewiz: Very good point - I don't think we actually need to the more I think about it. But I also think that with the V2 set of lambdas removing the strict dependency on insertion into the `queue` collection for job execution, it's a bit of a moot point - and I'd rather get this fix incorporated until V2 is used for content staging.

casthewiz: Heh - we were thinking the same thing at the same time. Just pushed a commit that adds awaits!

rayangler: Nice timing 🤠

rayangler (approved):
LGTM!

api/controllers/v1/jobs.ts
casthewiz: 👀 Just wrapped it - good catch!

PR #844, opened 2023-06-27
[DOP-3811]: Refactor enhanced app to close gracefully; increase timeo…

branberry:
…ut before SIGKILL

### DIFF
--- a/Dockerfile.enhanced
+++ b/Dockerfile.enhanced
@@ -4,7 +4,7 @@ WORKDIR /home/docsworker-xlarge
 COPY  config config/
 COPY package*.json ./
 COPY tsconfig*.json ./
-RUN npm ci
+RUN npm install
 COPY . ./
 RUN npm run build
 
--- a/cdk-infra/lib/constructs/worker/worker-construct.ts
+++ b/cdk-infra/lib/constructs/worker/worker-construct.ts
@@ -13,6 +13,7 @@ import { IQueue } from 'aws-cdk-lib/aws-sqs';
 import { Construct } from 'constructs';
 import path from 'path';
 import { isEnhanced } from '../../../utils/env';
+import { Duration } from 'aws-cdk-lib';
 
 interface WorkerConstructProps {
   environment: Record<string, string>;
@@ -92,6 +93,7 @@ export class WorkerConstruct extends Construct {
     taskDefinition.addContainer('workerImage', {
       image: ContainerImage.fromAsset(path.join(__dirname, '../../../../'), containerProps),
       environment,
+      stopTimeout: Duration.minutes(5),
       logging: LogDrivers.awsLogs({
         streamPrefix: 'autobuilderworker',
         logGroup: taskDefLogGroup,
--- a/src/enhanced/enhancedApp.ts
+++ b/src/enhanced/enhancedApp.ts
@@ -1,14 +1,31 @@
 import { handleJob } from './utils/job';
 import { listenToJobQueue } from './utils/queue';
+import mongodb from 'mongodb';
+import c from 'config';
+
+let client: mongodb.MongoClient;
 
 async function app() {
   console.log('starting application');
-  const { jobId } = await listenToJobQueue();
 
-  await handleJob(jobId);
+  try {
+    const { jobId } = await listenToJobQueue();
 
-  console.log('process completed');
+    const atlasURL = `mongodb+srv://${c.get('dbUsername')}:${c.get('dbPassword')}@${c.get(
+      'dbHost'
+    )}/?retryWrites=true&w=majority`;
 
+    client = new mongodb.MongoClient(atlasURL);
+    await client.connect();
+    const db = client.db(c.get('dbName'));
+
+    await handleJob(jobId, db);
+    console.log('process completed');
+  } catch (e) {
+    console.error('ERROR! Job failed', e);
+  }
+
+  client.close();
   process.exit(0);
 }
 
--- a/cdk-infra/lib/constructs/api/util/custom-node-function.ts
+++ b/cdk-infra/lib/constructs/api/util/custom-node-function.ts
@@ -1,0 +1,7 @@
+import { Construct } from 'constructs';
+
+export class FastNodejsFunction extends Construct {
+  constructor(scope: Construct, id: string) {
+    super(scope, id);
+  }
+}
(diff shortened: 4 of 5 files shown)


cdk-infra/package.json
```diff
@@ -8,14 +8,21 @@
...
     "test": "jest",
-    "cdk": "cdk"
+    "cdk": "cdk",
+    "deploy:enhanced": "rm -rf cdk.out && cdk deploy -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --all",
+    "deploy:enhanced:worker": "rm -rf cdk.out && cdk deploy auto-builder-stack-enhancedApp-worker -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never",
+    "deploy:enhanced:webhooks": "rm -rf cdk.out && cdk deploy auto-builder-stack-enhancedApp-webhooks -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never"
+
   },
   "devDependencies": {
+    "@swc/core": "^1.3.66",
```
branberry, line 18: Speeds up bundling process using this dependency

src/enhanced/utils/job/handle-job.ts
```diff
@@ -0,0 +1,67 @@
+import { JobManager, JobHandlerFactory } from '../../../job/jobManager';
```
branberry, line 1: Nothing new here, just moved handle job code from index to a separate file as there are now other functions that this directory exports.

cdk-infra/lib/constructs/worker/worker-construct.ts
```diff
@@ -89,9 +89,22 @@ export class WorkerConstruct extends Construct {
       executionRole,
     });
 
+    const updateTaskProtectionPolicy = new PolicyStatement({
```
branberry, line 92: grants the task the ability to protect itself from being stopped during a deployment or scale-in

cdk-infra/lib/constructs/worker/worker-construct.ts
```diff
@@ -103,6 +116,7 @@ export class WorkerConstruct extends Construct {
       taskDefinition,
       desiredCount: 5,
       minHealthyPercent: 100,
+      maxHealthyPercent: 200,
```
branberry, line 119: Allows up to 10 tasks per service. This allows for existing jobs to be processed in the event an update occurs.

Dockerfile.enhanced
```diff
@@ -4,7 +4,7 @@ WORKDIR /home/docsworker-xlarge
 COPY  config config/
 COPY package*.json ./
 COPY tsconfig*.json ./
-RUN npm ci
+RUN npm ci --legacy-peer-deps
```
rayangler, line 7: Is there a new dependency that warrants the `legacy-peer-deps` flag?

cdk-infra/lib/stacks/auto-builder-stack.ts
```diff
@@ -21,6 +22,7 @@ export class AutoBuilderStack extends Stack {
     super(scope, id, props);
 
     const queues = new AutoBuilderQueuesConstruct(this, 'queues');
+    const { vpc } = new AutoBuilderVpcStack(this, 'enhancedVpc');
```
rayangler, line 25: What is the difference between this VPC stack and the one made in `cdk-infra/bin/cdk-infra.ts`? Why do we need VPC stacks with 2 different IDs?

src/enhanced/enhancedApp.ts
```diff
@@ -1,15 +1,67 @@
 import { handleJob } from './utils/job';
 import { listenToJobQueue } from './utils/queue';
+import mongodb, { MongoClient } from 'mongodb';
+import c from 'config';
```
rayangler, line 4: (Super nit) Might be good to import these above our relative imports

rayangler (commented):
I think things look good overall! Mostly just a couple questions for my own understanding

Dockerfile.enhanced
branberry: I was running into build errors without this flag. I can test again to see if it works without needing it.

cdk-infra/lib/stacks/auto-builder-stack.ts
branberry: Thanks for pointing this out. I sort of added this without taking into consideration how it's actually being deployed. This whole stack can be deleted.

src/enhanced/enhancedApp.ts
```diff
@@ -1,15 +1,67 @@
...
+  client = new MongoClient(atlasURL);
+  console.log('[connectToDb]: Connecting to client');
+  await client.connect();
+  return client.db(c.get('dbName'));
+}
+
+async function cleanupJob(): Promise<never> {
+  try {
+    console.log('[cleanupJob]: Closing MongoDB client connection...');
+    await client?.close();
```
mmeigs, line 23: nit: Maybe check that `client` is defined earlier and return if not, so that we never get a log saying we're closing the connection if there isn't one. I also understand this will probably never happen and is only there for typescript reasons.

src/enhanced/utils/queue/index.ts
```diff
@@ -13,7 +14,7 @@ export async function listenToJobQueue(): Promise<JobsQueuePayload> {
 
   const client = new SQS({ region });
 
-  console.log('Polling jobsQueue');
+  console.log('[listenToJobQueue]: Polling jobsQueue');
 
   // We want to loop indefinitely so that we continue to poll the queue.
   while (true) {
```
mmeigs, line 20: Is there a limit on this loop? It seems that each possible error continues to iterate. Do we want a timebox or some other flag to end the process? I might have missed this somewhere.

mmeigs (commented):
Looks clean, love the clear naming. Couple of tiny comments.

src/enhanced/utils/queue/index.ts
branberry: If an error occurs due to an improperly formatted message, we continue looping because that improperly formatted message will have been deleted from the queue. This means that the queue will no longer have the problematic message, and the task can continue polling for a valid job. 

Basically, if we get a bad message, we toss it out and wait until we get a good message. These errors _should_ be extremely rare, but these extra measures were thrown in just in case something went wrong somewhere.

mmeigs: So, we have confidence that a good message will always be sent? Is there a time box we can set even if large so that we don't leave it running forever? Or is this not an issue and I'm not seeing the forest for the trees :)

Dockerfile.enhanced
rayangler: Gotcha. If this is necessary, I'm fine with leaving in 👍

rayangler (approved):
LGTM!

PR #845, opened 2023-06-27
DOP-3792: added featureFlagSearchUI

mmeigs:
Adds feature flag for Search UI Improvements project.

It is piped through to the frontend. Most likely will be utilized in Snooty, but entire implementation still undecided.

AWS Parameter Store values can be found [here](https://us-east-2.console.aws.amazon.com/systems-manager/parameters/?region=us-east-2&tab=Table#list_parameter_filters=Name:Contains:search_ui). For now, all are false. Staging will be changed to true at some point in near future.


src/job/jobHandler.ts
```diff
@@ -342,6 +342,7 @@ export abstract class JobHandler {
     const snootyFrontEndVars = {
       GATSBY_BASE_URL: this._config.get<string>('gatsbyBaseUrl'),
       PREVIEW_BUILD_ENABLED: this._config.get<string>('previewBuildEnabled'),
+      FEATURE_FLAG_SEARCH_UI: this._config.get<string>('featureFlagSearchUI'),
```
seungpark, line 345: minor nitpick. would prefix with `GATSBY_` but not a deal breaker

seungpark (approved):
this LGTM besides a minor naming nitpick!

src/job/jobHandler.ts
mmeigs: ahhh, I see this in the documentation now. I think that's reasonable. How do we feel about this (FE) being `GATSBY_TEST_SEARCH_UI` while the rest of the pipeline of the variable (BE) stays `FEATURE_FLAG_SEARCH_UI`? Should I change all of the naming to match?

mmeigs: Oh, sure! https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/

seungpark (approved):
this LGTM 👍

PR #846, opened 2023-07-05
[DOP-3798]: Add TTL indexes to updated documents

branberry:
### Ticket
[DOP-3798](https://jira.mongodb.org/secure/RapidBoard.jspa?rapidView=953&projectKey=DOP&view=detail&selectedIssue=DOP-3798)

### DIFF
--- a/modules/persistence/src/services/pages/index.ts
+++ b/modules/persistence/src/services/pages/index.ts
@@ -2,7 +2,7 @@ import AdmZip from 'adm-zip';
 import { deserialize } from 'bson';
 import isEqual from 'fast-deep-equal';
 import { AnyBulkWriteOperation, Document, FindCursor, ObjectId } from 'mongodb';
-import { bulkWrite, db, insert } from '../connector';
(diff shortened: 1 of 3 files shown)


modules/persistence/src/services/pages/index.ts
```diff
@@ -144,6 +150,7 @@ class UpdatedPagesManager {
                 static_assets: this.findUpdatedAssets(page.static_assets, prevPageData?.static_assets),
                 updated_at: this.updateTime,
                 deleted: false,
+                ttl,
```
seungpark, line 153: this is adding a `ttl` field to the document.

i think the ticket calls for adding a `ttl index` to the whole collection. we can do this as a adhoc, or come up with a a manager of db and collections. i think the former would be faster in getting the needful done, but the latter would be a centralized place for db management

seungpark (changes requested):
see comment below! i think we want a index on collection vs ttl field on document

PR #847, opened 2023-07-05
[DOP-3833]: Add new workflows

branberry:


.github/workflows/deploy-stg-enhanced-webhooks.yml
```diff
@@ -0,0 +1,32 @@
+on:
+  push:
+    paths: ["api/v2/**"]
+    branches:
+      - "master"
+      - "integration"
+      - 'DOP-3833'
```
branberry, line 7: Will remove this before merging

.github/workflows/deploy-stg-enhanced-worker.yml
```diff
@@ -0,0 +1,32 @@
+on:
+  push:
+    paths: ["src/**"]
+    branches:
+      - "master"
+      - "integration"
+      - 'DOP-3833'
```
branberry, line 7: Will remove this before merging

.github/workflows/deploy-stg-enhanced-worker.yml
```diff
@@ -0,0 +1,32 @@
...
+
+concurrency:
+  group: environment-stg-${{ github.ref }}
```
rayangler, line 10: Would having the same group name here conflict with the workflow for the webhook deployment (basing this question off of this [docs example](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow))? For example, if a new PR changes files in paths for both the worker and the API, I'm wondering if this might unintentionally cause one of the workflows to get cancelled

.github/workflows/deploy-stg-enhanced-worker.yml
```diff
@@ -0,0 +1,32 @@
...
+  cancel-in-progress: true
+name: Deploy Staging ECS
```
rayangler, line 12: It might also be worth giving these workflows different names to help distinguish them when filtering between different GH Actions in the UI. Notice the 2 labels for "Deploy Staging ECS" on [this page](https://github.com/mongodb/docs-worker-pool/actions).

rayangler (approved):

PR #848, opened 2023-07-10
DOP-3591

seungpark:
DOP-3591

**this is a POC** of invoking mut indexing as part of a job (vs [currently we are enqueueing another job ](https://github.com/mongodb/docs-worker-pool/blob/master/src/job/productionJobHandler.ts#L130C1-L130C1))

**[build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64ac64de85ac5a4d8f9a7d4a)**

relevant changes:
- production jobs do not enqueue another job for search
- `mut-index` is piped into `job.deployCommands` if this [job should include a search manifest generation](https://github.com/mongodb/docs-worker-pool/blob/master/src/job/jobHandler.ts#L513)
- calling this newly added deploy job with the [callWithBenchmark](https://github.com/mongodb/docs-worker-pool/blob/master/src/job/jobHandler.ts#L250) function in order to log search job times

**note:**
- builds will wait for search jobs to finish (as part of deploy steps) before updating the job with completion status.
- jobs are stamped with `searchStartTime` and `searchEndTime` to track job length (the displayed build log above are only showing front end properties)
image


.github/workflows/deploy-stg-ecs.yml
```diff
@@ -4,6 +4,7 @@ on:
       - "master"
       - "integration"
       - "close-lambda-mdb-clients"
+      - "DOP-3591"
```
casthewiz, line 7:
```suggestion
```

tests/unit/job/productionJobHandler.test.ts
```diff
@@ -173,7 +173,7 @@ describe('ProductionJobHandler Tests', () => {
     );
   });
 
-  test('Default production deploy kicks off manifest generation', async () => {
+  test('Default production deploy does not kick off manifest generation', async () => {
```
casthewiz, line 176: We should change the wording here - by default, we want production deploys (ones enqueued via the slack job) to generate a manifest, we just don't want it to do so in another job 😉 

```suggestion
  test('Default production deploy does not kick off additional manifest generation job', async () => {
```

casthewiz (approved):
IIRC the `prepSearchDeploy` is using logic from the controller - we should leave a comment to remove that code once this is merged

seungpark:
> IIRC the `prepSearchDeploy` is using logic from the controller - we should leave a comment to remove that code once this is merged

done. i left a comment in the manifestJobHandler

also verified this does not affect writers' times, as there are no search jobs done on writer builds

tests/unit/job/productionJobHandler.test.ts
seungpark: done

.github/workflows/deploy-stg-ecs.yml
seungpark: done

PR #850, opened 2023-07-10
DOP-3620

caesarbell:
This PR focuses on adding the preview webhook POST request to the `deploy()` method in the `stagingJobHandler`. This is wrapped in a new feature flag called `GATSBY_CLOUD_PREVIEW_WEBHOOK_ENABLED` which is enabled only for the following environment

- stg


src/job/jobHandler.ts
```diff
@@ -614,6 +615,24 @@ export abstract class JobHandler {
   public getLogger(): ILogger {
     return this._logger;
   }
+
+  protected async previewWebhook(): Promise<AxiosResponse> {
```
branberry, line 619: NIT: I don't think the `AxiosResponse` is the correct type. I believe this corresponds to the entire response we get from `axios.post`, and not the `data` property. This is the interface definition for `AxiosResponse`:

```ts
export interface AxiosResponse<T = any>  {
  data: T;
  status: number;
  statusText: string;
  headers: any;
  config: AxiosRequestConfig;
  request?: any;
}
```

If we know what the response data should be, we can define our own interface, and provide it as a type argument in our `axios.post` call. Here's an example of what that could look like:

```ts
interface GatsbyCloudBuildResponse {
  success: boolean;
  staging_url: string;
}


protected async previewWebhook(): Promise<GatsbyCloudBuildResponse> {
  const previewWebhookURL = 'https://webhook.gatsbyjs.com/hooks/data_source';
  const gatsbySiteDataSource = process.env.GATSBY_CLOUD_PREVIEW_WEBHOOK;
  const url = `${previewWebhookURL}/${gatsbySiteDataSource}`;

  // the type for data will now be GatsbyCloudBuildResponse
  const { data } = await axios.post<GatsbyCloudBuildResponse>(
    url,
    {
      project: this.currJob.payload.project,
      branch: this.currJob.payload.branchName,
    },
    {
      headers: { 'x-gatsby-cloud-data-source': 'gatsby-source-snooty' },
    }
  );

    return data;
  }
```

branberry (commented):
Hey Caesar! This looks good to me. I just have one nit comment for ya.

schmalliso:
We should not enable it for `prd` at this time, since that's the environment the writers work in and we're not ready for them to be hidding Gatsby Cloud 😁. I shall review the rest nowish, but just wanted to not forget that.

src/job/stagingJobHandler.ts
```diff
@@ -72,6 +72,14 @@ export class StagingJobHandler extends JobHandler {
       if (resp?.output?.includes('Summary')) {
         resp.output = resp.output.slice(resp.output.indexOf('Summary'));
       }
+      // Invoke Gatsby Preview Webhook
+      const featurePreviewWebhookEnabled = process.env.GATSBY_CLOUD_PREVIEW_WEBHOOK_ENABLED;
+      if (featurePreviewWebhookEnabled) {
+        // TODO: We may want to send a response to slack or something to let us know
+        // the outcome of the call
+        this.previewWebhook();
```
schmalliso, line 80: Agree with your TODO that we should probably be logging / slacking / something-ing here to capture the response from the post to the webhook. Do you want to just add some rudimentary logging for now?

branberry (approved):
LGTM!

PR #852, opened 2023-07-12
DOP-3769: Create webhook to handle post-build operations from Gatsby Cloud builds

rayangler:
### Ticket

DOP-3769

### Notes

* This webhook is expected to be called by Snooty's new source plugin for Gatsby Cloud directly after a build successfully completes. To help validate the expectation that Snooty is the caller of this webhook, we include a header for a signature to check against a special secret (similar to GitHub and Slack). This secret is currently in Parameter Store, and will be added to the env setups for Gatsby Cloud sites when it comes time to include this.
* It's worth noting that this current implementation assumes that the source plugin will be calling the webhook. If things change and Gatsby Cloud allows us to call this webhook through its built-in outgoing notifications system with the data we need to pass, then this implementation can be handled accordingly.


api/controllers/v1/jobs.ts
```diff
@@ -187,6 +204,11 @@ async function NotifyBuildSummary(jobId: string): Promise<any> {
       const prCommentId = await githubCommenter.getPullRequestCommentId(fullDocument.payload, pr);
       const fullJobDashboardUrl = c.get<string>('dashboardUrl') + jobId;
 
+      // We currently avoid posting the Gatsby Cloud preview url on GitHub to avoid
+      // potentially conflicting behavior with the S3 staging link with parallel
+      // frontend builds. This is in case the GC build finishing first causes the
+      // initial comment to be made with a nullish S3 url, while subsequent comment
+      // updates only append the list of build logs.
```
rayangler, line 211: This is something we should probably handle separately to test its behavior, and to avoid breaking the feature as part of this PR

src/repositories/jobRepository.ts
```diff
@@ -21,8 +21,14 @@ export class JobRepository extends BaseRepository {
...
-  async updateWithCompletionStatus(id: string, result: any): Promise<boolean> {
-    const query = { _id: id };
+  async updateWithCompletionStatus(
+    id: string | mongodb.ObjectId,
+    result: any,
+    shouldNotifySqs = true
+  ): Promise<boolean> {
+    // Safely convert to object ID
+    const objectId = new mongodb.ObjectId(id);
+    const query = { _id: objectId };
```
rayangler, line 31: For some reason, this expected a string, but queries should typically expect an ObjectId. This change should help enforce valid IDs to be used while ensuring jobs can be queried successfully regardless of whether the ID is in a string or ObjectId format

src/repositories/jobRepository.ts
```diff
@@ -35,7 +41,7 @@ export class JobRepository extends BaseRepository {
       update,
       `Mongo Timeout Error: Timed out while updating success status for jobId: ${id}`
     );
-    if (bRet) {
+    if (bRet && shouldNotifySqs) {
```
rayangler, line 44: Added to prevent needing to send out a new SQS message when we can handle calling any post-build functions directly in the new webhook implementation with custom fields. Feel free to let me know if there's a preference to rely on an SQS message here!

branberry, line 44: NIT: I believe line 45 (the `this.notify` call) now has a type error since the type of `id` is now `string | mongodb.ObjectID`. 

One approach you could take is to use the `objectId` variable that you defined on line 30, and call its `toString` method to get the string ID. It seems a bit wonky, but this would guarantee that the value will be a string and not potentially the `mongodb.ObjectID` type. 

ex.

```ts
      await this.notify(objectId.toString(), c.get('jobUpdatesQueueUrl'), JobStatus.completed, 0);
```

api/controllers/v1/jobs.ts
```diff
@@ -385,3 +412,100 @@ async function SubmitArchiveJob(jobId: string) {
...
+    const errMsg = 'Payload signature is incorrect';
+    consoleLogger.error('SnootyBuildCompleteError', errMsg);
+    return {
+      statusCode: 401,
+      headers: { 'Content-Type': 'text/plain' },
+      body: errMsg,
+    };
+  }
+
+  const payload: SnootyPayload = JSON.parse(event.body);
```
branberry, line 471: NIT: `JSON.parse` can potentially throw an error, so maybe to be safe we can wrap it in a `try/catch`

api/controllers/v1/jobs.ts
```diff
@@ -385,3 +412,100 @@ async function SubmitArchiveJob(jobId: string) {
...
+  const { jobId } = payload;
+
+  if (!jobId) {
+    const errMsg = 'Payload missing job ID';
+    consoleLogger.error('SnootyBuildCompleteError', errMsg);
+    return {
+      statusCode: 401,
```
branberry, line 478: NIT: I believe the status code should be a `400` if the payload is missing the `jobId` property

api/controllers/v1/jobs.ts
```diff
@@ -385,3 +412,100 @@ async function SubmitArchiveJob(jobId: string) {
...
+    const jobRepository = new JobRepository(db, c, consoleLogger);
+    await jobRepository.updateWithCompletionStatus(jobId, null, false);
+    // Placeholder preview URL until we iron out the Gatsby Cloud site URLs.
+    // This would probably involve fetching the URLs in the db on a per project basis
+    const previewUrl = 'https://www.mongodb.com/docs/';
+    await NotifyBuildSummary(jobId, { mongoClient: client, previewUrl });
+  } catch (e) {
+    consoleLogger.error('SnootyBuildCompleteError', e);
+    return {
+      statusCode: 400,
```
branberry, line 498: NIT: I believe `500` would be an appropriate status code here. 4XX status codes generally refer to client errors, but I believe if an error was thrown here, then it would be considered an internal server error.

rayangler: Thanks for catching these status code issues! I also ended up updating the final response code to 200, just because this function's operations are definitely more of a commitment than 202 would suggest

PR #854, opened 2023-07-13
Update `stagingJobHandler.ts`

caesarbell:
This is a quick patch to see the boolean value appear in the log. It will be removed once we verify the build in Gatsby Cloud.


branberry (approved):
LGTM!

PR #858, opened 2023-07-17
[DOP-3856]: Point enhanced app to existing buckets by default

branberry:
## Notes

This PR points the Enhanced Autobuilder to the `prd` environment. It also properly configures the Enhanced Autobuilder to upload to the S3 buckets for the `mut-publish` command. Also, this configures the slackbot to work with the enhanced infrastructure.

## Test
To test, simply use any docs site fork that you currently use for testing, and change the webhook URL to the following:
https://5ebq1gien6.execute-api.us-east-2.amazonaws.com/prod/webhook/githubEndpoint/trigger/build
To trigger the webhook, either make a change to the master branch, or make a branch on the fork. This should trigger the Enhanced Autobuilder.


cdk-infra/utils/buckets.ts
```diff
@@ -10,3 +16,149 @@ export const docsBucketNames = [
...
   'docs-node',
 ] as const;
+
+interface CustomBucketProps {
+  scope: Construct;
+  featureName: string;
+  env: AutoBuilderEnv;
+  bucketName: string;
+}
+export function createCustomBucket({ scope, featureName, env, bucketName }: CustomBucketProps): Bucket {
```
branberry, line 26: This logic existed previously in the `bucket-construct.ts` file, but moved it here since it isn't always needed. It's also nice to separate this out into its own function in a separate file.

api/controllers/v2/jobs.ts
```diff
@@ -58,14 +58,13 @@ export const HandleJobs = async (event: SQSEvent): Promise<void> => {
...
     messages.map(async (message: SQSRecord) => {
       const consoleLogger = new ConsoleLogger();
+      console.log('SQS RECORD: ', message);
       const body = JSON.parse(message.body);
       const jobId = body['jobId'];
       const jobStatus = body['jobStatus'];
+
       try {
         switch (jobStatus) {
-          case JobStatus[JobStatus.inQueue]:
```
branberry, line 66: Since there really isn't an in queue period, I decided to remove this. Basically, you get sent the identical message twice in a row which seems jarring when they come one right after another.

cdk-infra/lib/constructs/api/webhook-api-construct.ts
```diff
@@ -89,6 +89,7 @@ export class WebhookApiConstruct extends Construct {
       handler: 'HandleJobs',
       environment,
       bundling,
+      timeout,
```
branberry, line 92: Very small, but super important!! The default timeout is only 3 seconds. If the lambda times out before it finishes, SQS will treat that as a failed message read and will attempt to invoke the message again for the lambda to complete properly. If the lambda times out, the SQS queue does not delete the message which means it keeps trying forever.

cdk-infra/lib/constructs/api/webhook-env-construct.ts
```diff
@@ -15,24 +17,31 @@ export class WebhookEnvConstruct extends Construct {
...
+    const entitlementCollection = StringParameter.valueFromLookup(
+      this,
+      `${ssmPrefix}/atlas/collections/user/entitlements`
+    );
     const dbPassword = secureStrings['MONGO_ATLAS_PASSWORD'];
     this.environment = {
       ...secureStrings,
       MONGO_ATLAS_USERNAME: dbUsername,
       MONGO_ATLAS_PASSWORD: dbPassword,
+      MONGO_ATLAS_HOST: dbHost,
```
schmalliso, line 35: What's this field?

cdk-infra/static/custom-environment-variables.json
```diff
@@ -21,6 +21,8 @@
   "fastlyCloudManagerServiceId": "FASTLY_CLOUD_MANAGER_SERVICE_ID",
   "gatsbyBaseUrl": "GATSBY_BASE_URL",
   "previewBuildEnabled": "PREVIEW_BUILD_ENABLED",
+  "featureFlagUpdatePages": "FEATURE_FLAG_UPDATE_PAGES",
```
schmalliso, line 24: For my own edification -- why do we have this file and also the one at cdk-infa/static/config/custom-environment-variables.json ?

cdk-infra/utils/buckets.ts
```diff
@@ -10,3 +16,149 @@ export const docsBucketNames = [
...
+  const stackBucketName = `${featureName}-${bucketName}-${env}`.toLowerCase();
+  const websiteRoutingRules = bucketName === 'docs-mongodb-org' ? getMongoDocsBucketRoutingRules(env) : undefined;
```
schmalliso, line 28: It is very possible we'll have custom routing rules (e.g. bucket-level redirects) for other buckets at some point -- will we need to change this when that happens?

cdk-infra/utils/buckets.ts
```diff
@@ -10,3 +16,149 @@ export const docsBucketNames = [
...
+
+  const bucket = new Bucket(scope, stackBucketName, {
+    removalPolicy: RemovalPolicy.DESTROY,
+    websiteRoutingRules,
+    bucketName: stackBucketName,
+    websiteIndexDocument: 'index.html',
+    websiteErrorDocument: 'docs-qa/404/index.html',
```
schmalliso, line 35: Should this be `docs` rather than `docs-qa` ?

cdk-infra/utils/buckets.ts
```diff
@@ -10,3 +16,149 @@ export const docsBucketNames = [
...
+      blockPublicPolicy: false,
+      ignorePublicAcls: false,
+      restrictPublicBuckets: false,
+    }),
+  });
+
+  return bucket;
+}
+
+export function getMongoDocsBucketRoutingRules(env: AutoBuilderEnv): RoutingRule[] {
```
schmalliso, line 47: I realize this moved from the other file -- I'm just slightly confused about how our buckets end up actually working since the bucket that has all of these defined routing rules is docs-mongodb-org-dotcomprd and I don't see where we append the `-dotcomprd` to the end of the bucket… (To be fair I don't get how that worked in the previous code approach either…)

schmalliso (commented):
A cursory first pass with some questions for my edification.

cdk-infra/static/custom-environment-variables.json
branberry: Good question, we don't actually need this one I don't believe now that I think about it. We do need the `cd-infra/static/api/config`, though. This is because of how the Lambdas are bundled vs using Docker containers for the workers. When bundling the lambdas, the bundler can't access files above the root directory of the `cdk-infra` repo i.e. we can't get the these config JSON files from the root directory of `docs-worker-pool` so we have to copy them over.

 I'll remove it and test it to confirm that it still works.

cdk-infra/utils/buckets.ts
branberry: Good point. I was working in `dotcomstg` and hardcoded this in. I actually need to create a mapping of environment to `websiteErrorDocument` like [here](https://github.com/mongodb/docs-worker-pool/blob/cd75999ac346b96bb8d96a74125fa83d9b85eb90/infrastructure/ecs-main/serverless.yml#L197).

cdk-infra/utils/buckets.ts
branberry: I believe when checking the bucket configurations that all `docs-mongodb-org` buckets have routing rules defined for them. Essentially, the code here is copy-pasted from [here](https://github.com/mongodb/docs-worker-pool/blob/cd75999ac346b96bb8d96a74125fa83d9b85eb90/infrastructure/ecs-main/buckets.yml#L14).

This method doesn't create the bucket, but simply creates the routing rules object that we want to provide for the `docs-mongodb-org` bucket. We append the `env` name for the bucket [here](https://github.com/mongodb/docs-worker-pool/blob/cd75999ac346b96bb8d96a74125fa83d9b85eb90/cdk-infra/lib/constructs/worker/buckets-construct.ts#L24). 

For context, the `getEnv` method is tied to the Snooty/Autobuilder env values such as `dotcomstg`, `dotcomprd`, `prd`, etc.

cdk-infra/utils/buckets.ts
branberry: Yep! I think we can refactor how we get the redirect rules in the future to make them specific to a bucket, but I don't believe that should be too complicated. I can brainstorm ways in which we can handle different websiteRoutingRules for a given bucket. Maybe I could use a map of sorts.

schmalliso: Sweet, this doesn't need to be blocking, I just wanted to double check feasibility.

cdk-infra/utils/buckets.ts
schmalliso: AAAAH. Okay thanks. I was confuzzled! Perfect.

cdk-infra/package.json
```diff
@@ -9,12 +9,13 @@
     "watch": "tsc -w",
     "test": "jest",
     "cdk": "cdk",
-    "deploy:enhanced": "cdk deploy -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --all",
-    "deploy:enhanced:worker": "cdk deploy auto-builder-stack-enhancedApp-worker -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --exclusively",
-    "deploy:enhanced:webhooks": "cdk deploy auto-builder-stack-enhancedApp-webhooks -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --exclusively",
-    "deploy:feature": "cdk deploy -c enhanced=true -c env=dotcomstg --require-approval never --all",
-    "deploy:feature:worker": "cdk deploy  -c enhanced=true -c env=dotcomstg --require-approval never --exclusively",
-    "deploy:feature:webhooks": "cdk deploy -c enhanced=true -c env=dotcomstg --require-approval never --exclusively"
+    "deploy:enhanced": "cdk deploy -c enhanced=true -c customFeatureName=enhancedApp -c env=prd --require-approval never --all",
```
rayangler, line 12: Are these commands expected to be tied to the `prd` environment by default?

cdk-infra/lib/constructs/worker/buckets-construct.ts
```diff
@@ -1,159 +1,29 @@
...
-        websiteIndexDocument: 'index.html',
-        websiteErrorDocument: 'docs-qa/404/index.html',
-        blockPublicAccess: new BlockPublicAccess({
-          blockPublicAcls: false,
-          blockPublicPolicy: false,
-          ignorePublicAcls: false,
-          restrictPublicBuckets: false,
-        }),
-      });
+      const bucket = Bucket.fromBucketName(this, stackBucketName, `${bucketName}-${bucketEnv}`);
```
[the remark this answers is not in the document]
branberry: The `fromBucketName` method requires a unique ID value to identify the construct, which is the second argument of the method. I believe this is just necessary for CDK to uniquely identify constructs that are instantiated 🤪. I arbitrarily chose `stackBucketName` to be a combination of `${featureName}-${bucketName}-${bucketEnv}`, but it's also fine if I just did the bucket-name with the env. 

We exclude the `featureName` for the final argument because that is the name of the bucket we want to retrieve.

There won't be a conflict if 2 different features referred to the same bucket as we are retrieving that bucket so that we can grant permission for the enhanced autobuilder to read and write to the bucket. We actually do this now, these buckets are currently used by the standard autobuilder and the enhanced currently.

cdk-infra/package.json
branberry: yep! Would you like me to specify the environment in the command name? I could do something like `"deploy:enhanced:prd"` or something along those lines.

cdk-infra/lib/constructs/worker/buckets-construct.ts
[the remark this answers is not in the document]
rayangler: Oh, gotcha! For some reason I thought we were creating a new set of isolated `${featureName}-${bucketName}-${bucketEnv}` buckets per feature, but reusing the buckets makes sense here. Thanks for clearing that up for me!

cdk-infra/package.json
rayangler: Hm, `deploy:enhanced:prd` might work. I was just concerned about the GitHub Actions for staging calling these commands for the `prd` environment

cdk-infra/lib/constructs/worker/buckets-construct.ts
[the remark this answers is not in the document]
branberry: Np! The naming I will admit is confusing here. I could do something like `bucketConstructId` to make it more clear.

cdk-infra/package.json
```diff
@@ -9,12 +9,13 @@
     "watch": "tsc -w",
     "test": "jest",
     "cdk": "cdk",
-    "deploy:enhanced": "cdk deploy -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --all",
-    "deploy:enhanced:worker": "cdk deploy auto-builder-stack-enhancedApp-worker -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --exclusively",
-    "deploy:enhanced:webhooks": "cdk deploy auto-builder-stack-enhancedApp-webhooks -c enhanced=true -c customFeatureName=enhancedApp -c env=dotcomstg --require-approval never --exclusively",
-    "deploy:feature": "cdk deploy -c enhanced=true -c env=dotcomstg --require-approval never --all",
-    "deploy:feature:worker": "cdk deploy  -c enhanced=true -c env=dotcomstg --require-approval never --exclusively",
-    "deploy:feature:webhooks": "cdk deploy -c enhanced=true -c env=dotcomstg --require-approval never --exclusively"
+    "deploy:enhanced:prd": "cdk deploy -c enhanced=true -c customFeatureName=enhancedApp -c env=prd --require-approval never --all",
```
rayangler, line 12: Sorry, I should have considered this in the other thread, but would changing the names of the commands here impact the GH actions command calls? Do we need to have separate ones per env? Or is there something we could do like `npm run deploy:enhanced -- -c env=[prd|stg|dotcomstg|etc]`?

branberry: Hi Raymund! We could absolutely do something like you mentioned here, and that would probably be more straightforward as there will be fewer commands in the `package.json`

rayangler (approved):
LGTM!

PR #859, opened 2023-07-20
DOP-3818: Create webhook to mark build data for deletion

rayangler:
### Ticket

DOP-3818

### Testing

Assuming that this branch is in pre-production, you can do the following to test:
1. Ensure you have the `stg` GitHub build trigger webhook set up in a fork of a docs repo.
2. Create and push a new branch to **your fork**. Content changes are optional. This should trigger an Autobuilder job.
3. Follow the steps in this [wiki page](https://wiki.corp.mongodb.com/display/DE/Set+Up+a+GitHub+Repo+to+Mark+Build+Data+for+Deletion) to add the new deletion webhook in **your fork** of the docs repo. Since this is a test, this can be done on a fork and not an actual upstream (mongodb/10gen) repo.
4. Create a PR for the new branch. Make sure the base branch is **your fork** where you have the new deletion webhook added.
5. Once the Autobuilder job is complete, close or merge the new branch. This will trigger the new deletion webhook to mark data for deletion.
6. Ensure that the metadata and updated page ASTs associated with the project + branch you created have a `deleted` field set to true.

**Screenshots:**
Screen Shot 2023-07-21 at 1 17 02 PM
Screen Shot 2023-07-21 at 1 16 20 PM

### Notes

* Adds a new webhook to mark build data for deletion. The hope is to have the Gatsby source plugin delete its internal nodes pertaining to data where the `deleted` field is set to true. This will hopefully help reduce the memory usage for Gatsby Cloud sites in a more immediate term. The data in the collection will actually be deleted by the existing TTLs when the time comes.
* The new webhook/function has only be added in v1 right now. Some of the env setup has already been done in the CDK code just to have the values there, but adding the webhook/function to the v2 paths in a future separate PR might be more suitable to help keep the scope of changes and testing down.


src/repositories/metadataRepository.ts
```diff
@@ -0,0 +1,41 @@
...
+import { IConfig } from 'config';
+import { BaseRepository } from './baseRepository';
+import { ILogger } from '../services/logger';
+
+/**
+ * Manages metadata for parser builds.
+ */
+export class MetadataRepository extends BaseRepository {
+  constructor(db: Db, config: IConfig, logger: ILogger) {
+    super(config, logger, 'MetadataRepository', db.collection('metadata'));
```
rayangler, line 11: The collection name is sometimes passed down as an env variable passed down from param store for other existing repositories. But I'm not sure what its benefit is when the collection name is expected to be static for every environment.

I figured keeping it hardcoded here would reduce unnecessary envs and the name itself isn't sensitive, but let me know if the preference is to make env variables

tests/unit/api/github.test.ts
```diff
@@ -0,0 +1,104 @@
...
+      },
+      repository: {
+        name: 'docs-landing',
+      },
+    };
+
+    beforeAll(() => {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      jest.spyOn(mongodb, 'MongoClient').mockReturnValue({
```
rayangler, line 51: I think ideally, we use `MongoMemoryServer` to have an actual local db to test against, but I think this can be done as part of a future isolated change. It would be beneficial for all of the other tests that currently mock some sort of repository

api/controllers/v1/github.ts
```diff
@@ -109,3 +114,101 @@ export const TriggerBuild = async (event: any = {}, context: any = {}): Promise<
     body: 'Job Queued',
   };
 };
+
+/**
+ * Deletes build artifacts for a given project + branch combination.
+ * @param event
+ */
+export const MarkBuildArtifactsForDeletion = async (event: APIGatewayEvent) => {
```
branberry, line 122: This isn't specific to your PR, but I am wondering how we should handle changes that will exist in both the regular autobuilder and the enhanced autobuilder moving forward. The enhanced autobuilder uses the `api/controllers/v2` directory. Should we maybe create a shared directory as well?

api/controllers/v1/github.ts
```diff
@@ -109,3 +114,101 @@ export const TriggerBuild = async (event: any = {}, context: any = {}): Promise<
...
+    return {
+      statusCode: 400,
+      headers: defaultHeaders,
+      body: err,
+    };
+  }
+
+  let payload: PullRequestEvent | undefined;
+  try {
+    payload = JSON.parse(event.body) as PullRequestEvent;
```
branberry, line 157: NIT: You're already doing runtime validation below to check and see if the action is `closed`, but I think it could be helpful to check and see if the request is actually for a PR. I believe you can have the `action` equal `closed` for an issue as an example. [Looking at the documentation](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request), it appears that you can check to see if the `pull_request` property exists on the payload.

src/repositories/metadataRepository.ts
branberry: I think this is totally fine to use over environment variables. I think it would also be worthwhile to assign the collection name to a `const` variable if that collection name is used elsewhere, just to prevent the possibility of a typo in one place and not another.

branberry (commented):
Hey Raymund! This looks great. Just left some small comments.

mmeigs (approved):
Looks great

api/controllers/v1/github.ts
rayangler: There's currently some logic above to check if the event's header is for the "pull_request" event.

```js
const ghEventType = event.headers['X-GitHub-Event'];
if (ghEventType !== 'pull_request') {
  const errMsg = 'GitHub event type is not of type "pull_request"';
  return {
    statusCode: 400,
    headers: defaultHeaders,
    body: errMsg,
  };
}
```

Just curious if we want to have checks for both, or if we should prioritize checking for the `pull_request` object over the header

api/controllers/v1/github.ts
rayangler: Sure, I like the idea for having a shared directory for logic that might not necessarily have differences between v1 and v2. For something like this, maybe we can have a `api/handlers/github.ts` file? Wdyt?

api/controllers/v1/github.ts
branberry: Oh, totally missed that!

PR #863, opened 2023-07-26
DOP-3868: Use Gatsby Cloud site ID based on user

rayangler:
### Ticket

DOP-3868

### Notes

* Only DOP engineers should have a `gatsby_data_source` available in pre-production for now.
* See "(POST Webhook Status)200" in [this log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=64c15d3f87b3b36046427683) for successful request.
* See "(POST Webhook) Failed to POST to Gatsby Cloud webhook: Error: User rayangler does not have a Gatsby Cloud data source." in [this log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=64c15ed987b3b36046430d60) for unsuccessful request.
* The Gatsby Cloud site associated with the user is not expected to properly build atm. The goal of this PR is to simply call the "correct" site.


src/repositories/repoEntitlementsRepository.ts
```diff
@@ -47,6 +47,21 @@ export class RepoEntitlementsRepository extends BaseRepository {
...
   }
 
+  async getGatsbyDataSourceByGithubUsername(githubUsername: string) {
+    const query = { github_username: githubUsername };
+    const projection = { _id: 0, gatsby_data_source: 1 };
+    const res = await this.findOne(
+      query,
+      `Mongo Timeout Error: Timedout while retrieving entitlements for ${githubUsername}`,
+      { projection }
+    );
```
mmeigs, line 57: I'm guessing we're fetching this because we don't know its format right now and this is easiest to change while in development. In the future, if the preview site is as formulaic as I've gleaned, maybe we can save a fetch and cobble the string together. But that's once this is all nailed down.

mmeigs (approved):
LGTM!

src/repositories/repoEntitlementsRepository.ts
rayangler: The value here is the site's ID, rather than the name of the preview site. I should probably rename these `gatsby_data_source` variables and fields to make this clearer/more intuitive! Probably `gatsby_site_id` would be more accurate

rayangler: Just made the change. Let me know if this makes more sense!

mmeigs: Ah ok, I like that! But definitely my bad for misunderstanding :) sounds good!

rayangler: It's okay! This comment helped me see the light 🙏

PR #867, opened 2023-08-02
DOP-3722: meta - Persistence Module fail ends build

mmeigs:
### Ticket

DOP-3722

### Notes

The updated documents and metadata created by the Persistence Module have become integral to our builds moving forward. To support our continued Build Speed Enhancements project with Gatsby, we need any failure to update these documents to end the entire build. This is split into two PRs. 

This PR alters the meta branch Makefiles to not ignore errors stemming from the Persistence Module and to exit with an error status immediately.

PR #868, opened 2023-08-02
DOP-3722: Allow errors to exit the Persistence Module

mmeigs:
### Ticket

DOP-3722

### Example of Failure

[Logs](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=64ca559b87b3b3604634fc15)
- one difference: the error messaging is cleaner now. No more "bubbles up" :) 

### Notes

This is a companion PR to [PR#867](https://github.com/mongodb/docs-worker-pool/pull/867)

The asynchronous nature of the module was not allowing errors to exit the process with the necessary status. Also, the conditional was blocking a failure status because it never evaluated to a truthy value.

### DIFF
--- a/modules/persistence/index.ts
+++ b/modules/persistence/index.ts
@@ -55,18 +55,32 @@ const app = async (path: string, githubUser: string) => {
   }
 };
 
-try {
-  console.log(argv);
-  app(argv['path'], argv['githubUser']);
-} catch (error) {
+console.log(argv);
+app(argv['path'], argv['githubUser']).catch((error) => {
   console.error('HEY WE ARE IN FINAL CATCH');
   console.error('argv error ', argv);
   console.log('argv ', argv);
-  console.log('caught in terminal handling');
+  console.log('caught in terminal handling BABY', error);
   // only exit with non zero error code if running with strict mode on
   if (['y', 'yes', 'true'].includes(argv['strict'].toLowerCase())) {
     process.exit(1);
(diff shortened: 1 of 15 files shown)


modules/persistence/index.ts
```diff
@@ -55,15 +55,7 @@ const app = async (path: string, githubUser: string) => {
...
-  console.log('caught in terminal handling');
-  // only exit with non zero error code if running with strict mode on
-  if (['y', 'yes', 'true'].includes(argv['strict'].toLowerCase())) {
-    process.exit(1);
-  } else {
-    process.exit(0);
-  }
-}
+app(argv['path'], argv['githubUser']).catch(() => {
+  console.error('Persistence Module Failure. Ending build.');
```
bianca-laube, line 59: I noticed that you're not printing the error that the catch catches. for example:
```
.catch((err) = > 
      console.error('Persistence Module failure, ending build:', err)
```
However I don't know we need to print this, maybe its redundant so please feel free to ignore this comment.

mmeigs: Great callout. We talked in person, but putting here for visibility. I did originally log out the error here, but I realized all of the thrown errors throughout the module already log out the error before throwing, so it resulted in a double log.

seungpark (approved):
LGTM. thanks for _catching_ the non async try catch! 😏

PR #869, opened 2023-08-02
:package: DOP-3840 updates nodejs version to 18.x

caesarbell:


seungpark (changes requested):
failing test can be resolved by updating baseRepository `findOne` and `find` functions. they are catching and logging but not await.

```
      return this.promiseTimeoutS(
        this._config.get('MONGO_TIMEOUT_S'),
        this._collection.findOne(query, options),
        errorMsg
      );
```

to
```
      return await this.promiseTimeoutS(
        this._config.get('MONGO_TIMEOUT_S'),
        this._collection.findOne(query, options),
        errorMsg
      );
```

caesarbell:
> failing test can be resolved by updating baseRepository `findOne` and `find` functions. they are catching and logging but not await.
> 
> ```
>       return this.promiseTimeoutS(
>         this._config.get('MONGO_TIMEOUT_S'),
>         this._collection.findOne(query, options),
>         errorMsg
>       );
> ```
> 
> to
> 
> ```
>       return await this.promiseTimeoutS(
>         this._config.get('MONGO_TIMEOUT_S'),
>         this._collection.findOne(query, options),
>         errorMsg
>       );
> ```

I definitely missed that, but your right the other ones were awaiting. Good catch on that.

seungpark (approved):
LGMT 👍

src/repositories/baseRepository.ts
```diff
@@ -73,7 +73,7 @@ export abstract class BaseRepository {
 
   protected async findOne(query: any, errorMsg: string, options: mongodb.FindOptions = {}): Promise<any> {
     try {
-      return this.promiseTimeoutS(
+      return await this.promiseTimeoutS(
```
rayangler, line 76: Thanks for fixing! Super weird that the lack of `await` ends up breaking the test here! 😮 I would expect that this `await` is unnecessary since the function already returns a promise.

rayangler (approved):
LGTM!

src/repositories/baseRepository.ts
caesarbell: Agree. I was a bit shocked myself, but I guess the `findOne` would return an empty value before the actual `this.promiseTimeoutS` finished because it wasn't awaiting.

PR #870, opened 2023-08-03
DOP-3895: Gatsby Preview URL

mmeigs:
### Ticket

DOP-3895

### Notes

Replaces the dummy url with the Gatsby Preview URL for the user.

Also, if there is a preview url, which means it is a Gatsby Preview build, this preview URL is used in the `GithubCommenter` class in place of the standard staging url.

This preview url structure is still up for debate and depends on implementation.

Note: deletes unused imports as well.

### DIFF
--- a/api/handlers/jobs.ts
+++ b/api/handlers/jobs.ts
@@ -239,7 +239,7 @@ export async function snootyBuildComplete(event: APIGatewayEvent): Promise<APIGa
     const db = client.db(c.get<string>('dbName'));
     const jobRepository = new JobRepository(db, c, consoleLogger);
     const { payload } = await jobRepository.updateWithCompletionStatus(jobId, null, false);
-    const previewUrl = getPreviewUrl();
+    const previewUrl = getPreviewUrl(payload, c.get<string>('env'));
     await notifyBuildSummary(jobId, { mongoClient: client, previewUrl });
   } catch (e) {
     consoleLogger.error('SnootyBuildCompleteError', e);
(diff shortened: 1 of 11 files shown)


api/handlers/jobs.ts
```diff
@@ -235,10 +238,8 @@ export async function snootyBuildComplete(event: APIGatewayEvent): Promise<APIGa
     await client.connect();
     const db = client.db(c.get<string>('dbName'));
     const jobRepository = new JobRepository(db, c, consoleLogger);
-    await jobRepository.updateWithCompletionStatus(jobId, null, false);
-    // Placeholder preview URL until we iron out the Gatsby Cloud site URLs.
-    // This would probably involve fetching the URLs in the db on a per project basis
-    const previewUrl = 'https://www.mongodb.com/docs/';
+    const { payload } = await jobRepository.updateWithCompletionStatus(jobId, null, false);
+    const previewUrl = getPreviewUrl();
```
rayangler, line 242: Should we call this function with the `payload` object?

api/handlers/jobs.ts
```diff
@@ -257,3 +258,15 @@ export async function snootyBuildComplete(event: APIGatewayEvent): Promise<APIGa
...
+/**
+ * Assembles Gatsby Preview URL address for Job post-build
+ * @param payload
+ * @returns string|undefined
+ */
+function getPreviewUrl(payload?: Payload): string | undefined {
+  if (!payload) return;
+  const { repoOwner, branchName, project } = payload;
+  const githubUsernameNoHyphens = repoOwner.split('-').join('').toLowerCase();
+  return `https://preview-mongodb${githubUsernameNoHyphens}.gatsbyjs.io/${project}/${branchName}/index`;
```
rayangler, line 271: If we wanted to support pre-production sites for Gatsby Cloud, we should probably have a way to differentiate them by name. Is it possible to do something like check the Autobuilder's current environment, and if it's `stg`, append `stg` after `preview-mongodb${githubUsernameNoHyphens}`?

mmeigs: That sounds perfect!

api/handlers/jobs.ts
mmeigs: Oh yes, I was testing out it being undefined. So sorry, thanks for catching!!

api/handlers/jobs.ts
mmeigs: @rayangler changed! However, I always get confused about this. Should it encompass '`dotcomstg`' as well?

rayangler: `stg` webhooks should be reserved for GitHub push/content staging jobs while `dotcomstg` should be for test deploys, to sort of emulate the expected `prd` vs. `dotcomprd` behavior in production (but I don't think we enforce nor care too much, especially in pre-production). I suggested only `stg` under the assumption that we'd only ever use this webhook for either `prd` or `stg` data for now

rayangler (approved):
LGTM!

PR #875, opened 2023-08-10
Return the value of the update

rayangler:
`findOneAndUpdate` apparently returns the original document nested in a response object. This PR hopes to fix this by returning the actual value.


mmeigs (approved):
Thank you!

PR #876, opened 2023-08-10
Update deploy-prd-enhanced-worker.yml

branberry:
Add additional files that trigger a build for the enhanced webhook and worker. If these directories are not included, it could cause errors such as missing environment variables. For example, if we add a secure string for the worker in `utils/ssm.ts`, that will not trigger a deploy for the worker infrastructure, leading to errors from the missing environment variable.

### DIFF
--- a/.github/workflows/deploy-prd-enhanced-worker.yml
+++ b/.github/workflows/deploy-prd-enhanced-worker.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    paths: ["src/**", "cdk-infra/lib/constructs/worker/**"]
+    paths: ["src/**", "cdk-infra/lib/constructs/worker/**", "Dockerfile.enhanced", "cdk-infra/utils"]
     tags:
       - 'v*'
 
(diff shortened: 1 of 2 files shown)


schmalliso (approved):

.github/workflows/deploy-prd-enhanced-worker.yml
```diff
@@ -1,6 +1,6 @@
 on:
   push:
-    paths: ["src/**", "cdk-infra/lib/constructs/worker/**"]
+    paths: ["src/**", "cdk-infra/lib/constructs/worker/**", "Dockerfile.enhanced", "cdk-infra/utils"]
```
rayangler, line 3: Just a heads up that pushes for paths aren't evaluated for pushes to tags, based on the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore). 🤔 

> Path filters are not evaluated for pushes of tags.

rayangler (approved):
LGTM! But just wanted to flag some behavior

.github/workflows/deploy-prd-enhanced-worker.yml
branberry: Good catch, @rayangler! I will just remove the paths for the `prd` actions which shouldn't be an issue

rayangler (approved):

.github/workflows/deploy-prd-enhanced-worker.yml
rayangler: Yeah, it's interesting behavior, but makes sense. We can definitely apply this path filter to a pre-prod environment in the future

PR #877, opened 2023-08-10
Add GitHub Actions to deploy enhanced app to staging

branberry:
Add GitHub workflows to deploy to the enhanced app staging environment when a commit is made to either the `master` or `integration` branches


.github/workflows/deploy-stg-enhanced-worker.yml
```diff
@@ -0,0 +1,31 @@
+on:
+  push:
+    paths: ["src/**", "cdk-infra/lib/constructs/worker/**"]
```
rayangler, line 3: Should we add `Dockerfile.enhanced` or is that unnecessary? Would we also need to consider updates to anything in the `modules/` directory or package files?

.github/workflows/deploy-stg-enhanced-webhooks.yml
```diff
@@ -0,0 +1,31 @@
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-2
+      - name: Deploy Enhanced Infrastructure
+        run: |
+          cd cdk-infra/
+          npm ci
+          npm run deploy:enhanced:webhooks -- -c env=dotcomstg
```
rayangler, line 30: Would we want to support deploying to both `dotcomstg` and `stg` environments?

.github/workflows/deploy-stg-enhanced-worker.yml
branberry: Yes to both! I missed the `Dockerfile.enhanced` and I hadn't thought about modules. Will update.

rayangler (approved):

PR #878, opened 2023-08-10
add ci command for webhooks and increase visibility timeout

branberry:
I increased the `visibilityTimeout` property for the queues because there was the following error occurred when attempting to deploy the staging environment:

```
auto-builder-stack-enhancedApp-dotcomstg-webhooks | 9:28:34 PM | CREATE_FAILED        | AWS::Lambda::EventSourceMapping | api/handleJobsLambda/SqsEventSource:autobuilderstackenhancedAppdotcomstgqueuesJobUpdatesQueue862D290D (apihandleJobsLambdaSqsEventSourceautobuilderstackenhancedAppdotcomstgqueuesJobUpdatesQueue862D290D35BB8AD7) Resource handler returned message: "Invalid request provided: Queue visibility timeout: 30 seconds is less than Function timeout: 120 seconds (Service: Lambda, Status Code: 400, Request ID: 14ed7a5c-9f22-4c81-948a-35ae2e028b63)" (RequestToken: cabb6d1e-271d-ac18-479d-157538643662, HandlerErrorCode: InvalidRequest)

```

### DIFF
--- a/cdk-infra/lib/constructs/queue/queues-construct.ts
+++ b/cdk-infra/lib/constructs/queue/queues-construct.ts
@@ -1,3 +1,4 @@
+import { Duration } from 'aws-cdk-lib';
 import { IQueue, Queue } from 'aws-cdk-lib/aws-sqs';
 import { Construct } from 'constructs';
 
@@ -13,6 +14,7 @@ export class AutoBuilderQueuesConstruct extends Construct {
     const jobsQueueDlq = new Queue(this, 'jobsQueueDlq');
 
     const jobsQueue = new Queue(this, 'JobsQueue', {
+      visibilityTimeout: Duration.seconds(120),
       deadLetterQueue: {
         queue: jobsQueueDlq,
         maxReceiveCount,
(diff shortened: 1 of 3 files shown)


cdk-infra/lib/constructs/queue/queues-construct.ts
```diff
@@ -13,6 +14,7 @@ export class AutoBuilderQueuesConstruct extends Construct {
     const jobsQueueDlq = new Queue(this, 'jobsQueueDlq');
 
     const jobsQueue = new Queue(this, 'JobsQueue', {
+      visibilityTimeout: Duration.seconds(120),
```
branberry, line 17: This doesn't cause problems with deploying to `prd`, but creating a new stack without adjusting the `visibilityTimeout` to be greater than or equal to a consuming lambda function's `timeout` leads to the error when deploying that is mentioned in the OP. 

Basically, the `visibilityTimeout` is the period in which a message has been picked up by a consumer, and is no longer visible to other consumers. During this period, it is expected that the consumer that picked up the message will process the message and delete it. Otherwise, when the `visibilityTimeout` duration is reached, the message will now be available to other consumers if it is not deleted.

[A more detailed explanation for this can be found here.](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html)

schmalliso (approved):
This looks good to me, upon review of the linked AWS docs.

casthewiz (approved):

PR #881, opened 2023-08-14
DOP-3943: Mark data for deletion based on user

rayangler:
### Ticket

DOP-3943

### Notes

* Fixes a potential future bug where if 2 users have the data with the same project + branch, the deletion webhook could mark every data with that project + branch for deletion instead of only the user's fork of the data.

### DIFF
--- a/api/handlers/github.ts
+++ b/api/handlers/github.ts
@@ -87,6 +87,7 @@ export const markBuildArtifactsForDeletion = async (event: APIGatewayEvent) => {
 
   const { repository, pull_request: pullRequest } = payload;
   const branch = pullRequest.head.ref;
+  const user = pullRequest.user.login;
 
   const client = new mongodb.MongoClient(c.get('dbUrl'));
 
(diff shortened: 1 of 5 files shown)


api/handlers/github.ts
```diff
@@ -87,6 +87,7 @@ export const markBuildArtifactsForDeletion = async (event: APIGatewayEvent) => {
 
   const { repository, pull_request: pullRequest } = payload;
   const branch = pullRequest.head.ref;
+  const user = pullRequest.user.login;
```
seungpark, line 90: question for myself - where is this payload constructed?

from [this block](https://github.com/octokit/webhooks/blob/v7.1.0/payload-types/schema.d.ts#L5596), it seems like it should be `sender` attribute? CMIIW

seungpark (commented):
left a comment on the payload destructuring. not sure if this should be `user` or `sender`

api/handlers/github.ts
rayangler: I believe it should be the `PullRequestClosedEvent` object, but yes, `sender` is present there too. The only thing I'm unsure/concerned about is if `sender` corresponds with the user who's actually closing or merging the PR, and if the person closing the PR is not the same person who authored the PR, then we could potentially be checking for the wrong user

seungpark: :+1: https://github.com/octokit/webhooks/blob/v7.1.0/payload-types/schema.d.ts#L2819

PR #882, opened 2023-08-15
DOP-3942: Find previous ASTs based on user

rayangler:
### Ticket

DOP-3942

### Notes

* This change should allow 2 or more users to have ASTs for the same project + branch, without conflicting with one another.

### DIFF
--- a/modules/persistence/index.ts
+++ b/modules/persistence/index.ts
@@ -30,19 +30,16 @@ const missingPathMessage = 'No path specified in arguments - please specify a bu
 // Load command line args into a parameterized argv
 const argv: ModuleArgs = minimist(process.argv.slice(2));
 
-const app = async (path: string, githubUser: string) => {
+const app = async (path: string, githubUser?: string) => {
   try {
     if (!path) throw missingPathMessage;
+    const user = githubUser || 'docs-builder-bot';
     const zip = new AdmZip(path);
     // atomic buildId for all artifacts read by this module - fundamental assumption
     // that only one build will be used per run of this module.
     const buildId = new mongodb.ObjectId();
-    const metadata = await metadataFromZip(zip, githubUser);
-    await Promise.all([
-      insertAndUpdatePages(buildId, zip, githubUser),
-      insertMetadata(buildId, metadata),
-      upsertAssets(zip),
-    ]);
+    const metadata = await metadataFromZip(zip, user);
+    await Promise.all([insertAndUpdatePages(buildId, zip, user), insertMetadata(buildId, metadata), upsertAssets(zip)]);
     await insertMergedMetadataEntries(buildId, metadata);
     // DOP-3447 clean up stale metadata
     await deleteStaleMetadata(metadata);
--- a/modules/persistence/src/services/pages/index.ts
+++ b/modules/persistence/src/services/pages/index.ts
@@ -43,29 +43,30 @@ const UPDATED_AST_COLL_NAME = 'updated_documents';
 // Service responsible for memoization of page level documents.
 // Any extraneous logic performed on page level documents as part of upload should be added here
 // or within subfolders of this module
-const pagesFromZip = (zip: AdmZip, githubUser?: string) => {
+const pagesFromZip = (zip: AdmZip, githubUser: string) => {
   const zipPages = zip.getEntries();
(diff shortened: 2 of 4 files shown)


modules/persistence/tests/metadata/metadata.test.ts
```diff
@@ -48,11 +48,6 @@ describe('metadata module', () => {
 
   describe('metadataFromZip', () => {
     it('should get metadata from site.bson', async () => {
-      const metaFromZip = await _metadataFromZip(zip);
-      expect(metaFromZip).toEqual({ ...meta, github_username: 'docs-builder-bot' });
-    });
-
-    it('should add github username to metadata', async () => {
```
rayangler, line 55: Consolidating these two tests since we now want to enforce a GitHub user before calling these functions

modules/persistence/src/services/pages/index.ts
```diff
@@ -43,29 +43,30 @@ const UPDATED_AST_COLL_NAME = 'updated_documents';
...
  */
-const findPrevPageDocs = async (pageIdPrefix: string, collection: string) => {
+const findPrevPageDocs = async (pageIdPrefix: string, collection: string, githubUser: string) => {
   const dbSession = await db();
   const findQuery = {
     page_id: { $regex: new RegExp(`^${pageIdPrefix}/`) },
+    github_username: githubUser,
     deleted: false,
   };
   const projection = {
```
seungpark, line 72: do we want to include in the projection here for safekeeping?

modules/persistence/src/services/pages/index.ts
seungpark, line 1: minor nitpick. i didn't see where we are reading page.github_username but would add to the interface in ln 20

edit: `export interface Page extends Document` and UpdatedPagesManager should have currentPages be a Pages[] type with the constructor and updatePages taking Page[] arguments

modules/persistence/tests/metadata/metadata.test.ts
seungpark: oh i think that answers my question about throwing from index if no user(?)

seungpark (commented):
minor comments below @rayangler ! overall LGTM to associate one user per job -> one user per persistence module call

mmeigs (approved):

modules/persistence/index.ts
```diff
@@ -33,16 +33,13 @@ const argv: ModuleArgs = minimist(process.argv.slice(2));
 const app = async (path: string, githubUser: string) => {
   try {
     if (!path) throw missingPathMessage;
+    const user = githubUser || 'docs-builder-bot';
```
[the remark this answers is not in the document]
rayangler: For now, I'm just leaving the existing behavior of having the default `docs-builder-bot` user

modules/persistence/src/services/pages/index.ts
rayangler: I don't think the `github_username` field is important to project here at the moment. It's not really used within the context of the returned page document

modules/persistence/src/services/pages/index.ts
rayangler: Done! I didn't have `Page` extend `Document` because TS may not call out any errors for any new fields added to `Page` in the future. It might be better to be explicit than to extend `Document` to prevent any false positives for fields.

Let me know if you disagree and I can make the change!

modules/persistence/src/services/pages/index.ts
```diff
@@ -148,7 +156,7 @@ class UpdatedPagesManager {
                 static_assets: this.findUpdatedAssets(page.static_assets, prevPageData?.static_assets),
                 updated_at: this.updateTime,
                 deleted: false,
-                github_username: page.github_username || 'docs-builder-bot',
+                github_username: page.github_username,
```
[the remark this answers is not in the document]
rayangler: Thanks for calling out! Also didn't realize that upserted documents will automatically include fields from the filter 🤠

modules/persistence/src/services/pages/index.ts
seungpark: i'm fine with leaving it out, can re insert if we need to for debugging purposes

seungpark (approved):

PR #884, opened 2023-08-18
DOP-3801: Ensure db pools/topologies are used efficiently

mmeigs:
### Ticket

DOP-3801

### Notes

The Persistence Module already had functionality implemented to cache db connections, but the asynchronous nature of the Promise.all call was not allowing time for the first connection to succeed before each function in turn needed it. This led to four connections rather than the desired one to be shared.

Both db connections needed for the module are now awaited, created, and cached before the module moves on.

### DIFF
--- a/modules/persistence/index.ts
+++ b/modules/persistence/index.ts
@@ -5,7 +5,7 @@ dotenv.config();
 import AdmZip from 'adm-zip';
 import minimist from 'minimist';
 import * as mongodb from 'mongodb';
-import { teardown as closeDBConnection } from './src/services/connector';
+import { db as snootyDb, pool as poolDb, teardown as closeDBConnection } from './src/services/connector';
 import { insertAndUpdatePages } from './src/services/pages';
 import {
   insertMetadata,
@@ -38,6 +38,9 @@ const app = async (path: string, githubUser: string) => {
     // that only one build will be used per run of this module.
     const buildId = new mongodb.ObjectId();
     const metadata = await metadataFromZip(zip, githubUser);
+    // initialize db connections to handle shared connections
+    await snootyDb();
+    await poolDb();
     await Promise.all([
       insertAndUpdatePages(buildId, zip, githubUser),
       insertMetadata(buildId, metadata),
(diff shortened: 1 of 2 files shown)


modules/persistence/index.ts
```diff
@@ -38,6 +38,9 @@ const app = async (path: string, githubUser: string) => {
     // that only one build will be used per run of this module.
     const buildId = new mongodb.ObjectId();
     const metadata = await metadataFromZip(zip, githubUser);
+    // initialize db connections to handle shared connections
+    await snootyDb();
+    await poolDb();
```
rayangler, line 43: Since the 2 dbs are initiated from the same client connection, I think it might be safe to just initiate one of them. But no worries if we want to keep both for safety

rayangler (approved):
Can confirm that logging topology opening/closing is no longer repeated with this change locally. 😎🤞

seungpark (approved):

modules/persistence/index.ts
mmeigs: Oh, but does that still work if they are connecting to different databases in these different functions?

rayangler: From what I could tell, yes? I commented out one of them, rebuilt, then reran the module, and it seemed like the result was the same. But again, I'm totally okay with keeping both for safety

mmeigs: When I run it locally, one of them (poolDb) is never called. So, no matter what you won't see multiple connections locally from the poolDb.

mmeigs: I really should learn more about this database called MongoDB Atlas :)

rayangler: Hm, I'd imagine `poolDb` would still eventually be called by one of the functions further down the module. But all good. Down to keep as-is to avoid any weird behavior with connections 😨

seungpark: these still might have a [race condition (?) ](https://github.com/mongodb/node-mongodb-native/blob/v4.9.1/src/operations/connect.ts#L18-L19) if the topology is making simultaneous connections?

think ideally we would call client.connect once

rayangler: Both dbs are referenced through the same/single client, so as long as `client.connect` is called once (by `await snootyDb()`, for example) before everything else, I don't think there would be a race condition in simultaneous connections. Any future calls to respective dbs should just attempt to connect, see it's already connected, and then set the `dbInstance` as needed

PR #889, opened 2023-08-22
DOP-3951: Call Gatsby Cloud webhook after persistence module

rayangler:
### Ticket

DOP-3951

### Notes

Multiple related Gatsby Cloud-related behaviors have been brushed up in this PR to prepare for UAT. The ticket's A/Cs should cover this, but for convenience:

* The Autobuilder now makes a call to the persistence module as a separate Makefile command.
* The Autobuilder now calls the Gatsby Cloud preview webhook after the new persistence module command rather than at the end of the job. This will allow Gatsby Cloud builds to be kicked off in parallel to Autobuilder frontend builds.
* The Autobuilder will no longer trigger a database update to mark content staging jobs for completion if the user has a Gatsby Cloud site assigned to them. This will prevent (1) repeated build summaries on Slack and (2) repeated marking of job completion. Behavior should remain the same for users who do not have Gatsby Cloud sites.
* Start and end times for Gatsby Cloud builds are now added to the job document.


api/handlers/jobs.ts
```diff
@@ -239,6 +240,7 @@ export async function snootyBuildComplete(event: APIGatewayEvent): Promise<APIGa
     await client.connect();
     const db = client.db(c.get<string>('dbName'));
     const jobRepository = new JobRepository(db, c, consoleLogger);
+    await jobRepository.updateExecutionTime(jobId, { gatsbyCloudEndTime: new Date() });
```
rayangler, line 243: Inserting as a `Date` object instead of as a `number` from `performance.now()` (which the other benchmarked steps use) since the context of `performance` may be lost between the processes that call start and end, respectively.

api/handlers/jobs.ts
```diff
@@ -270,5 +272,5 @@ function getPreviewUrl(payload: Payload | undefined, env: string): string | unde
   const { repoOwner, branchName, project } = payload;
   const githubUsernameNoHyphens = repoOwner.split('-').join('').toLowerCase();
   const possibleStagingSuffix = env === 'stg' ? 'stg' : '';
-  return `https://preview-mongodb${githubUsernameNoHyphens}${possibleStagingSuffix}.gatsbyjs.io/${project}/${branchName}/index`;
```
rayangler, line 273: This has been fixed on the source plugin side, so making the change here to reflect this

src/job/jobHandler.ts
```diff
@@ -303,6 +317,12 @@ export abstract class JobHandler {
         const makeCommandsResp = await this._commandExecutor.execute([command]);
         await this.logBuildDetails(makeCommandsResp);
       }
+
+      // Call Gatsby Cloud preview webhook after persistence module finishes for staging builds
+      const isFeaturePreviewWebhookEnabled = process.env.GATSBY_CLOUD_PREVIEW_WEBHOOK_ENABLED?.toLowerCase() === 'true';
+      if (key === 'persistence-module' && this.name === 'Staging' && isFeaturePreviewWebhookEnabled) {
+        await this.callGatsbyCloudWebhook();
```
rayangler, line 324: I'd rather not expose this call as another Makefile rule (plus we may not want to expose the user's site ID). I thought about potentially allowing the executor be compatible with both `make` commands and function calls, but it seemed like a bigger change than necessary for the scope of this PR

src/job/jobHandler.ts
```diff
@@ -635,27 +655,43 @@ export abstract class JobHandler {
...
-    const previewWebhookURL = 'https://webhook.gatsbyjs.com/hooks/data_source';
-    const githubUsername = this.currJob.user;
-    const gatsbySiteId = await this._repoEntitlementsRepo.getGatsbySiteIdByGithubUsername(githubUsername);
-    if (!gatsbySiteId) {
-      const message = `User ${githubUsername} does not have a Gatsby Cloud Site ID.`;
-      this._logger.warn('Gatsby Cloud Preview Webhook', message);
-      throw new Error(message);
-    }
+  // Invokes Gatsby Cloud Preview Webhook
+  protected async callGatsbyCloudWebhook(): Promise<void> {
```
rayangler, line 659: Logic should be relatively untouched. Consolidating the logic from 2 separate functions into 1 since the logic no longer needs to live in 2 different places.

tests/data/jobDef.ts
```diff
@@ -51,6 +51,47 @@ export const getBuildJobPlain = (): Job => ({
   } as Payload,
 });
 
+export const getStagingJobDef = (): Job => ({
```
rayangler, line 54: For some reason, we were using a `Job` representative of a deployment in `stagingJobHandler` tests. Created this separate function to be more accurate and to help pass the new test

tests/utils/jobHandlerTestHelper.ts
```diff
@@ -56,20 +69,32 @@ export class JobHandlerTestHelper {
...
+    if (hasGatsbySiteId) {
+      this.repoEntitlementsRepo.getGatsbySiteIdByGithubUsername.mockResolvedValue('gatsby_site_id');
+    }
+
     if (returnValue) {
-      this.jobCommandExecutor.execute.mockReturnValueOnce(returnValue);
+      this.jobCommandExecutor.execute.mockResolvedValue(returnValue);
     } else {
-      this.jobCommandExecutor.execute.mockReturnValueOnce({ status: 'success', output: 'Great work', error: null });
-      this.jobCommandExecutor.execute.mockReturnValueOnce({ status: 'Failed', output: publishOutput[0], error: null });
```
rayangler, line 72: Removing this line since it seems odd to overwrite the mocked value for success with a status representative of failure. Also `mockResolvedValue` is preferred for async functions.

Unfortunately, I could not do something similar to the function that sets up the job handler for failure, without needing to delve deeper into the code to refactor. Fixing this behavior for the other function is out of scope

mmeigs (approved):
LGTM

src/job/jobHandler.ts
```diff
@@ -635,27 +655,43 @@ export abstract class JobHandler {
...
+          jobId: this.currJob._id,
+        },
+        {
+          headers: { 'x-gatsby-cloud-data-source': 'gatsby-source-snooty-preview' },
+        }
+      );
+      await this._jobRepository.updateExecutionTime(this.currJob._id, { gatsbyCloudStartTime: new Date() });
+      await this.logger.save(this.currJob._id, `${'(POST Webhook Status)'.padEnd(15)}${response.status}`);
+    } catch (err) {
+      await this.logger.save(
```
seungpark, line 690: minor, should we throw here to kill execution? 

also, would prefer `logger.error` over .save

src/job/jobHandler.ts
```diff
@@ -635,27 +655,43 @@ export abstract class JobHandler {
...
+        this._logger.warn('Gatsby Cloud Preview Webhook', message);
+        throw new Error(message);
       }
-    );
+
+      const url = `${previewWebhookURL}/${gatsbySiteId}`;
+      const response = await axios.post(
+        url,
+        {
+          jobId: this.currJob._id,
```
branberry, line 681: I noticed that the the post request had the `status` property as a part of the object body. Was this intentionally removed?

seungpark (approved):
minor feedback on try catch block but otherwise LGTM

branberry (approved):
LGTM! Just had one question for clarification

src/job/jobHandler.ts
rayangler: Good call on throwing.

I want to say that `save` in this case will print it out in the build log and in CloudWatch, while `error` would just log it in CloudWatch. It might be good to have this in both places for visibility so that users can surface the webhook error from the log if needed?

src/job/jobHandler.ts
rayangler: Yes, this is intentional. Right now, the source plugin doesn't do anything with the `status` (which should always be `inProgress` in this case) that it receives

src/job/jobHandler.ts
seungpark: i'm fine with throwing and saving log 👍 

in theory, this should be a failed job and updated with error status later on

PR #898, opened 2023-08-25
Print hotfix

branberry:


rayangler (approved):
Ah, this makes sense. I should've caught this during the initial review!

PR #899, opened 2023-08-28
DOP-3973: Remove additional call to persistence module

rayangler:
### Ticket

DOP-3973

### Notes

* Removes the call to the persistence module for individual Makefiles.
* I haven't removed the conditional to set `GH_USER_ARG` for the individual Makefiles because I think the existing `shared.mk` only defines it if the individual Makefiles do not define custom build steps. I don't think it should have any negative impact to leave as-is, but can probably be cleaned up in the future

### DIFF
--- a/makefiles/Makefile.docs
+++ b/makefiles/Makefile.docs
@@ -61,13 +61,6 @@ GH_USER_ARG=${GH_USER}
 endif
 
 next-gen-html:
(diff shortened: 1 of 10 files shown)


seungpark (approved):
🙌

mmeigs (approved):
Ohhh fantastic!

PR #901, opened 2023-08-29
DOP-3963: Render only latest version of spec at base spec component

casthewiz:
# Context
DOP-3963 has most details, but TLDR: 

We want to render only the latest spec at the base route of any versioned OpenAPI spec - this resolves some out-of-order issues with resources in dropdowns vs version in the left sidenav, and consumers of the spec typically want to see the latest versions regardless.

## Staging Link
https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/atlas/master/reference/api-resources-spec/v2/

### DIFF
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,7 +3,6 @@ on:
     branches:
       - "master"
       - "integration"
-      - "DOP-3963"
 concurrency:
   group: environment-stg-${{ github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/deploy-stg-ecs.yml
+++ b/.github/workflows/deploy-stg-ecs.yml
@@ -3,7 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
-      - "POC-base-spec"
+      - "DOP-3963"
 concurrency:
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy-stg-ecs.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-3963"
```
```suggestion
```

Just a reminder for myself to remove this before merge

modules/oas-page-builder/tests/unit/services/pageBuilder.test.ts
```diff
@@ -192,7 +192,7 @@ describe('pageBuilder', () => {
 
     await buildOpenAPIPages(testEntries, testOptions);
 
-    expect(mockExecute).toBeCalledTimes(testEntries.length * 2);
+    expect(mockExecute).toBeCalledTimes(4);
```
casthewiz, line 195: Given that we specify `toBeCalledWith` for each invocation, I'm fine with hardcoding the number of calls now that the logic isn't as straightforwards.

Open to any other equations that determine that intended number of calls though.

mmeigs (approved):
LGTM!

PR #902, opened 2023-08-31
DOP-3977: encourage writers to switch webhooks 🥕

schmalliso:
Adds blurb to encourage writers to update to the new webhook. I am very open to copy enhancement suggestions.

Message for build on non-enhanced stg webhook:
Screenshot 2023-08-31 at 12 17 16 PM

Message for build on enhanced stg webhook:
Screenshot 2023-08-31 at 12 20 23 PM

Message for /test-deploy:
Screenshot 2023-08-31 at 12 20 01 PM

### DIFF
--- a/api/controllers/v2/jobs.ts
+++ b/api/controllers/v2/jobs.ts
@@ -122,7 +122,7 @@ function prepProgressMessage(
         const webhookWikiUrl =
           'https://wiki.corp.mongodb.com/display/DE/How-To%3A+Use+Snooty%27s+Autobuilder+to+Build+Your+Content';
         const updatePlease = `:exclamation: You used the old webhook for this build. <${webhookWikiUrl}|Update to the new webhook> in your fork of this repo to save 90s per build.`;
-        inQueueMsg = updatePlease + '\n' + msg;
+        inQueueMsg = updatePlease + '\n\n' + msg;
       }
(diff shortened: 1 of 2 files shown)


api/controllers/v2/jobs.ts
```diff
@@ -108,13 +108,23 @@ function prepProgressMessage(
...
+
   switch (status) {
     case 'inQueue':
-      return msg + 'has successfully been added to the ' + env + ' queue.';
+      // Encourage writers to update to new webhook on githubPush jobs
+      let inQueueMsg = msg;
+      if (jobType == 'githubPush') {
+        const webhookWikiUrl =
+          'https://wiki.corp.mongodb.com/display/DE/How-To%3A+Use+Snooty%27s+Autobuilder+to+Build+Your+Content';
+        const updatePlease = `:exclamation: You used the old webhook for this build. <${webhookWikiUrl}|Update to the new webhook> in your fork of this repo to save 90s per build.`;
```
rayangler, line 124: This v2 file should only be for the enhanced Autobuilder. I'm actually not sure if this case would ever be hit, or if we would even want to. I think it might be safe to omit either way

rayangler (approved):

branberry (approved):
LGTM!

PR #903, opened 2023-08-31
[FEATURE]: Comment on PRs with webhook URL for feature branch

branberry:
Added functionality for the initial deploy of the feature branch infrastructure to comment on the PR with the webhook URL.

The last build error can be ignored. I cancelled the feature branch cleanup workflow, so it marks the commit as a failure.

### DIFF
--- a/.github/scripts/webhook-pr-comment.js
+++ b/.github/scripts/webhook-pr-comment.js
@@ -1,7 +1,7 @@
 const fs = require('fs');
 
 module.exports = ({ github }) => {
-  const outputsFile = fs.readFileSync('outputs.json').toString();
+  const outputsFile = fs.readFileSync('cdk-infra/outputs.json').toString();
   const outputs = JSON.parse(outputsFile);
   const webhook = Object.values(outputs[`auto-builder-stack-enhancedApp-stg-${github.head_ref}-webhooks`])[0];
   return webhook;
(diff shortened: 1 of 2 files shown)


.github/scripts/webhook-pr-comment.js
```diff
@@ -0,0 +1,14 @@
+const fs = require('fs');
+const path = require('path');
+module.exports = () => {
```
branberry, line 3: I created this script because I couldn't figure out from the documentation how to get properties with dashes using the `fromJson`. Also, the property that holds the webhook URL is also an object that contains an arbitrary key, so using the script made it way easier to pluck it out.

.github/scripts/webhook-pr-comment.js
```diff
@@ -0,0 +1,14 @@
...
+  try {
+    const outputsFile = fs.readFileSync('cdk-infra/outputs.json').toString();
+    const outputs = JSON.parse(outputsFile);
+
+    const webhook = Object.values(outputs[`auto-builder-stack-enhancedApp-stg-${process.env.GIT_BRANCH}-webhooks`])[0];
```
branberry, line 8: the value for the `auto-builder-stack-enhancedApp-stg-${process.env.GIT_BRANCH}-webhooks` property contains an object that looks something like:

``` json
 {
  "AS12FWebhookEWE312": "url"
 }
```

And that key is different, so it's not really easy to predict how to index it, so I just get the first value in the array, since there should only be one.

rayangler (approved):

PR #904, opened 2023-09-06
DOP-3999 Exclude oldgen jobs from "don't send notifications of non-GC builds" logic

anabellabuckvar:
[DOP-3999](https://jira.mongodb.org/browse/DOP-3999)

Current Behavior:
Completed job notifications from autobuilder-built front-end jobs for users using Gatsby Cloud are intentionally suppressed. People who are in the Gatsby Cloud UAT group but also building oldgen aren't getting their notifications about oldgen builds

Use below URL for feature branch webhook with fork of [docs-k8s operator](https://github.com/10gen/docs-k8s-operator) or [mms-docs](https://github.com/10gen/mms-docs) and ensure that you're part of Gatsby UAT to test functionality: Gatsby Cloud UAT group users building oldgen should get job-complete notifications


src/job/jobHandler.ts
```diff
@@ -105,7 +105,11 @@ export abstract class JobHandler {
         // completed after the Gatsby Cloud build via the SnootyBuildComplete lambda.
         const { _id: jobId, user } = this.currJob;
         const gatsbyCloudSiteId = await this._repoEntitlementsRepo.getGatsbySiteIdByGithubUsername(user);
-        if (gatsbyCloudSiteId && this.currJob.payload.jobType === 'githubPush') {
+        if (
+          !['mms-docs', 'docs-k8s-operator', 'docs-php-library'].includes(this.currJob.payload.repoName) &&
```
rayangler, line 109: `this.currJob.payload` has an `isNextGen` field. Is this field something that we could leverage instead of a hardcoded array of oldgen repos?

anabellabuckvar: That would make sense! I can change it, but [DOP-3999](https://jira.mongodb.org/browse/DOP-3999) specified that "oldgen jobs will be jobs for: docs-k8s-operator, mms-docs, docs-php-library (there may be others, but they are edge cases and/or EOL'd branches we don't care about as much)" which is why I did it this way.

schmalliso: The ticket writer may have not remembered we have an `payload.isNextGen` field 😁. My bad - Raymund's suggested approach is preferable, IMO.

rayangler: Gotcha, that makes sense. (Admittedly, I did not read the full ticket so I missed a bit of context there.) 

I think as long as the `isNextGen` field works (verification would be nice, just in case), leveraging that field might be the best way to go to avoid us from having to remember to update this array in the future, if any of the listed repos are converted to nextgen.

schmalliso (approved):

schmalliso (approved):
Approved for reals this time.

rayangler (approved):

PR #905, opened 2023-09-07
DOP-3994: add feature flag for chatbot

seungpark:
DOP-3994 : adding [front end feature flag](https://github.com/mongodb/snooty/pull/884/files#diff-a89ac12c8ec97364db532676bd572c7e552dcb933901a3d3358fa0285fabe445R12) via front end env variable `GATSBY_SHOW_CHATBOT`

Build log: https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=64fa641624fcc731b43f6e6d

gatsby cloud build: https://preview-mongodbseungparkstg.gatsbyjs.io/landing/test-deploy/

S3 build (with flag turned on in `stg` environment): https://docs-mongodb-org-stg.s3.us-east-2.amazonaws.com/landing/docsworker-xlarge/test-deploy/

**note:** param store has these values set to `false` <string>

image


src/job/jobHandler.ts
```diff
@@ -369,6 +369,7 @@ export abstract class JobHandler {
       GATSBY_BASE_URL: this._config.get<string>('gatsbyBaseUrl'),
       PREVIEW_BUILD_ENABLED: this._config.get<string>('previewBuildEnabled'),
       GATSBY_TEST_SEARCH_UI: this._config.get<string>('featureFlagSearchUI'),
+      GATSBY_SHOW_CHATBOT: this._config.get<string>('gatsbyUseChatbot'),
```
branberry, line 372: I know this PR came before the Slack discussion last Friday, but would you like to use the regular env variables as opposed to using the node config library?

branberry (approved):
LGTM! I left one comment with regards to the discussion last Friday about getting feature flags right from `process.env` as opposed to using the node-config library. Doesn't need to be done now, but just putting it out there 👍

src/job/jobHandler.ts
seungpark: i think we can leave this as a conversion across all env vars instead of a mix usage. WDYT?

PR #913, opened 2023-09-21
DOP-3723: Use Autobuilder job id as persistence module build id

rayangler:
### Ticket

DOP-3723

### Notes
* Passes Autobuilder job ID to the persistence module so that build artifacts can be associated with a single build. By default or when the job ID argument is false-y, the persistence module will generate its own build ID like it currently does.
* When there are updates to an `updated_document`, the build ID that triggered the update will be added to the document. This should help with increased observability of document updates.
* Related Makefile change can be found at #915.
* See job ID passed to the persistence module in this [build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=650db41e492df295cfdba545). Related screenshot with the same build ID in `updated_documents` can be found below.

Screen Shot 2023-09-22 at 2 16 03 PM


branberry (approved):
LGTM

seungpark (approved):
these changes LGTM

two questions:
- when are we reading for `updated_pages.job_id`, and should we be indexing this new field for any performance gains?
- seems like the pages.test has been outdated but still passing? ie. `_updatedPages` takes in 4 args, 2 are defined in tests

rayangler:
@seungpark 

> when are we reading for updated_pages.job_id, and should we be indexing this new field for any performance gains?

The field may be useful for @schmalliso's metrics for Gatsby Cloud builds to see which Autobuilder job last updated a page. I would be down to add an index once the field is in use.

> seems like the pages.test has been outdated but still passing? ie. _updatedPages takes in 4 args, 2 are defined in tests

Wow, thanks for catching this! I updated the tests accordingly

PR #917, opened 2023-09-26
[DOP-4033]: Replicate shared.mk steps within Autobuilder code

branberry:
## Ticket

[DOP-4033](https://jira.mongodb.org/browse/DOP-4033)

## Notes

This story converts the contents of the `shared.mk` file into TypeScript code. 

## Testing

1. Checkout the branch and run `npm ci` as there are some new dependencies
2. Run `docker buildx build --file Dockerfile.local . --build-arg NPM_BASE_64_AUTH=<NPM_BASE_64_AUTH> --build-arg NPM_EMAIL=<NPM_EMAIL> -t autobuilder/local:latest `; the email and auth values can be found in parameter store
3. Update the default profile in `~/.aws/credentials` with temporary credentials from AWS; the credentials will have a randomly named profile that you should rename to default when pasting into the file
4. Run `docker run  -v ~/.aws:/home/docsworker/.aws -e ENHANCED=true -e MONGO_ATLAS_USERNAME="<ATLAS_USERNAME>" -e MONGO_ATLAS_PASSWORD="<ATLAS_PASSWORD>"  -e MONGO_ATLAS_HOST="<ATLAS_HOST>" autobuilder/local:latest` using the staging environment values


src/entrypoints/localApp.ts
```diff
@@ -0,0 +1,74 @@
+import { nextGenParse } from '../commands/src/shared/next-gen-parse';
+import { nextGenHtml } from '../commands/src/shared/next-gen-html';
+import { prepareBuildAndGetDependencies } from '../commands/src/helpers/dependency-helpers';
+import { nextGenStage } from '../commands/src/shared/next-gen-stage';
+import { oasPageBuild } from '../commands/src/shared/oas-page-build';
+import { persistenceModule } from '../commands/src/shared/persistence-module';
+import { nextGenDeploy } from '../commands/src/shared/next-gen-deploy';
+
+async function localApp() {
+  // TODO: Fetch this from repos_branches
```
branberry, line 10: The `localApp.ts` is used by the `Dockerfile.local`. Created this to test and verify that the commands are working as expected.

src/commands/src/shared/next-gen-stage.ts
```diff
@@ -0,0 +1,46 @@
...
+export async function nextGenStage({
+  repoDir,
+  mutPrefix,
+  projectName,
+  bucket,
+  url,
+  patchId,
+  commitBranch,
+}: StageParams) {
+  let hostedAtUrl = `${url}/${mutPrefix}/docsworker/${commitBranch}/`;
```
rayangler, line 22: Should `docsworker` be `docsworker-xlarge` to maintain existing staging url path? Or are we intentionally changing it here?

src/commands/src/shared/next-gen-stage.ts
```diff
@@ -0,0 +1,46 @@
...
+  let prefix = mutPrefix;
+
+  const commandArgs = ['public', bucket, '--stage'];
+
+  if (patchId && projectName === mutPrefix) {
+    const [commitHash, patchId] = await Promise.all([getCommitHash(repoDir), getPatchId(repoDir)]);
```
rayangler, line 28: Question: Why do we get the commit hash and patch ID again? Is there a reason we don't pass in and use `patchId` and `commitHash` from the call to `prepareBuildAndGetDependencies` earlier?

src/commands/src/shared/next-gen-stage.ts
branberry: Ah, I changed this for the local testing, but you're right. I'll need to rename it to `docsworker-xlarge` for when we use regular builds with these functions

src/commands/src/helpers/index.ts
```diff
@@ -0,0 +1,279 @@
...
+    executedCommand.on('close', (exitCode) => {
+      if (writeStream) writeStream.end();
+
+      if (exitCode !== 0) {
+        console.error(`ERROR! The command ${command} closed with an exit code other than 0: ${exitCode}.`);
+        console.error('Arguments provided: ', args);
+        console.error('Options provided: ', options);
+
+        if (outputText) {
+          console.error(outputText.join());
```
caesarbell, line 189: @branberry just out of curiosity, is there a reason why if there is `outputText` we are passing that to a `console.error`? Is there a case where this would provide an error message? I thought `errorText` was for capturing error messages.

src/commands/src/helpers/index.ts
```diff
@@ -0,0 +1,279 @@
...
+      // exits without error
+      if (hasRejected) return;
+
+      if (exitCode !== 0) {
+        console.error(`ERROR! The command ${cmdToParams.command} closed with an exit code other than 0: ${exitCode}.`);
+        console.error('Arguments provided: ', cmdToParams.args);
+        console.error('Options provided: ', cmdToParams.options);
+
+        if (outputText) {
+          console.error('output', outputText.join());
```
caesarbell, line 123: @branberry same here as well? Is there a reason why we aren't using a different `console` , i.e. `console.log` or `console.info`.

caesarbell:
I may have missed this, but would we want to keep the benchmarking in the same place, or would we want to have it in the `localApp.ts` file or something similar if we have something like that for production use?

caesarbell:
Would we want to add the instructions (steps) that are in the description live in the README somewhere?

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+  // doing these in parallel
+  const commandPromises = [
+    getCommitHash(repoDir),
+    getCommitBranch(repoDir),
+    getPatchId(repoDir),
+    existsAsync(path.join(process.cwd(), 'config/redirects')),
+    createEnvProdFile(repoDir, projectName, baseUrl),
+  ];
+
+  const dependencies = await Promise.all(commandPromises);
```
caesarbell, line 50: @branberry would we want to provide a way of catching or handling the error message if one of these promises are rejected.

src/commands/src/shared/next-gen-stage.ts
```diff
@@ -0,0 +1,46 @@
+import { executeCliCommand, getCommitHash, getPatchId } from '../helpers';
+
+interface StageParams {
```
caesarbell, line 3: nit: would we want to just extend off of `NextGenParseParams` interface, since those types are shared with `StageParams`

src/commands/src/shared/next-gen-stage.ts
caesarbell: nit: would we want to assign `docsworker` to a constant since it is used in two places?

src/commands/src/shared/next-gen-stage.ts
```diff
@@ -0,0 +1,46 @@
+import { executeCliCommand, getCommitHash, getPatchId } from '../helpers';
```
caesarbell, line 1: @branberry do we want to remove `getCommitHash` and `getPatchId` as they seem to be not be used in this file anymore.

src/commands/src/helpers/index.ts
branberry: This is because a command can error out, but it still might have outputted to `stdout`. This is so we are able to see everything that was logged by this command. For example, if a command fails halfway through, there will be non-error logs that could be helpful to see, so I output it in case it does exist. I probably should check for `outputText.length` now that I think about it though.

src/commands/src/helpers/dependency-helpers.ts
branberry: That's a good question, that is something I have been thinking about. I was thinking that maybe the error handling could occur in the place where this function gets called to simplify this code, but I am not set in stone on that.

src/commands/src/shared/next-gen-stage.ts
branberry: That's an option for sure, but I think it might be a bit more readable to list them all out. I also was thinking these interfaces to be isolated specifically for each function so they wouldn't need to reference another interface for another command.
 
 That's me being nitpicky though, haha! What are your thoughts?

src/commands/src/helpers/index.ts
caesarbell: Thanks for the explanation. I agree that it would probably make more sense to check the length because an empty array will still be truthy. Would it also make sense to maybe use `console.log` or `console.info` since `console.error` might be misleading because that output wasn't an error rather it was more information about at what point the command had an issue?

src/commands/src/helpers/dependency-helpers.ts
caesarbell: @branberry this was a pretty good [read](https://kinsta.com/blog/errors-in-javascript/#2-do-not-swallow-any-exceptions), on how and where to raise an error to be handled.

src/commands/src/shared/next-gen-stage.ts
caesarbell: I think that is fair. and well thought out.

src/commands/src/helpers/index.ts
branberry: Yeah! I think that makes sense. I'll change this to be a `console.log` instead.

rayangler (approved):
Lgtm!

PR #919, opened 2023-10-01
:triangular_flag_on_post: DOP-4060 adds new featrue flag to the docs-worker-pool

caesarbell:
This PR introduces the new feature flag for hiding the locale on the unified footer on the docs page in production and showing it for non-product environments.

### DIFF
--- a/cdk-infra/lib/constructs/worker/worker-env-construct.ts
+++ b/cdk-infra/lib/constructs/worker/worker-env-construct.ts
@@ -37,7 +37,7 @@ export class WorkerEnvConstruct extends Construct {
     );
 
     // font-end feature flag for unified footer locale selector
-    const unifiedFooterLocale = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/hide_locale`);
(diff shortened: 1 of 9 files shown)


seungpark (approved):
this LGTM. on a minor note, seems like this env flag is set to false? `/env/legacy/docs/worker_pool/flag/hide_locale` not sure if we want to show the locale selection for legacy

caesarbell:
> this LGTM. on a minor note, seems like this env flag is set to false? `/env/legacy/docs/worker_pool/flag/hide_locale` not sure if we want to show the locale selection for legacy

@seungpark I can change this. Legacy is a non-production env correct?

branberry (approved):
LGTM!

schmalliso:
> > this LGTM. on a minor note, seems like this env flag is set to false? `/env/legacy/docs/worker_pool/flag/hide_locale` not sure if we want to show the locale selection for legacy
> 
> @seungpark I can change this. Legacy is a non-production env correct?

It is, but it also probably doesn't matter since (a) the Legacy environment doesn't work and (b) doesn't build with the consistent nav.

PR #925, opened 2023-10-11
DOP-4044: Get repo branches for non-internal doc in docset

rayangler:
### Ticket

DOP-4044

### Notes

This PR adds `internalOnly: false` as a field in the match filter in `getRepoBranchesEntry` and `getAllAssociatedRepoBranchesEntries`. We can make the assumption that the single non-internal-only document in a docset contains the most complete data (branches/versions) that would be expected from a docset since internal docs are often used for privacy reasons with minimal branch data. If this assumption changes in the future, then the match filter may be incorrect.

Checking the `prodDeployable` field may also be possible, but may be awkward if there's a site that isn't currently deployable, but should still have its data pulled in. The switch should be easy, so the solution is flexible here if it's needed to change.


modules/persistence/src/services/metadata/repos_branches/index.ts
```diff
@@ -82,7 +83,7 @@ export const getAllAssociatedRepoBranchesEntries = async (metadata: Metadata) =>
 
   try {
     const db = await pool();
-    const aggregationPipeline = getAggregationPipeline({ project: { $in: fetch } });
+    const aggregationPipeline = getAggregationPipeline({ project: { $in: fetch }, prodDeployable: true });
```
rayangler, line 86: AC in the original ticket called for adding this to only `getRepoBranchesEntry`, but it seems like this might be useful for the `getAllAssociatedRepoBranchesEntries` function as well. If there is a project with more than 1 repo (1 deployable, 1 not deployable) in it, both repos could've been returned, but I'm not sure if that's desired. Let me know!

mmeigs: Good catch, yes - this is needed!

mmeigs (commented):
This should work! For a second I was unsure about what will happen with our switch to sourcing from the monorepo, because that is the other use-case for `prodDeployable`, but I believe it should work how we expect it to! It should be an immediate toggle for everyone, so there shouldn't be any issue with always pinning the `prodDeployable` document. 

Although, I think we might want to ask Cassidy when he's back. Because now I'm wondering if we want to have jobs still available for those separate repos apart from their directory in the monorepo... My mind might be just running in circles, though.

rayangler:
@casthewiz any thoughts on @mmeigs's comment above? Wondering if there are any potential issues with querying for only the prod-deployable document. I believe it should be fine since I don't think the persistence module would ever need non-prod-deployable versions, especially since we return early for builds on staging branches... but wanted to confirm

modules/persistence/src/services/metadata/repos_branches/index.ts
```diff
@@ -114,20 +115,31 @@ export const getRepoBranchesEntry = async (project: project, branch = ''): Promi
...
+    const returnedEntry = res[0];
+
+    if (res.length > 1) {
+      console.warn(
+        `Expected 1 deployable repo for docset with project "${project}", but found ${res.length} instead. Defaulting to first found: "${returnedEntry.repoName}".`
+      );
+    }
 
     // if not already set, set cache value for repo_branches
     if (!internals[project]) {
```
caesarbell, line 139: @rayangler, this might be a silly question, but I see we are caching the value, but wouldn't we want to check the cache before querying the `docsets`?

rayangler: The cache is checked earlier [here](https://github.com/mongodb/docs-worker-pool/blob/b0253fb8c4c9e9c9437f0b23f6acda2197a29196/modules/persistence/src/services/metadata/repos_branches/index.ts#L105).

caesarbell (approved):
LGTM

rayangler:
Spoke offline with Cassidy about this and I switched the focus from querying against `prodDepoyable` to `internalOnly`. This might allow us more flexibility in which data we fetch, especially with all of the monorepo work going on. Plus, it's super easy to switch between whichever one we need accordingly cc @mmeigs

mmeigs (approved):

PR #931, opened 2023-11-02
:pencil: DOP-4098 updates comment

caesarbell:
Part of me searching through all DOP codebases to ensure we are not pulling in `prefix`, `bucket`, and `URL` from repo_branches, I stumbled across an outdated comment, and this PR updates that to reflect that we are now caching from `docsets`

This PR has been updated for updating the Types as well to reflect the changes between`repos_branches` and `docsets`


mmeigs (approved):
❤️

mmeigs (approved):
Seems good to me! Thank you for the clean-up!

modules/persistence/src/services/metadata/repos_branches/index.ts
```diff
@@ -16,17 +16,23 @@ export interface BranchEntry {
...
   project: string;
   branches: BranchEntry[];
-  url: EnvKeyedObject;
-  prefix: EnvKeyedObject;
   internalOnly: boolean;
   [key: string]: any;
 }
 
-const internals: { [key: project]: ReposBranchesDocument } = {};
+export type ReposBranchesDocsetsDocument = ReposBranchesDocument & DocsetsDocument;
```
seungpark, line 33: what is this new document type?

IIUC it contains `url, prefix, bucket` but also `repoName, project, ...`

caesarbell: That is correct. The new document type is a combination of `repos_branches` and `docsets`. The following properties come from `docsets` now, which are `prefix`, `bucket` and `URL`.

seungpark: are we constructing a document with both of these properties? seeing that neither `repos_branches` or `docsets` has all of these properties defined. think we should be defining `ReposBranchesDocument` and `DocsetsDocument` separately, not combining them into one. (if connecting the two types, it should be that docsets have a `repos` property that is `ReposBranchesDocument[]` as in the data:

image

image

mmeigs: We're aggregating the two together to have a flat access to all of the properties!

PR #929, opened 2023-10-31
DOP-4079: Slack Deploy Monorepo

mmeigs:

PR #932, opened 2023-11-03
DOP-4079 follow-up - ensure monorepo jobs access correct Makefiles

mmeigs:
### Ticket

DOP-4079

### Notes

As a follow-up to [PR 929](https://github.com/mongodb/docs-worker-pool/pull/929), this alteration to the `meta` branch is necessary for successful monorepo `Job` builds. 

Both changes have to do with the now-double-layered monorepo paradigm (ie. `docs-monorepo/cloud-docs` vs `cloud-docs`).

1. The new PATH_TO_SNOOTY correctly points to the directory level where Snooty is installed. 
2. New publishedBranches files to account for the nested level of all requests. In the future, would love to not have duplicates such as these, but with our Makefiles being removed soon - it could just be a small necessary evil for now

### DIFF
--- a/makefiles/Makefile.cloud-docs
+++ b/makefiles/Makefile.cloud-docs
@@ -37,11 +37,11 @@ help:
 	@echo '  ARGS         - Arguments to pass to mut-publish'
 
 ifneq (,$(findstring docs-monorepo,$(REPO_DIR)))
-    # Found
-		RELATIVE_PATH_OF_PUB_BRANCHES=docs-monorepo/cloud-docs
+	# Found
+	RELATIVE_PATH_OF_PUB_BRANCHES=docs-monorepo/cloud-docs
 else
-    # Not found
-		RELATIVE_PATH_OF_PUB_BRANCHES=cloud-docs
+	# Not found
+	RELATIVE_PATH_OF_PUB_BRANCHES=cloud-docs
 endif
 
 get-build-dependencies:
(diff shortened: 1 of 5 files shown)


makefiles/Makefile.cloud-docs
```diff
@@ -38,4 +38,4 @@ help:
 
 
 get-build-dependencies:
-	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/cloud-docs.yaml > ${REPO_DIR}/published-branches.yaml
+	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/{REPO_DIR}.yaml > ${REPO_DIR}/published-branches.yaml
```
branberry, line 41: I don't know if we can use `REPO_DIR` as a part of the URL, or at least we need to slice it to get the part of the path containing the directory we want. When you run `pwd`, it gives you the full path to the directory:
image

makefiles/Makefile.docs-landing
```diff
@@ -18,7 +18,7 @@ DOCS_TOOLS_THEME=docs-tools/themes/mongodb/static
 
 
 get-build-dependencies:
-	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-landing.yaml > ${REPO_DIR}/published-branches.yaml
+	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/{REPO_DIR}.yaml > ${REPO_DIR}/published-branches.yaml
```
branberry, line 21: Same thing here for this one as well

branberry (approved):
LGTM!

PR #936, opened 2023-11-16
DOP-3960: use dev endpoint to fetch versioned API spec in staging

anabellabuckvar:
### Ticket
[DOP-3960](https://jira.mongodb.org/browse/DOP-3960)

Fetches the versioned API specs using the equivalent pre-production endpoints for the specs. 

### Production
[api spec v2](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/)

### Notes
There are both Dev and QA endpoints for the API spec. I've initially implemented the changes with the Dev url because the  the goal of the ticket says to "use dev endpoint to fetch version API spec in staging".  However, this can easily be changed if desired


modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -7,13 +7,23 @@ import { findLastSavedVersionData, saveSuccessfulBuildVersionData } from './data
 import { VersionData } from './models/OASFile';
 import { normalizeUrl } from '../utils/normalizeUrl';
 
-const OAS_FILE_SERVER = 'https://mongodb-mms-prod-build-server.s3.amazonaws.com/openapi/';
-const GIT_HASH_URL = 'https://cloud.mongodb.com/version';
+const env = process.env.SNOOTY_ENV ?? '';
+
+console.log(env);
```
rayangler, line 12: Are the console logs in this file still needed?

modules/oas-page-builder/src/services/pageBuilder.ts
```diff
@@ -274,7 +284,7 @@ export const buildOpenAPIPages = async (
     if (totalSuccess && sourceType == 'atlas') {
       try {
         const gitHash = await fetchGitHash();
-        const versions = await fetchVersionData(gitHash);
+        const versions = await fetchVersionData(gitHash, OAS_FILE_SERVER);
         await saveSuccessfulBuildVersionData(source, gitHash, versions);
```
rayangler, line 288: `saveSuccessfulBuildVersionData` currently saves everything to either the `dotcomprd` db or the `dotcomstg` ([mapping code](https://github.com/mongodb/docs-worker-pool/blob/37ef215c1c088a0558d9fa97fdde8907856782a3/modules/oas-page-builder/src/services/database.ts#L30)) since we used to always pull in the latest/production-ready spec. If we're fetching non-prod specs, I think we should either save to the non-prod db. Alternatively, we could just not save at all for non-prod builds (might not be ideal if their dev endpoint also suffers from consistency issues like their prod endpoint once did).

(Tangentially, the old mapping code can probably be deprecated and simplified in favor of using `process.env.SNOOTY_DB_NAME`, similar to how the [persistence module currently does it](https://github.com/mongodb/docs-worker-pool/blob/37ef215c1c088a0558d9fa97fdde8907856782a3/modules/persistence/src/services/connector/index.ts#L31), but that's not necessarily important for this PR)

anabellabuckvar: I updated the mapping just so we can save to the non-prod db, let me know if that's what you had in mind

modules/oas-page-builder/src/services/database.ts
```diff
@@ -26,6 +26,7 @@ const getDbName = () => {
   switch (env) {
     // Autobuilder's prd env
     case 'production':
+      return 'snooty_prod';
     case 'dotcomprd':
       return 'snooty_dotcomprd';
     // Autobuilder's pre-prd env
```
rayangler, line 32: If we're gonna add the mapping for the `production` environment, we should also add it for the `staging` environment for completeness

rayangler (approved):

PR #938, opened 2023-11-20
DOP-4127: Monorepo builds without Makefiles

mmeigs:
### Ticket

DOP-4127

### Notes

The ticket was to build `docs-monorepo/cloud-docs` locally. 
This PR **actually** encompasses building `docs-monorepo/cloud-docs` and `docs-monorepo/docs-landing` sites with the local debugger and also with github pushes. Probably half of this code is also merged with @branberry 's wonderful local debugger build capabilities. A real-life team PR in the flesh.

Note: as of now `docs-monorepo/docs-mongodb-internal` does build in development pretty well, but there's a bug with the newly-merged "curled" build dependencies. This will probably be settled in a future ticket unless I can figure it out soon!

#### Examples of builds and build logs

`mmeigs/docs-monorepo/cloud-docs` [build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=6578bafc7a5cb5fef57254c0)
- gatsby preview [link](https://preview-mongodbmmeigsstg.gatsbyjs.io/cloud-docs/mm-forky/)

To show no regressions when building a standard repo:
`10gen/docs-node` [build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=6578b6677a5cb5fef570c4f6)
- gatsby preview [link](https://preview-mongodbmmeigsstg.gatsbyjs.io/node/mm-forky-nodey/)

#### Added Benchmarks back in to this process

Screenshot 2023-12-14 at 3 29 39 PM


cdk-infra/lib/constructs/worker/worker-env-construct.ts
```diff
@@ -50,7 +50,7 @@ export class WorkerEnvConstruct extends Construct {
     // doing this for the time being, but I think we don't need to necessarily retrieve this from ssm for feature branches, nor would we want to in that case
     const previewBuildEnabled = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/preview_build/enabled`);
     const featureFlagUpdatePages = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/update_pages`);
-    const featureFlagMonorepoPath = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/monorepo_path`);
+    const featureFlagMonorepoPath = StringParameter.valueFromLookup(this, `${ssmPrefix}/DOP-4127/flag/monorepo_path`);
```
branberry, line 53: NIT: We will want to change this back

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -7,29 +7,28 @@ import { BuildDependencies } from '../../../entities/job';
...
+) {
   const prodFileName = `${process.cwd()}/snooty/.env.production`;
 
   try {
     await writeFileAsync(
       prodFileName,
       `GATSBY_SITE=${projectName}
       GATSBY_MANIFEST_PATH=${repoDir}/bundle.zip
-      GATSBY_PARSER_USER=${process.env.USER}
+      GATSBY_PARSER_USER=${process.env.USER ?? `docsworker-xlarge`}
```
branberry, line 24: Super nit: Since we aren't doing any interpolation, we can use regular single quotes instead of backticks.

src/job/jobHandler.ts
```diff
@@ -554,8 +647,20 @@ export abstract class JobHandler {
       this._currJob._id,
       `* Starting Job with ID: ${this._currJob._id} and type: ${this._currJob.payload.jobType}`
     );
+    await this._logger.save(this._currJob._id, `* Starting Job with repo name: ${this._currJob.payload.repoName}`);
     try {
-      await this.build();
+      // TODO: MONOREPO feature flag needed here
+      if (
+        // process.env.FEATURE_FLAG_MONOREPO_PATH === 'true' &&
```
branberry, line 654: Do we still want to add in the feature flag here?

src/job/jobHandler.ts
```diff
@@ -504,12 +544,65 @@ export abstract class JobHandler {
     return await this.executeBuild();
   }
 
+  /* New build process without Makefiles */
+  @throwIfJobInterupted()
+  protected async build(): Promise<boolean> {
+    this.cleanup();
+    const job = this._currJob;
+    const logger = (msg: string) => this.logger.save(this.currJob._id, msg);
```
branberry, line 552: I talked to @casthewiz about the real-time logging, and he brought up a valid concern that we could potentially run into issues with the number of writes to the database. `this.logger.save` writes to the database every time, and so we might want to avoid this. 

Here's what I think we can do: First, call `logger` on the final result of the job i.e. we just log the resulting CLI output at once. It would look like this:

```ts

const logger = (output: CliCommandResponse) => {
  this._logger.save(this.currJob._id, parseResults.outputText);
  this._logger.save(this.currJob._id, parseResults.errorText);

}
const parseResults = await nextGenParse({ job });

```

We'll also want to maintain the real-time logging which you have added for sure! I'll address that specifically in a comment for the `executeCliCommand` function. It's a change that I should have added initially.

src/repositories/docsetsRepository.ts
```diff
@@ -52,9 +52,11 @@ export class DocsetsRepository extends BaseRepository {
     ];
   }
 
-  async getProjectByRepoName(repoName: string): Promise<any> {
+  async getProjectByRepoName(repoName: string, directory?: string): Promise<any> {
```
branberry, line 55: NIT: This is opening up a can of worms, and I know that this isn't your change specifically, but we should provide a different return type than `Promise<any>`. Specifically, let's replace the `any` with a more concrete type.

The reason why this is opening up a can of worms is because of how we type the collection, which is assigned a type annotation in the super class... That might be a bit of work, so don't feel like you need to get this changed for this PR.

src/job/stagingJobHandler.ts
```diff
@@ -75,10 +77,35 @@ export class StagingJobHandler extends JobHandler {
     }
   }
   async deploy(): Promise<CommandExecutorResponse> {
+    let resp;
     try {
-      const resp = await this.deployGeneric();
+      // TODO: MONOREPO feature flag needed
+      if (
+        // process.env.FEATURE_FLAG_MONOREPO_PATH === 'true' &&
+        this.currJob.payload.repoName === MONOREPO_NAME
```
branberry, line 85: Do we want to add the feature flag here?

src/job/jobHandler.ts
```diff
@@ -504,12 +544,65 @@ export abstract class JobHandler {
...
+
+    await this.cloneRepo(this._config.get<string>('repo_dir'));
+    this._logger.save(job._id, 'Cloned Repo');
+    await this.commitCheck();
+    this._logger.save(job._id, 'Checked Commit');
+    await this.pullRepo();
+    this._logger.save(job._id, 'Pulled Repo');
+    await this.setEnvironmentVariables();
+    this.logger.save(job._id, 'Prepared Environment variables');
+    const buildDependencies = await this.getBuildDependencies();
```
branberry, line 562: Since we aren't using the makefiles, we don't need to do everything here and before.

src/commands/src/shared/next-gen-stage.ts
```diff
@@ -1,46 +1,76 @@
...
-  if (patchId && projectName === mutPrefix) {
-    prefix = `${commitHash}/${patchId}/${mutPrefix}`;
-    hostedAtUrl = `${url}/${commitHash}/${patchId}/${mutPrefix}/${DOCS_WORKER_USER}/${commitBranch}/`;
+  if (patch && newHead && project === mutPrefix) {
+    prefix = `${newHead}/${patch}/${mutPrefix}`;
+    hostedAtUrl = `${url}/${newHead}/${patch}/${mutPrefix}/${DOCS_WORKER_USER}/${branchName}/`;
   }
 
-  commandArgs.push(`--prefix="${prefix}"`);
+  commandArgs.push(`--prefix=${prefix}`);
```
branberry, line 42: Good catch on this one!

src/commands/src/helpers/index.ts
```diff
@@ -166,24 +168,29 @@ export async function executeCliCommand({
 
     if (writeStream) executedCommand.stdout?.pipe(writeStream);
     executedCommand.stdout?.on('data', (data: Buffer) => {
+      if (logger) logger(data.toString());
```
branberry, line 171: This is a nice addition! I mentioned in a previous comment that I would describe how we would change this. Instead of using the `logger` function, let's just use a plain old `console.log` so that we aren't writing to the database. We won't get the real-time result from the dashboard, but in AWS CloudWatch, the `console.log` statements will show up in real time.

src/commands/src/scripts/local-build/utils/get-args.ts
```diff
@@ -0,0 +1,54 @@
+export function getArgs() {
+  const helpIdx = process.argv.findIndex((str) => str === '--help' || str === '-h');
+
+  if (helpIdx !== -1) {
+    console.log('TODO: Fill this out');
```
branberry, line 5: I'll update this

src/commands/src/scripts/local-build/utils/create-job.ts
```diff
@@ -0,0 +1,52 @@
...
+interface Props {
+  branchName: string;
+  repoOwner: string;
+  repoName: string;
+  commit: CommitGetResponse;
+  project: string;
+  directory?: string;
+}
+
+export function createLocalJob({
```
branberry, line 13: I'll add a comment for this

api/controllers/v2/github.ts
```diff
@@ -110,15 +112,16 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
 
     consoleLogger.info(job.title, 'Creating Job');
     const jobId = await jobRepository.insertJob(job, c.get('jobsQueueUrl'));
     jobRepository.notify(jobId, c.get('jobUpdatesQueueUrl'), JobStatus.inQueue, 0);
     consoleLogger.info(job.title, `Created Job ${jobId}`);
   }
 
-  if (process.env.FEATURE_FLAG_MONOREPO_PATH === 'true' && body.repository.name === MONOREPO_NAME) {
+  if (body.repository.name === MONOREPO_NAME) {
+    // TODO: MONOREPO feature flag needed here
```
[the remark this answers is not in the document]
mmeigs: Ah, I should have added in the PR notes! I still need to add a check for the feature flag in all these conditionals! The reason I haven't is because we never figured out how with the feature branch. But also... I guess it does seem redundant? I guess the only benefit would be that no one could accidentally release a monorepo site.

src/job/jobHandler.ts
mmeigs: As of now, yes!

src/commands/src/helpers/index.ts
mmeigs: Alright, so the logs will be very short and possibly not have much logging to show if there's been an issue with a "successful" command output. Do we think that's worth the risk? It's the only way we caught multiple issues in this process, honestly

src/repositories/docsetsRepository.ts
mmeigs: Great callout!

api/controllers/v2/github.ts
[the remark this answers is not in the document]
branberry: I think the reason why it doesn't work is because of the [line here](https://github.com/mongodb/docs-worker-pool/blob/b57f0ac0af793b14db7bdf20a2496a1d07bb73be/cdk-infra/lib/constructs/api/webhook-env-construct.ts#L53C69-L53C69). If we remove `.stringValue`, then it should be defined. 

But, if you do find it redundant, I am more than fine with removing it entirely.

src/repositories/docsetsRepository.ts
branberry: Just checked, and I believe that is correct!

src/job/jobHandler.ts
branberry: Sorry, I was getting all mixed up. I was working on another branch where I added the clone functionality to the [`getBuildDependencies` function](https://github.com/mongodb/docs-worker-pool/blob/32ae0a7c709f8a3071b4cec36c20c1a020acbaa1/src/commands/src/helpers/dependency-helpers.ts#L18). 

Looking at the commitCheck more closely, it looks like we do need something to do that logic if we were to remove it. It looks like it [checks out the branch and make sure the commit exists](https://github.com/mongodb/docs-worker-pool/blob/040d812d52b4be5bfcd978f6be7a9efc78fed76b/src/services/commandExecutor.ts#L92).

Now that I've thought about it a bit more, I'll leave it up to you if you want to keep this!

src/job/stagingJobHandler.ts
```diff
@@ -75,10 +77,35 @@ export class StagingJobHandler extends JobHandler {
...
+        }
+
+        const { bucket, url } = await this.getEnvironmentVariables();
+        resp = await nextGenStage({ job: this.currJob, logger, bucket, url });
+      } else {
+        resp = await this.deployGeneric();
+      }
       const summary = '';
-      if (resp?.output?.includes('Summary')) {
+      if (resp.output?.includes('Summary')) {
```
branberry, line 108: the response object from `nextGenStage` doesn't have the `output` property. Instead, it has `outputText` and `errorText`.

src/commands/src/helpers/index.ts
branberry: We can still log the result after it has finished completely. For example:

```ts
  const result = await executeCliCommand({});
  
  this.logger.save(..., result.outputText);
  this.logger.save(..., result.errorText);

```

For this, we are looking to avoid calling `logger.save` which uploads the logs to Atlas every time. The `stdout.on('data')` `stderr.on('data')` get called frequently, so this would lead to us uploading to Atlas much more often.

src/job/stagingJobHandler.ts
mmeigs: It does! I've modified `nextGenStage` to have the same response structure that `deployGeneric` does! We can always change this, though!

branberry: Oh, sorry! I missed that. It's totally fine to keep it how it is. One thing I've noticed is that sometimes logs will be sent to `stderr` meaning that it will be in the `errorText` for some reason. I don't think it's an issue here, but if we don't see logs here at some point, that could be why.

src/entrypoints/localApp.ts
```diff
@@ -7,14 +7,61 @@ import {
...
+  action: '',
+  branchName: 'master',
+  isFork: false,
+  repoOwner: '10gen',
+  url: 'https://github.com/10gen/cloud-docs', // ???
+  mutPrefix: 'docs-qa/atlas/mmeigs-build',
+  prefix: '',
+  includeInGlobalSearch: true,
+} as Payload;
+// Bucket??
```
rayangler, line 27: Just curious what the question marks are for

src/entrypoints/localApp.ts
```diff
@@ -41,61 +88,49 @@ async function localApp() {
 
   const { commitHash, patchId, bundlePath, commitBranch, hasRedirects, repoDir } = await prepareBuildAndGetDependencies(
```
rayangler, line 89: Looks like with the recent changes, some of these fields are no longer used. We should consider cleaning these up to avoid any confusion in usage

rayangler (commented):
Thanks for working on this! A lot of my comments are just questions for my own understanding 🤠

Dockerfile.local
```diff
@@ -39,63 +41,73 @@ RUN git clone -b v${MUT_VERSION} --depth 1 https://github.com/mongodb/mut.git \
     && make package \
     && mv dist/mut /opt/
 
-RUN curl  -L -o redoc.zip https://github.com/mongodb-forks/redoc/archive/refs/tags/v${REDOC_CLI_VERSION}.zip \
```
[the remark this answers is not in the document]
mmeigs: I do want to try again today, but I was having the same issues locally with it even after using the rc. However, it works within our regular pipeline!

src/clients/githubClient.ts
```diff
@@ -2,11 +2,11 @@ import { Octokit } from '@octokit/rest';
 
 let client: Octokit;
 
-export function getOctokitClient(): Octokit {
+export function getOctokitClient(password?: string): Octokit {
   if (client) return client;
 
   try {
-    const { GITHUB_BOT_PASSWORD } = process.env;
+    const GITHUB_BOT_PASSWORD = process.env.GITHUB_BOT_PASSWORD ?? password;
```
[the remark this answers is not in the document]
mmeigs: @branberry I feel like this is something you figured out!

src/commands/src/helpers/index.ts
```diff
@@ -174,16 +174,17 @@ export async function executeCliCommand({
...
     });
 
     executedCommand.on('close', (exitCode) => {
       if (writeStream) writeStream.end();
 
       if (exitCode !== 0) {
-        console.error(`ERROR! The command ${command} closed with an exit code other than 0: ${exitCode}.`);
-        console.error('Arguments provided: ', args);
-        console.error('Options provided: ', options);
+        console.log(`ERROR! The command ${command} closed with an exit code other than 0: ${exitCode}.`);
```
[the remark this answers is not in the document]
mmeigs: Ah, great catch. This is an error when I reverted in my part. I originally was using the IJobLogger, but we economized how many times we wrote to the database. I changed back, but I did it from memory. I'll change this back to stderr!

src/commands/src/scripts/local-build/utils/get-args.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    process.exit(0);
+  }
+  const repoOwnerIdx = process.argv.findIndex((str) => str === '--repo-owner' || str === '-o');
+  const repoNameIdx = process.argv.findIndex((str) => str === '--repo-name' || str === '-n');
+  const directoryIdx = process.argv.findIndex((str) => str === '--directory' || str === '-d');
```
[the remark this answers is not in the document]
mmeigs: Absolutely. I'll move it below the comment as well

src/commands/src/shared/next-gen-deploy.ts
```diff
@@ -1,38 +1,76 @@
...
   hasConfigRedirects: boolean;
-  url: string;
 }
 
-export async function nextGenDeploy({ bucket, mutPrefix, gitBranch, hasConfigRedirects, url }: NextGenDeployParams) {
-  if (hasConfigRedirects && (gitBranch === 'main' || gitBranch === 'master')) {
-    // equivalent to: mut-redirects config/redirects -o public/.htaccess
-    await executeCliCommand({ command: 'mut-redirects', args: ['config/redirects', '-o', 'public/.htaccess'] });
-  }
+/* This is still in development - use with caution */
```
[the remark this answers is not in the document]
mmeigs: The worry was more that all the functions replacing Makefile commands have changed quite a bit without really working on this one - so most likely out of date and possibly buggy. But I think that's a great suggestion. Will add!

src/commands/src/shared/next-gen-deploy.ts
```diff
@@ -1,38 +1,76 @@
...
+    console.log(`${outputText}\n Hosted at ${url}/${mutPrefix}`);
+    return {
+      status: 'inProgress',
+      output: outputText,
+      error: '',
+    };
+  } catch (error) {
+    console.log(`nextGenDeploy has failed.`);
+    return {
+      status: 'failure',
```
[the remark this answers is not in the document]
mmeigs: So glad you saw this. I remember having an error (which had nothing to do with this) that had "Invalid Status" and I was sure this just had the wrong value. 

But [here is where](https://github.com/mongodb/docs-worker-pool/blob/c15b72c82c6b7ddab666fbb351132f0011c70dd4/src/services/commandExecutor.ts#L60) the "strings" are defined. We/I should change the typescript from `string` to "success"|"failed".

This is great to realize because some places I put "failed" and some I put "failure". Sloppy. Thank you for seeing.

src/entrypoints/localApp.ts
mmeigs: oh god. I swear I went through multiple times to clean up. This is v old. Deleting.

src/clients/githubClient.ts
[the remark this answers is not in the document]
branberry: @rayangler Pretty much! When creating a local job, the data is pulled from the GitHub commit so we need to provide the Octokit password to clone private repos. [Here](https://github.com/mongodb/docs-worker-pool/blob/93d2642ebaef5039e0337c6d1c3eb01a573e5a0e/src/commands/src/scripts/local-build/index.ts#L49) is where the change is needed.

src/commands/src/scripts/local-build/utils/get-env-vars.ts
```diff
@@ -0,0 +1,174 @@
+import { SSMClient, GetParameterCommand } from '@aws-sdk/client-ssm';
+
+// This array contains the Parameter Store paths of the secure strings
+// we want to add to the worker environment. These are mapped
+// to their environment variable name in the workerParamPathToEnvName map.
```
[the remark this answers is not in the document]
branberry: This is correct. I think that's a great idea, but I am not sure of the best way to go about it since the `cdk-infra` repository is separate from the Autobuilder code. I have had issues trying to import stuff outside of `cdk-infra`, but maybe it would be okay for the Autobuilder to import functionality from `cdk-infra`? This is something worth discussing I think with the team more generally of how we want to share stuff in a monorepo environment. 

I've also had this question with respect to types that are shared throughout different parts of the project.

src/commands/src/scripts/local-build/index.ts
```diff
@@ -0,0 +1,85 @@
...
+      'X-GitHub-Api-Version': '2022-11-28',
+    },
+  });
+
+  const buildPromise = buildDockerImage(env.NPM_BASE_64_AUTH);
+  const dbClient = new MongoClient(env.MONGO_ATLAS_URL);
+
+  console.log('Building docker container....');
+  const [commit, connectedDbClient, buildResult] = await Promise.all([commitPromise, dbClient.connect(), buildPromise]);
+  console.log(buildResult.errorText);
```
branberry, line 65: I know I added this in, but I will remove these extra logs since we log in real time now in `executeCliCommand`

src/commands/src/shared/next-gen-deploy.ts
```diff
@@ -1,38 +1,76 @@
...
-        `--prefix=${mutPrefix}`,
-        '--deploy',
-        `--deployed-url-prefix=${url}`,
-        '--json',
-        '--all-subdirectories',
-      ],
-      options: {
-        cwd: `${process.cwd()}/snooty`,
-      },
+    const bucket = process.env.BUCKET;
```
branberry, line 19: NIT: I think we should pass these values in as parameters instead of using them from the environment variable. These get set in `setupEnvironmentVariables` as a way to pass information into the makefiles, but since these values are a part of the `Job` object, we can get them from there. That way, we know where these values are coming from a bit better, and can do one less mental hoop jump to figure out how they get defined.

src/commands/src/shared/next-gen-deploy.ts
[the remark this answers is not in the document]
branberry: They are also defined as an enum [here](https://github.com/mongodb/docs-worker-pool/blob/01a9d71706d4b0dcc08654278e61b83c53c7180e/src/entities/job.ts#L4) if you want to reuse it. They have string values, so you could do something like `const status: JobStatus = 'failed'` and it should work (I think)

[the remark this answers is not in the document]
mmeigs: Well, these are not JobStatuses sadly, these are... CommandExecutorResponse statuses lol

src/commands/src/scripts/local-build/utils/get-env-vars.ts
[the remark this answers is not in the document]
rayangler: Oh, gotcha. I wasn't aware that importing stuff from outside of `cdk-infra` was an issue. Is that just a limitation of how CDK works/builds things?

Also wanted to note that this comment isn't necessarily a blocker for this PR. It's just a quality-of-life concern for us. So if there isn't an easy way to resolve this immediately, that's okay

src/commands/src/shared/next-gen-deploy.ts
mmeigs: I don't believe they are on the `Job`, but I do think that they could definitely be parameters for this. That's a good idea.

branberry (commented):
Hi Matt! Looking really good, and almost ready for an approval! Just left some more comments on various, small stuff.

src/commands/src/shared/next-gen-parse.ts
```diff
@@ -1,33 +1,45 @@
...
   isProd?: boolean;
 }
-export async function nextGenParse({
-  repoDir,
-  patchId,
-  commitHash,
-  isProd,
-}: NextGenParseParams): Promise<CliCommandResponse> {
-  const commandArgs = ['build', repoDir, '--output', `${repoDir}/bundle.zip`, RSTSPEC_FLAG];
+export async function nextGenParse({ job, isProd }: NextGenParseParams): Promise<CliCommandResponse> {
```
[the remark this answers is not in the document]
branberry: I was looking at it from the perspective of each individual function more. For example, `nextGenParse` only uses `job.payload.newHead` and `job.payload.patch`, but `job` has many more properties than that. If we were to test this function, mocking out a fake `job` object would be a bit tedious.

With that said, I think you have a valid point that the `jobHandler` would then be much larger having to break out the individual properties that we want to pass to each function, which definitely makes it less readable from that perspective. And like you mentioned, some of the other functions have many more properties that are used from the `job` object. 

To be honest, I am totally fine keeping it as is since either way there are some trade-offs from which perspective you are looking at. Thanks for the thoughtful response!

src/commands/src/scripts/local-build/utils/get-env-vars.ts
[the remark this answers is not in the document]
branberry: I think it's more of a limitation of a TypeScript/node project's ability to import stuff outside of the given root project directory. I guess though, we should have the ability to grab these functions from `/cdk-infra` since it is a part of the `docs-worker-pool` root directory. 

To be honest, this was done in an expedient fashion haha. I wanted to get it working, but knew this was less than ideal essentially copying and pasting stuff. I think it's worth flagging this though and creating a ticket to refactor this. One common pain point lately has been having to do duplicate work, and this definitely introduces more of that.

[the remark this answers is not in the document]
rayangler: Makes sense. Thanks!

src/commands/src/shared/next-gen-parse.ts
```diff
@@ -1,33 +1,45 @@
...
+  const repoDir = path.resolve(process.cwd(), `repos/${getDirectory(job)}`);
+  const commitHash = job.payload.newHead;
+  const patchId = job.payload.patch;
```
branberry, line 14: Hey Matt! I just realized that this `job.payload.patch` is the content for the patch file and not the patch ID. The patch ID from the makefile is retrieved [here](https://github.com/mongodb/docs-worker-pool/blob/a720d98b7e58418a92c06f7b4d346e3feacd7fd8/makefiles/shared.mk#L12). It's a bit odd, but we also get the commit hash [here](https://github.com/mongodb/docs-worker-pool/blob/a720d98b7e58418a92c06f7b4d346e3feacd7fd8/makefiles/shared.mk#L1) as well, but I think using the `job.payload.newHead` value is fine since it should be the same as what we get in the makefile. 

Since we aren't using the makefiles, the patch ID is being retrieved from the[prepareBuildAndGetDependencies function](https://github.com/mongodb/docs-worker-pool/blob/d1367a7702a97d2ed26b3e0d1ede9ac7c1cb2248/src/commands/src/helpers/dependency-helpers.ts#L96]. 

Not sure how much this all really matters since I'm assuming only a small number of builds use the patch ID (I'm assuming.

@rayangler @casthewiz, thoughts on this? In what situations do we use a patch? And is it something we may be able to forgo entirely?

rayangler: Iirc, patches were used primarily for pushless staging builds. Pushless staging isn't really used by a lot of people, and seems like it hasn't been regularly used since August 2023. I think we've discussed fully removing support for pushless staging in the future, but not sure when we'll commit to that.

src/commands/src/shared/next-gen-stage.ts
```diff
@@ -1,46 +1,74 @@
...
-  bucket,
-  url,
-  patchId,
-  commitBranch,
-  commitHash,
-}: StageParams) {
-  let hostedAtUrl = `${url}/${mutPrefix}/${DOCS_WORKER_USER}/${commitBranch}/`;
-  let prefix = mutPrefix;
+export async function nextGenStage({ job, bucket, url }: StageParams): Promise<CommandExecutorResponse> {
+  const { mutPrefix, branchName, patch, project, newHead } = job.payload;
```
branberry, line 13: Just like in the [`next-gen-parse.ts` file](https://github.com/mongodb/docs-worker-pool/pull/938#discussion_r1431646872), we will want to use the `patchId` value that we get when we call the `prepareBuildAndGetDependencies` function as opposed to the `patch` property from `job.payload`.

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -22,14 +15,14 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
       prodFileName,
       `GATSBY_SITE=${projectName}
       GATSBY_MANIFEST_PATH=${repoDir}/bundle.zip
-      GATSBY_PARSER_USER=${process.env.USER}
+      GATSBY_PARSER_USER=${process.env.USER ?? 'docsworker-xlarge'}
```
branberry, line 18: This is something that I forgot to do when writing this, but we will need to conditionally include the properties `PATCH_ID` and `COMMIT_HASH` if `patchId` exists so that it matches what the [makefiles are doing](https://github.com/mongodb/docs-worker-pool/blob/12a321a5e8163074c463a79502ff07b930e7f8e9/makefiles/shared.mk#L78). 

We can provide these values after we get those dependencies, but we will have to move [the function call](https://github.com/mongodb/docs-worker-pool/blob/ffe258c96da8516743c5844fcadd209da194715e/src/commands/src/helpers/dependency-helpers.ts#L87) outside of the array so that we have `patchId` and `commitHash` as well.

For context, the `nextGenHtml` function depends on these values.

src/job/stagingJobHandler.ts
```diff
@@ -75,10 +77,30 @@ export class StagingJobHandler extends JobHandler {
...
+          const errorMessage = `
+            Docset Repo data not found in Atlas for repoName: ${this.currJob.payload.repoName}\n
+            ${this.currJob.payload.directory ? `directory: ${this.currJob.payload.directory}\n` : ''}
+            project: ${this.currJob.payload.project}
+          `;
+          throw Error(errorMessage);
+        }
+
+        const { bucket, url } = await this.getEnvironmentVariables();
+        resp = await nextGenStage({ job: this.currJob, bucket, url });
```
branberry, line 97: NIT: It appears that we no longer throw an error in `nextGenStage`, we will never log to `stdErr` [here](https://github.com/mongodb/docs-worker-pool/blob/ffe258c96da8516743c5844fcadd209da194715e/src/job/stagingJobHandler.ts#L110C28-L110C28). We can either add throw statements back to `nextGenStage` or check the result of the the response status to see if it has failed.

src/job/jobHandler.ts
```diff
@@ -504,12 +564,84 @@ export abstract class JobHandler {
...
+      job.payload.project,
+      baseUrl,
+      buildDependencies,
+      job.payload.directory
+    );
+    this._logger.save(this._currJob._id, 'Downloaded Build dependencies');
+
+    let buildStepOutput: CliCommandResponse;
+
+    const parseFunc = async () => nextGenParse({ job, patchId });
```
branberry, line 604: It appears that the new functions we wrote (e.g. `nextGenParse`, `nextGenHtml`, etc.)  don't throw errors, and we don't check if a step has failed. Because of this, I would assume that the builds will call all of the steps, even if previous ones fail. 

This isn't a huge issue, but I believe we should stop calling subsequent commands if previous ones fail. One thing that this could lead to is incorrect benchmark values as we still save them even if there is an error.

src/commands/src/shared/next-gen-stage.ts
mmeigs: Great call out!

src/job/jobHandler.ts
```diff
@@ -297,13 +297,20 @@ export abstract class JobHandler {
   // call this method when we want benchmarks and uses cwd option to call command outside of a one liner.
   private async wrapWithBenchmarks(
     buildStepFunc: () => Promise<CliCommandResponse>,
-    stage: string
+    stage: string,
+    logger: IJobRepoLogger
```
branberry, line 301: NIT: We don't need the logger parameter as we can access it via `this.logger` in this method.

src/job/jobHandler.ts
```diff
@@ -297,13 +297,20 @@ export abstract class JobHandler {
...
   ): Promise<CliCommandResponse> {
     const start = performance.now();
 
     const resp = await buildStepFunc();
 
     const end = performance.now();
+
+    if (resp.errorText?.length > 0) {
```
branberry, line 309: I think this may not be a reliable check, unfortunately. 

In some instances, `errorText` will be populated even if the command doesn't actually fail. For example, the `snooty build` command writes its output to `stderr` which is what `errorText` contains. Because of this, we may throw even if the build is successful. 

I think we can either modify the commands to include the `status` property, or have them throw the error.

branberry (commented):
Hi Matt! Thank you for the last few adjustments. Just have a few more comments. Sorry to keep piling stuff on, but I believe we are really close to the finish line with this, thank you for all of your really quick responses!! I'll make sure to keep an eye on this.

src/job/jobHandler.ts
mmeigs: Great point. I'm so used to the new functions that I was not thinking straight. Thanks!

src/job/jobHandler.ts
mmeigs: Ah, I did this instead because we still need them to log an error so we know why any command failed. This at least let us not pipe through the logger, as you disliked having that throughout! So, we can't throw unless we log first. 

I think I'm confused over when the errorText would be filled. I've created multiple successful builds since this change and everything performed as expected. Can you give an example so I can test?

branberry (approved):
Hey Matt! Looks amazing. Did another pass through, and it looks good to me. Feel free to take your time double checking of course if you'd like before you merge. Also, I would feel comfortable with at least one other person approving.

Great work!!

Dockerfile.local
```diff
@@ -39,63 +41,82 @@ RUN git clone -b v${MUT_VERSION} --depth 1 https://github.com/mongodb/mut.git \
...
+COPY --chown=docsworker-xlarge modules/oas-page-builder/index.ts ./modules/oas-page-builder
 
 RUN cd ./modules/oas-page-builder \
     && npm run build
 
+COPY --chown=docsworker-xlarge ./modules/oas-page-builder/dist/ ./modules/oas-page-builder
+
+FROM oas as root
+
+COPY --from=persistence --chown=docsworker-xlarge ${WORK_DIRECTORY}/modules/persistence/ ./modules/persistence
```
mmeigs, line 99: @branberry Just making a last-minute check because of our new `COPY` knowledge. Is it ok that we have six `COPY` statements without using `./` at the beginning of the first path? Also, there's a seventh (in a way) on line 99 because `WORK_DIRECTORY` does not begin with a `.`

rayangler: For what it's worth, I think the one on line 99 is pretty safe since `--from` would explicitly ensure we're copying from the intended source. If we wanted to play it safe, we can try using `--from` on the other copy commands. We do so on other Dockerfiles, but not sure if the use case is different?

rayangler (approved):
Things generally lgtm

Dockerfile.local
rayangler: Oh wait, nvm, I think `--from` would only be applicable for previous build steps, not if you're copying from local source. Feel free to ignore!

Dockerfile.local
```diff
@@ -39,63 +41,82 @@ RUN git clone -b v${MUT_VERSION} --depth 1 https://github.com/mongodb/mut.git \
...
+COPY --chown=docsworker-xlarge ./modules/persistence/dist/ ./modules/persistence/
```
rayangler, line 100: Oh, I'm actually receiving an error from this line when running locally: 

```
Dockerfile.local:100
--------------------
  98 |     
  99 |     COPY --from=persistence --chown=docsworker-xlarge ${WORK_DIRECTORY}/modules/persistence/ ./modules/persistence
 100 | >>> COPY --chown=docsworker-xlarge ./modules/persistence/dist/ ./modules/persistence/
 101 |     
 102 |     # Root project build
--------------------
ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref c9d5839d-a372-45d3-aab0-f686e004deb8::jjcvfvfz7bc2nrmjzqquqor02: "/modules/persistence/dist": not found

View build details: docker-desktop://dashboard/build/desktop-linux/desktop-linux/skhd7adzlmqayhfural4uxgq8
```

Is the expectation here that we build the OAS module locally ourselves first?

mmeigs: Hmm, but we build it on line 78...
@branberry Do we need line 99 anymore? Isn't line 100 doing "the needful" in this situation? Maybe 99 is actually a barrier for 100?

rayangler: Yeah, I'm a bit confused. Are you running into this issue locally? @mmeigs

mmeigs: Just started running it locally a couple minutes ago to check! Sadly, it's been a second, so this will be a long cold build first.

mmeigs: @rayangler very odd. Mine is failing about four lines later... oof. This was working yesterday. Great catch.

mmeigs: @branberry I just commented out line 99 and it seems to be proceeding well!

PR #935, opened 2023-11-15
[DOP-4097]: Add NO_CACHING flag for shared.mk

branberry:

PR #940, opened 2023-11-21
[DOP-4097]: Ensure we use the --no-caching flag for production builds

branberry:
### Ticket

[DOP-4097](https://jira.mongodb.org/browse/DOP-4097)

### Notes

To verify, I have added the flag for staging jobs to ensure that the flag is passed into the makefile without error. Here is a [link](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=655cd5685b686d6bd9813840) to a job that has run with the `--no-caching` flag.

The shared.mk PR can be found [here.](https://github.com/mongodb/docs-worker-pool/pull/935/files)


src/commands/src/shared/next-gen-parse.ts
```diff
@@ -18,5 +23,9 @@ export async function nextGenParse({ repoDir, patchId, commitHash }: NextGenPars
     commandArgs.push(patchId);
   }
 
+  if (isProd) {
```
seungpark, line 26: IIUC, this is for when we do use this export fn `nextGenParse` for prod, since this is not being used yet. can we add a comment for this?

src/job/stagingJobHandler.ts
```diff
@@ -60,6 +60,7 @@ export class StagingJobHandler extends JobHandler {
 
   prepStageSpecificNextGenCommands(): void {
     if (this.currJob.buildCommands) {
+      // TODO: Remove no-caching after done testing
```
seungpark, line 63: minor, can remove!

seungpark (approved):
minor comments below! LGTM with the changes targeted for meta branch

rayangler (approved):

PR #946, opened 2023-12-01
DOP-4073: Adding support for curling dependencies stored in repos_branches

mayaraman19:
[DOP-4073](https://jira.mongodb.org/browse/DOP-4073)

This PR codifies makefile building dependencies in .ts files. It pulls the dependencies from `repos_branches` and curls them to a specific file/directory. The new field being added to `repos_branches` has the following structure:

```
optionalBuildSteps: {
     buildDependencies: [
          {
               targetDir: string, // optional, if not specified the target directory (directory where things are curled into) is REPO_DIR
               dependencies: [
                    {
                         url: string,
                         filename: string
                    },
               ]
          },
     ]
}
```

This is designed so as to ensure that we can match the output directory and filenames to what is currently in the makefiles, as well as to try to reduce duplication since many dependencies will be downloaded into the same directory.

Note: a subset of the `optionalBuildSteps` object has only been added to the `docs-mongodb-internal` entry in `repos_branches` for testing purposes right now. The logs should look like this:
image

To test, add an object of the above structure to a `repos_branches` document of your choice and populate it with some values from the corresponding [Makefile](https://github.com/mongodb/docs-worker-pool/tree/meta/makefiles). When testing the feature branch infrastructure, curl commands like the above should be printed in the logs.

(Also, I was kinda just figuring out this structure on my own - if anyone has suggestions for a better data structure or better way of formatting the logs I'm all ears👂)


src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -33,12 +34,42 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
   }
 }
 
-export async function prepareBuildAndGetDependencies(repoName: string, projectName: string, baseUrl: string) {
+export async function downloadBuildDependencies(buildDependencies: BuildDependencies | null, repoName: string) {
```
branberry, line 37: NIT: It might be simpler to read this if we invert the if statement conditional and return early:
```ts

if (!buildDependencies) return;

const commands: string[] = [];
buildDependencies.map((dependencyInfo) => {
    const repoDir = getRepoDir(repoName);
    const buildDir = dependencyInfo.buildDir ?? repoDir;
    executeCliCommand({
    command: 'mkdir',
    args: ['-p', buildDir],
    });
    commands.push(`mkdir -p ${buildDir}`);
    dependencyInfo.dependencies.map((dep) => {
    executeCliCommand({
        command: 'curl',
        args: ['-SfL', dep.url, '-o', `${buildDir}/${dep.filename}`],
    });
    commands.push(`curl -SfL ${dep.url} -o ${buildDir}/${dep.filename}`);
    });
});
return commands;
```

Or, it might be nice to make the `buildDependencies` object required, and check when we call the `downloadBuildDependencies` function that it is defined.

branberry (commented):
Hey Maya! Just a few small comments, but overall this looks awesome. Really good work!

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -33,12 +34,45 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
   }
 }
 
-export async function prepareBuildAndGetDependencies(repoName: string, projectName: string, baseUrl: string) {
+export async function downloadBuildDependencies(buildDependencies: BuildDependencies, repoName: string) {
+  const commands: string[] = [];
+  buildDependencies.map(async (dependencyInfo) => {
```
mmeigs, line 39: I believe with a map, this will return a Promise immediately, but the js thread won't actually await the completion of the promises. You could use Promise.all or allSettled to handle this asynchronicity!

src/repositories/repoBranchesRepository.ts
```diff
@@ -8,6 +8,15 @@ export class RepoBranchesRepository extends BaseRepository {
     super(config, logger, 'RepoBranchesRepository', db.collection(config.get('repoBranchesCollection')));
   }
 
+  async getBuildDependencies(repoName: string) {
```
mmeigs, line 11: nit: Could you add a return type to this just to keep us on the right path and because you already have those gorgeous types ready :)

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -33,12 +34,47 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
   }
 }
 
-export async function prepareBuildAndGetDependencies(repoName: string, projectName: string, baseUrl: string) {
+export async function downloadBuildDependencies(buildDependencies: BuildDependencies, repoName: string) {
+  const commands: string[] = [];
+  await Promise.all(
+    buildDependencies.map(async (dependencyInfo) => {
+      try {
```
branberry, line 41: NIT: I think we should narrow this try/catch to cover where in the code an exception might be thrown i.e. we can wrap just the `executeCliCommand` calls with a try catch. This allows us to be more fine-grained in our errors. For example:

```ts
try {
    await executeCliCommand({
       command: 'mkdir',
       args: ['-p', targetDir],
    });
} catch (e) {
  console.error(`Error! Could not create dependency directory: ${targetDir}. Dependency Information: `, dependencyInfo);
  throw e;

}
```

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -33,12 +34,47 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
...
+        const repoDir = getRepoDir(repoName);
+        const targetDir = dependencyInfo.targetDir ?? repoDir;
+        await executeCliCommand({
+          command: 'mkdir',
+          args: ['-p', targetDir],
+        });
+        commands.push(`mkdir -p ${targetDir}`);
+        dependencyInfo.dependencies.map((dep) => {
```
branberry, line 49: I think we will want to do an `await Promise.all()` here as well so that we ensure that these promises complete before continuing.

branberry (commented):
Hey Maya! Just a few more small comments!

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -33,12 +34,57 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
...
+      const repoDir = getRepoDir(repoName);
+      const targetDir = dependencyInfo.targetDir ?? repoDir;
+      try {
+        await executeCliCommand({
+          command: 'mkdir',
+          args: ['-p', targetDir],
+        });
+      } catch (error) {
+        console.error(
+          `ERROR! Could not create target directory ${targetDir}. Dependency information: ${dependencyInfo}`
```
branberry, line 50: NIT: Since `dependencyInfo` is an object, it'll display [Object object] when we log it like this. A simple trick that works will is to pass `dependencyInfo` as a second argument in `console.error`. That way, it handles formatting the object nicely for us.

```ts
console.error(`ERROR! Could not create target directory ${targetDir}. Dependency information: `,  dependencyInfo);
```

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -33,12 +34,57 @@ async function createEnvProdFile(repoDir: string, projectName: string, baseUrl:
...
+      await Promise.all(
+        dependencyInfo.dependencies.map((dep) => {
+          try {
+            executeCliCommand({
+              command: 'curl',
+              args: ['-SfL', dep.url, '-o', `${targetDir}/${dep.filename}`],
+            });
+          } catch (error) {
+            console.error(
+              `ERROR! Could not curl ${dep.url} into ${targetDir}/${dep.filename}. Dependency information: ${dependencyInfo}`
```
branberry, line 64: NIT: Same thing as mentioned above

PR #952, opened 2023-12-12
DOP-4201: Increase redundancy limit on metadata

casthewiz:
Increases our redundancy limit on `metadata` entries to be effectively 50, reducing likelihood that overlapping deploys of the same docset will impact one another.


seungpark (approved):
LGTM

PR #953, opened 2023-12-13
DOP-4177: Replacing job.useWithBenchmark with job.payload.isNextGen

mayaraman19:
[DOP-4177](https://jira.mongodb.org/browse/DOP-4177)

Removing use of `job.useWithBenchmark` and using `job.payload.isNextGen` instead to avoid relying on hardcoded array to determine if something is new gen or old gen. Also removing failing tests that were reliant on implementation and failed without use of`useWithBenchmark`.


mmeigs (commented):
If we're fully doing away with this, should we delete `useWithBenchmark` from our Job definitions and types?

mmeigs (approved):

PR #954, opened 2023-12-19
[DOP-4204]: Update README for local Autobuilder

branberry:
### Ticket 

[DOP-4204](https://jira.mongodb.org/browse/DOP-4204)


src/commands/src/scripts/local-build/utils/get-args.ts
```diff
@@ -2,7 +2,15 @@ export function getArgs() {
...
 
   if (helpIdx !== -1) {
-    console.log('TODO: Fill this out');
+    console.log(`
+      This command builds and runs the Autobuilder for local debugging.
+
+      Flags:
+      --repo-owner,  -o  (required)       The owner of the repo. Typically this is 'mongodb' or '10gen'. This should be your username for a fork.
+      --repo-name,   -n  (required)       The name of the repo e.g. docs-java or cloud-docs.
+      --directory,   -d  (optional)       The project directory path for a monorepo project.
```
mmeigs, line 11: Could you specify that this does not include a leading forward slash? 
ie the directory arg should be `cloud-docs` not `/cloud-docs`

src/commands/src/scripts/local-build/utils/get-args.ts
```diff
@@ -2,7 +2,15 @@ export function getArgs() {
...
+      --branch-name, -d  (optional)       The branch name we want to parse. If not provided, the value 'master' is used by default.
```
mmeigs, line 12: Typo: this needs to be `-b`

rayangler (commented):
I was planning on following along these steps, but I realized I don't have Docker set up yet on this current machine. 😓 

Either way, I think it looks good and is written well enough for someone to follow along 👍

rayangler (approved):

src/commands/src/scripts/local-build/utils/get-args.ts
mmeigs: Hmm, we could do this! It would fail right now if it had a leading slash because it is used throughout the pipeline in different ways, sometimes adding the slash, sometimes not.

PR #958, opened 2024-01-03
DOP-4209

mayaraman19:
[DOP-4209](https://jira.mongodb.org/browse/DOP-4209)

`docs-monorepo/docs-mongodb-internal` was not building without makefiles for multiple reasons:

- retrieving the build dependencies from `repos_branches` did not take into consideration monorepo naming/directory structure, causing builds to just be pulling from `docs-monorepo/cloud-docs` since that was the first monorepo entry to show up in `repos_branches`.
- migrating from makefiles did not take into account which error codes should/shouldn't cause the build to fail.

These have since been fixed, however, and `docs-monorepo` projects should be able to build now! (shoutout to @branberry for the help!)


cdk-infra/lib/constructs/worker/worker-env-construct.ts
```diff
@@ -50,7 +50,7 @@ export class WorkerEnvConstruct extends Construct {
     // doing this for the time being, but I think we don't need to necessarily retrieve this from ssm for feature branches, nor would we want to in that case
     const previewBuildEnabled = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/preview_build/enabled`);
     const featureFlagUpdatePages = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/update_pages`);
-    const featureFlagMonorepoPath = StringParameter.valueFromLookup(this, `${ssmPrefix}/flag/monorepo_path`);
+    const featureFlagMonorepoPath = StringParameter.valueFromLookup(this, `${ssmPrefix}/DOP-4127/flag/monorepo_path`);
```
mmeigs, line 53: Note to change back before merge

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -43,16 +43,25 @@ async function createEnvProdFile({
...
+export async function downloadBuildDependencies(
+  buildDependencies: BuildDependencies,
+  repoName: string,
+  directory?: string
+) {
   const commands: string[] = [];
   await Promise.all(
     buildDependencies.map(async (dependencyInfo) => {
-      const repoDir = getRepoDir(repoName);
+      const repoDir = directory ? getRepoDir(repoName, directory) : getRepoDir(repoName);
```
mmeigs, line 54: No need to do this conditional! The `directory` argument can be undefined. You can simple state:

`const repoDir = getRepoDir(repoName, directory);`

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -63,17 +72,19 @@ export async function downloadBuildDependencies(buildDependencies: BuildDependen
...
             });
           } catch (error) {
             console.error(
               `ERROR! Could not curl ${dep.url} into ${targetDir}/${dep.filename}. Dependency information: `,
               dependencyInfo
             );
+            return commands;
           }
           commands.push(`curl -SfL ${dep.url} -o ${targetDir}/${dep.filename}`);
         })
```
mmeigs, line 90: Hmm, interesting thing here. I could be wrong, but I ***believe*** that because we're returning a static value (even though it's a Promise because of the async keyword) it can return before the `executeCliCommand` function finishes. 

I think this can be slightly changed to do what you're trying to do and cleaner too. If you take away the `async` keyword and add `return` keyword to the `executeCliCommand` invocation that will ensure that the Promise.all waits for each Promise to finish. Then move the `commands.push` line up into the try block before the return. Then you can completely delete the unused line `return commands`. Doing this will ensure that the Node thread will not continue on until all of these commands have either finished or errored.

I have an example below.

Like this: 
```
await Promise.all(
     dependencyInfo.dependencies.map((dep) => {
          try {
            commands.push(`curl -SfL ${dep.url} -o ${targetDir}/${dep.filename}`);
            return executeCliCommand({
              command: 'curl',
               args: ['--max-time', '10', '-SfL', dep.url, '-o', `${targetDir}/${dep.filename}`],
              options: options,
            });
          } catch (error) {
            console.error(
              `ERROR! Could not curl ${dep.url} into ${targetDir}/${dep.filename}. Dependency information: `,
              dependencyInfo
            );
          }
       })
)
```

mmeigs (commented):
Okay, some of these comments are a little long, so please reach out if you have any questions!

src/commands/src/helpers/dependency-helpers.ts
branberry: These are all great points, @mmeigs. The one small nit with the example though is that we won't hit this catch block, I think. This is because we return the promise, but we don't await it, not giving the promise the opportunity to settle before the catch block can be entered. 

I think we can add the `async` back and do `return await executeCliCommand...` so that it will wait for the promise to settle before returning.

This also has me thinking as well, do we want to do an `allSettled` here maybe as well? That way, we can know all of the dependencies that are failing for whatever reason.

branberry:
We can ignore the failing `Update Feature Branch Infrastructure` jobs. This stems from an edge case with the new updated feature branch deploys that depend on being run as a part of the first job.

src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -63,36 +72,29 @@ export async function downloadBuildDependencies(buildDependencies: BuildDependen
       }
       commands.push(`mkdir -p ${targetDir}`);
       await Promise.all(
```
mayaraman19, line 74: Note: I did not use `allSettled` here because it seemed to not exist with this version of Javascript?

src/commands/src/helpers/dependency-helpers.ts
mayaraman19: Testing this out with some incorrect curl links but this approach still doesn't seem to catch the error 😞

mayaraman19:
I think I'm going to bump error logging in `downloadBuildDependencies` to another ticket because I can't figure it out 😭

branberry:
Hey all! For those reviewing, these build failures with the `Update Feature Branch Infrastructure` workflow are due to a change I recently introduced. I am currently working on a hot fix that should be out pretty soon here. Apologies for this!

src/commands/src/helpers/dependency-helpers.ts
branberry: What version of node are you using?

mayaraman19: v18.17.1!

src/commands/src/helpers/dependency-helpers.ts
branberry: Hmm, I wonder if doing `return await executeCliCommand(...)` we try:

```ts
 const cliResult = await executeCliCommand(...);

return cliResult;
```

in the try/catch? I wonder if return is sort of overwriting the `await` in a sense

mmeigs (approved):
Thanks for getting this working!

branberry (approved):
LGTM!

PR #965, opened 2024-01-18
DOP-4274: Label inactive branches in Slack Deploy Modal

anabellabuckvar:
### Context
It is very easy to accidentally deploy inactive versions in the Slack dropdown at /deploy - they need to be deployable to enable corrective edits or deploy-related platform functions, but there isn't visibility on which are inactive within the modal.

[DOP-4274](https://jira.mongodb.org/browse/DOP-4274)

### Staging
The changes were tested and staged in preprd. Inactive branch entries are labeled within the /deploy and /test_deploy UI.

Screenshot 2024-01-23 at 11 35 25 AM


api/handlers/slack.ts
```diff
@@ -24,14 +24,19 @@ export async function buildEntitledBranchList(entitlement: any, repoBranchesRepo
     const [repoOwner, repoName, directoryPath] = repo.split('/');
     const branches = await repoBranchesRepository.getRepoBranches(repoName, directoryPath);
     for (const branch of branches) {
-      let buildWithSnooty = true;
       if ('buildsWithSnooty' in branch) {
```
mmeigs, line 27: I know this is a holdover from the past, but I think we can get rid of this conditional. It doesn't gate anything and we're just checking true/false which undefined will work for as well.

api/handlers/slack.ts
```diff
@@ -24,14 +24,19 @@ export async function buildEntitledBranchList(entitlement: any, repoBranchesRepo
...
-          `${repoOwner}/${repoName}${directoryPath ? '/' + directoryPath : ''}/${branch['gitBranchName']}`
-        );
+        const buildWithSnooty = branch['buildsWithSnooty'];
+        if (buildWithSnooty) {
+          const active = branch['active'];
+          const repoPath = `${repoOwner}/${repoName}${directoryPath ? '/' + directoryPath : ''}/${
+            branch['gitBranchName']
+          }`;
+          if (!active) {
+            entitledBranches.push(`!Inactive ` + `${repoPath}`);
```
mmeigs, line 35: total nit: Because of the font-style of Slack, I think it might be more readable with a lower-case `i`

mmeigs: And up for debate - I'm wondering how to make it the most pleasing when we have no access to style. I wonder if it would look better in parentheses? 
`(!inactive) 10gen/docs/master`
wdyt?

mmeigs: Oh another nit: since you're already using template strings, you can get rid of the concatenation!

`(!inactive) ${repoPath}`

src/services/slack.ts
```diff
@@ -176,13 +176,19 @@ export class SlackConnector implements ISlackConnector {
...
+        valueBranchPath = fullPath.slice(10);
+      }
       const opt = {
         text: {
           type: 'plain_text',
-          text: fullBranchPath,
+          text: displayBranchPath,
         },
-        value: fullBranchPath,
+        value: valueBranchPath,
```
mmeigs, line 191: Maybe we can use a quick `startsWith` method to make this more legible. [Link to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith)

api/handlers/slack.ts
anabellabuckvar: I definitely think in parentheses would look better! I was following the ticket exactly but thats a great idea, thank you; I'll confirm in the channel

.github/workflows/deploy-stg-ecs.yml
```diff
@@ -3,6 +3,7 @@ on:
     branches:
       - "master"
       - "integration"
+      - "DOP-4274"
```
mmeigs, line 6: Just a reminder to remove this before merging

mmeigs (approved):

PR #966, opened 2024-01-18
DOP-4264

seungpark:
**Context:** noticed an unwanted outcome in persistence module when merging ToC's - A merged metadata entry for an associated product would have the full `url` option, instead of having the `slug` option to navigate to related pages. This was due to mutation of the originally referenced node, where it mutates previously stored values as well.

Example of entry here (see entry for _id: ObjectId('65a81d3ab914fa9f351aee3c') in snooty_dotcomstg.metadata)

image

The above ToC nodes should have a `slug` field (as opposed to `url` field) since this is a metadata entry for atlas-cli

To mitigate this issue, added a line to[ structuredClone ](https://developer.mozilla.org/en-US/docs/Web/API/structuredClone) that is available since node v17. Updated the git test and nvm versions accordingly


mmeigs (commented):
TIL `structuredClone`!

mmeigs (approved):

PR #967, opened 2024-01-19
DOP-2345: Add redirect rule for php library from master -> current

schmalliso:
FWIW, I've already YOLO done this in the AWS console 😁 😬


branberry (approved):
LGTM!

PR #969, opened 2024-01-24
DOP-3949: Rename 'master' to 'main'

i80and:
Clarity on how to test this welcome!


rayangler (approved):
I think this looks good to me given that it's just a name change for the branches

PR #970, opened 2024-01-25
[DOP-4171]: Create API Gateway endpoint and Lambda to handle snooty parser cache updates

branberry:
## TIcket

[DOP-4171](https://jira.mongodb.org/browse/DOP-4171)

## Testing 

[Verification of a successful run](https://github.com/mongodb/docs-worker-pool/actions/runs/7673143846/job/20915504108)
## Notes

[Spike document for context](https://docs.google.com/document/d/1ZJY8deWAqmAVrsvsb1emImerhqOT8q8EM4UJaPhCyWU/edit#heading=h.wyrxqhxvy787)

This PR implements the functionality to handle requests to update the `snooty-parse-cache` S3 bucket.

This ticket is dependent on [DOP-4294](https://github.com/mongodb/docs-worker-actions/pull/71). DOP-4294 provides the functionality to send requests to update the cache files in the `snooty-parse-cache` bucket.


src/cache-updater/index.ts
```diff
@@ -123,10 +123,18 @@ function getRepos(): RepoInfo[] {
...
-const repos = getRepos();
+async function main() {
+  const repos = getRepos();
 
-repos.forEach((repo) =>
-  handler(repo).catch((error) => {
-    console.error('An error occurred!', error);
-  })
-);
+  await Promise.all(
```
branberry, line 129: Added this because the job seemed to be running for a long time after the uploads were complete.

api/controllers/v2/cache.ts
```diff
@@ -0,0 +1,185 @@
...
+      commit.removed.some((fileName) => fileName === 'snooty.toml') ||
+      commit.modified.some((fileName) => fileName === 'snooty.toml')
+  );
+
+  if (!snootyTomlChanged) {
+    return { statusCode: 202, body: 'snooty.toml has not changed, no need to rebuild cache' };
+  }
+
+  const ref = body.ref;
+  // For webhook requests, this should only run on the primary branch, and if the repository belongs to the 10gen or mongodb orgs.
```
rayangler, line 157: Does this mean we'll have to add these new webhooks to every individual docs repo?

api/controllers/v2/cache.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  const repoOwner = body.repository.owner.login;
+  const repoName = body.repository.name;
+
+  // Checks the commits to see if there have been changes made to the snooty.toml file.
+  const snootyTomlChanged = body.commits.some(
+    (commit) =>
+      commit.added.some((fileName) => fileName === 'snooty.toml') ||
+      commit.removed.some((fileName) => fileName === 'snooty.toml') ||
+      commit.modified.some((fileName) => fileName === 'snooty.toml')
```
rayangler, line 149: Since we're looking for `snooty.toml` as the filename, does this mean that this won't be compatible with the monorepo?

api/controllers/v2/cache.ts
branberry: Yes, this is correct

api/controllers/v2/cache.ts
branberry: Yes, that is correct. The monorepo is not currently supported, but will be in the future

rayangler (approved):
Looks good based on my understanding 👍

PR #978, opened 2024-02-01
Remove Dockerfile.enhanced

branberry:
### Notes

PR to remove the `Dockerfile.enhanced`

[Successful build](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=65bc07336b2f17c95ad6eb21)


Dockerfile.enhanced
```diff
@@ -1,120 +0,0 @@
-# Build the Typescript app
```
rayangler, line 1: If we're removing this file, we should ensure that any references to this file are updated accordingly. I see references to this file in the cdk-infra [README](https://github.com/mongodb/docs-worker-pool/blob/5bd8a71397095051cbb95f47ee730bea6d1374f2/cdk-infra/README.md?plain=1#L74), Autobuilder GH actions workflows ([1](https://github.com/mongodb/docs-worker-pool/blob/5bd8a71397095051cbb95f47ee730bea6d1374f2/.github/workflows/deploy-stg-enhanced-worker.yml#L4) and [2](https://github.com/mongodb/docs-worker-pool/blob/5bd8a71397095051cbb95f47ee730bea6d1374f2/.github/workflows/update-feature-branch.yml#L106)), and the new `rebuild-parse-cache` [GH action](https://github.com/mongodb/docs-worker-actions). Not sure where else this is referenced

branberry: Ah, yes! Thank you. I should have remembered that one, haha!

rayangler (approved):

PR #979, opened 2024-02-01
DOP-4288: Add correct error handling to downloadBuildDependencies

mayaraman19:
DOP-4288

This PR corrects `downloadBuildDependencies` by enabling correct logging of successful/unsuccessful downloading of build dependencies. It also switches from using `curl` commands to `axios` to make the dependency request. Additionally, the build will fail if a dependency is not successfully downloaded.

Example build log link [here](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=65cceb56cbcd269fa52967ed).

To test, you can edit the `optionalBuildSteps.buildDependencies` object in a `reposBranches` entry for whatever docs property you want to test with, and mess with the links needed for building the property. Currently I've only really tested 404 errors - would be curious how other errors perform.


src/commands/src/helpers/dependency-helpers.ts
```diff
@@ -71,23 +72,23 @@ export async function downloadBuildDependencies(
...
+            return `curl -SfL ${dep.url} -o ${rootDir}${targetDir}/${dep.filename}`;
+          })
+          .catch((error) => {
+            return `ERROR! Could not curl ${dep.url} into ${rootDir}${targetDir}/${dep.filename}. ${error}`;
+          });
+        return curlString;
+      });
+
+      const responseSync = await Promise.all(response);
+      commands = commands.concat(responseSync);
```
mmeigs, line 91: Such a nit: I think I'd rather see us keep commands as a constant, and use the push method rather than concat which rewrites the entire array each time.

mmeigs (commented):
Looks good. I'd love more input from others on if we want an error in curling dependencies to fail the build. I lean toward yes. If a dependency does not download, I think we maybe want to have that be a fatal error. Maybe that's another ticket? Or maybe that's part of this one.

rayangler:
> If a dependency does not download, I think we maybe want to have that be a fatal error.

I would agree with this. An error in a build dependency would most likely cause downstream errors in built pages anyways, so I imagine failing the build earlier would be beneficial

rayangler:
I think I'm okay with either of the two approaches. Showing all failed dependencies together instead of just a single one might be the best in terms of convenience, so I think this would be my preference between the two if it's simple enough to do

src/job/jobHandler.ts
```diff
@@ -218,6 +218,8 @@ export abstract class JobHandler {
     await this._logger.save(this._currJob._id, 'Identified Build dependencies');
     const commands = await downloadBuildDependencies(buildDependencies, this.currJob.payload.repoName, directory);
     await this._logger.save(this._currJob._id, `${commands.join('\n')}`);
+    const err = commands.find((element) => element.includes('ERROR'));
+    if (err) throw new Error('Could not download some dependencies');
```
branberry, line 222: (Not a blocker) Do we want to throw an error if we don't have some build dependencies? I am not sure what the current behavior is. Wonder if some failed build dependencies would still be ok

branberry (approved):
Hi Maya! Looks great.

PR #985, opened 2024-02-05
DOP-4306: Remove legacy build code from autobuilder

anabellabuckvar:
### Context 
[DOP-4306](https://jira.mongodb.org/browse/DOP-4306)

Going forward, we will only support builds on Snooty so we can remove all of the not isNextGen code paths, and remove the checking logic

Using feature branch webhook, deployed README changes in `cloud-docs`. Built successfully
[Link to build logs](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=65c2a8c36b2f17c95a871de7)
[Link to preprd site](https://docs-atlas-staging.mongodb.com/cloud-docs/docsworker-xlarge/master/)

Pushed branch to preprd, used standard webhook to deploy README changes in `cloud-docs`. Built successfully without errors.
[Link to build logs](https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65c2ac266b2f17c95a88f394)
[Link to preprd site](https://docs-atlas-staging.mongodb.com/cloud-docs/docsworker-xlarge/anabellabuckvar-patch-7/)


src/job/jobHandler.ts
```diff
@@ -256,25 +254,20 @@ export abstract class JobHandler {
...
 
-      this.prepStageSpecificNextGenCommands();
-      this.constructEnvVars();
-      this.currJob.payload.isNextGen = true;
-      if (this._currJob.payload.jobType === 'productionDeploy') {
-        this._validator.throwIfNotPublishable(this._currJob);
-      }
-    } else {
-      this.currJob.payload.isNextGen = false;
+    this.prepStageSpecificNextGenCommands();
```
branberry, line 267: We should update the `prepStageSpecificNextGenCommands();`. That method contains stuff that we can clean up/simplify. More specifically, that function depends on mutations made to the `this.currJob.buildCommands` property in the `this.prepBuildCommands` method.

Anywhere we see `. /venv/bin/activate` or `make (parse/html/stage/deploy)` we should remove that.

branberry (changes requested):
Hey Anabella! Looks good so far, we just need to make some more changes where we reference legacy build commands and whatnot.

src/job/jobHandler.ts
anabellabuckvar: should `make (parse/html/stage/deploy)` be replaced with something else?

branberry: Good question. They should be replaced with the next-gen equivalent  e.g. make parse should be replaced with next-gen-parse. `make publish` should be removed as well, good catch!

caesarbell:
Probably unrelated, but figure I would ask, for my own learning purposes. With these changes would `buildsWithSnooty` in the `pool.repo_branches` collection still serve a purpose?

branberry (approved):
Looks great!

PR #989, opened 2024-02-07
DOP-4307: Remove legacy build steps from makefiles

anabellabuckvar:
### Context
[DOP-4307](https://jira.mongodb.org/browse/DOP-4307)
Simple PR to remove steps in Makefiles to that supported legacy tooling for building using Giza.

Changes tested by running autobuilder locally and updating the Dockerfile.local in main to use DOP-4307 instead of meta for the makefile download URL
 [docs-mongoid job logs](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=65ce658690c4e432a17b360f)
[cloudgov-docs job logs](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=65ce67aaf7a7320eab58cba0)
[docs-k8s-operator job logs](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=65ce6afd116c2613e360a724)


makefiles/Makefile.cloudgov-docs
```diff
@@ -20,6 +20,3 @@ include ~/shared.mk
 get-build-dependencies: 
 	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/cloudgov-docs.yaml > ${REPO_DIR}/published-branches.yaml
 
-
-html: ## Builds this branch's HTML under build/<branch>/html
-	giza make html
```
anabellabuckvar, line 25: I removed this rule and several in other files with this same format; however, there is no next-gen-html rule in this and some of the other makefiles to replace it. Is that of concern?

makefiles/Makefile.docs-ruby
anabellabuckvar, line 1: Can the `stage` rules directly following legacy publish rules also be removed?

makefiles/Makefile.cloudgov-docs
branberry: Nope! That shouldn't be of concern. These makefiles have an `include ~/shared.mk` statement at the top. The [`shared.mk`](https://github.com/mongodb/docs-worker-pool/blob/meta/makefiles/shared.mk) file contains the next-gen rules.

makefiles/Makefile.docs-ruby
branberry: Yep! They should be safe to remove

branberry (approved):
Looks great!

makefiles/Makefile.docs-mongoid
```diff
@@ -70,62 +70,4 @@ next-gen-stage: ## Host online for review
...
-#      			BUCKET    Amazon s3 bucket that holds the production docs
-#      --prefix=${PROJECT}   ${PROJECT} is the folder in the prod. bucket in which to place the artifacts
-#      --deploy             indicates that mut-publish should publish to the production bucket (not the staging bucket)
-#      --verbose            prints out a detail of what files are being uploaded/deleted/etc.
-#      --dry-run            instructs mut-publish to do everything *except* actually put stuff on the internet.
-#      if ${ARGS}, then additonal arguments
-	mut-publish build/public ${BUCKET} --prefix=${PROJECT} --deploy ${ARGS}
-
-	@echo "Hosted at ${URL}/${PROJECT}/index.html"
+	fi
```
schmalliso, line 73: There's no terminal newline on this one and Makefiles are weird about things, so please put one there in case it matters.

schmalliso (changes requested):
One quibble that probably won't matter but might because Makefiles are frustrating.

makefiles/Makefile.docs-mongoid
anabellabuckvar: Will do!

anabellabuckvar: Should be fixed, let me know if that's not what you had in mind

schmalliso: Sorry - I should have been more clear. We don't need a terminal `\`, we just need like literally to hit enter at the end of the `fi` line to move the cursor to the next row (so that little red negative circle disappears).

schmalliso (approved):

PR #992, opened 2024-02-12
DOP-4345 add docs-cpp makefile

anabellabuckvar:
Makefile for new repo

https://github.com/mongodb/docs-cpp

### DIFF
--- a/makefiles/Makefile.docs-cpp
+++ b/makefiles/Makefile.docs-cpp
@@ -1,0 +1,21 @@
+GIT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
+
+ifeq ($(STAGING_USERNAME),)
+	USER=$(shell whoami)
+else
+	USER=$(STAGING_USERNAME)
+endif
+
+PROJECT=cpp
+MUT_PREFIX ?= $(PROJECT)
+
+REPO_DIR=$(shell pwd)
+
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-cpp
```diff
@@ -0,0 +1,21 @@
...
+REPO_DIR=$(shell pwd)
+
+SNOOTY_DB_USR = $(shell printenv MONGO_ATLAS_USERNAME)
+SNOOTY_DB_PWD = $(shell printenv MONGO_ATLAS_PASSWORD)
+
+include ~/shared.mk
+
+
+get-build-dependencies: 
+	@curl https://raw.githubusercontent.com/mongodb/docs-worker-pool/meta/publishedbranches/docs-mongodb-shell.yaml > ${REPO_DIR}/published-branches.yaml
```
mmeigs, line 21: You don't need this anymore, yay!! However, for now the get-build-dependencies command needs to be here and cannot be empty. Look at newer repos's Makefiles to see how we've added echo statements instead!

mmeigs:
Left one comment!

makefiles/Makefile.docs-cpp
anabellabuckvar: updated it with the echo statement many of the makefiles have. Is that what you had in mind?

mmeigs (approved):
Yep! Looks good!

PR #1011, opened 2024-03-07
DOP-4414: Lambda function to handle automated test deploys

anabellabuckvar:
### Stories/Links:

[DOP-4414](https://jira.mongodb.org/browse/DOP-4414)

### Notes

This PR includes a new webhook that is triggered by workflow action updates. However, there is logic in docs-worker-pool such that the smoke test deploy lambda will only be triggered, and therefore, only run, upon successful completion of a Deploy Staging ECS push job to main.

This PR also includes some refactoring and simplification of preexisting code for slack deploys and feature branch push jobs.

Please take note of error handling and the status codes returned for different cases and leave feedback if applicable! 

### Testing
Webhook, endpoint, and lambda were all tested by pushing to this branch and therefore triggering the workflow. Github pushes on feature branches were also tested  and verified to still be working as expected; although this PR did not make meaningful changes to that functionality, it was incorporated in refactoring

Cloud Docs Build example:
[Build Logs]( https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=65fc57846fc5f04850aacdd5)
[Build URL](https://mongodbcom-cdn.website.staging.corp.mongodb.com/docs-qa/atlas)

### README updates


api/controllers/v2/github.ts
```diff
@@ -87,6 +141,138 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
+        body.workflow_run.head_branch +
+        ' will not trigger site deployments as it was not on master branch in upstream repo',
+    };
+  }
+
+  //automated test builds will always deploy in dotcomstg
+  const env = 'dotcomstg';
+
+  async function createAndInsertJob() {
+    for (const s in SMOKETEST_SITES) {
```
branberry, line 194: NIT: You can use the `of` keyword instead of `in` to get the value instead of the index i.e. `for (const repoName in SMOKEST_SITES)` would get you the value right away instead of having to index into the array

branberry, line 194: You could refactor the `for` loop and use `SMOKTEST_SITES.map` to allow for concurrent requests. Right now, these will occur sequentially since the for loop will `await` at each iteration at various points. If we use the map method built into arrays, we can get back an array of `Promise<void>` which we then can call `await Promise.all()`. This will make things a bit quicker 👍

src/repositories/projectsRepository.ts
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Project information from docs_metadata.projects for parser builds.
+ */
+export class ProjectsRepository extends BaseRepository {
+  constructor(db: Db, config: IConfig, logger: ILogger) {
+    super(config, logger, 'ProjectsRepository', db.collection('projects'));
+  }
+
+  async getProjectEntry(name: string): Promise<any> {
```
branberry, line 14: If we can, it would be great to properly type the return type. I know that the types are tied to the `BaseRepository`, so it might be a good chunk of work to make that possible. If that is the case, then no worries about having to do the work of updating all the types.

api/controllers/v2/github.ts
```diff
@@ -1,71 +1,125 @@
...
+  };
+}
+
+export const triggerSmokeTestAutomatedBuild = async (event: APIGatewayEvent): Promise<APIGatewayProxyResult | null> => {
   const client = new mongodb.MongoClient(c.get('dbUrl'));
   await client.connect();
   const db = client.db(c.get('dbName'));
   const consoleLogger = new ConsoleLogger();
   const jobRepository = new JobRepository(db, c, consoleLogger);
+  const projectsRepository = new ProjectsRepository(client.db('docs_metadata'), c, consoleLogger);
```
branberry, line 122: I think it would be good to put the DB name `'docs_metadata'` as a const variable in case that we use this repository somewhere else. This would make sure we don't accidentally misspell the name of the database somewhere else.

api/controllers/v2/github.ts
```diff
@@ -87,6 +141,138 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
+      statusCode: 202,
+      headers: { 'Content-Type': 'text/plain' },
+      body:
+        'Workflow ' +
+        body.workflow_run.name +
+        ' completed successfully but only Deploy Staging ECS workflow completion will trigger smoke test site deployments',
+    };
+
+  // if the build was not building master, no need for smoke test sites
+  if (body.workflow_run.head_branch != 'main' || body.repository.fork) {
```
branberry, line 178: Good idea to check for forks, I did not think of that!

branberry (commented):
Hi Anabella, thank you so much for your work on this! Looks great. I don't have any big concerns, just lots of small stuff. Great work.

branberry:
Hey Anabella, just one more thing. I do think it would be worth updating the README so that we know of this new process that occurs.

src/repositories/projectsRepository.ts
```diff
@@ -0,0 +1,23 @@
...
+import { IConfig } from 'config';
+import { BaseRepository } from './baseRepository';
+import { ILogger } from '../services/logger';
+
+/**
+ * Project information from docs_metadata.projects for parser builds.
+ */
+export class ProjectsRepository extends BaseRepository {
+  constructor(db: Db, config: IConfig, logger: ILogger) {
+    super(config, logger, 'ProjectsRepository', db.collection('projects'));
```
mmeigs, line 11: Like @branberry mentioned above with `docs_metadata`, could we make `projects` a constant like in the other ____Repository classes

mmeigs (commented):
Just one comment, and co-sign on all Brandon said! :) good work!

api/controllers/v2/github.ts
anabellabuckvar: Will do, thanks for the idea!

api/controllers/v2/github.ts
```diff
@@ -87,6 +141,138 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
+      const repoName = SMOKETEST_SITES[s];
+      const jobTitle = 'Smoke Test ' + repoName;
+      let repoInfo, projectEntry, repoOwner;
+
+      try {
+        repoInfo = await docsetsRepository.getRepo(repoName);
+        projectEntry = await projectsRepository.getProjectEntry(repoInfo.project);
+        repoOwner = projectEntry.github.organization;
+      } catch {
+        return 'repoInfo, projectEntry, or repoOwner not found for docs site ' + repoName;
```
mmeigs, line 204: nit: Could we implement some really basic logic here to try to narrow down the possibilities of the error? 
It could either use quick conditionals to find what is undefined, or we could simply try to log them all out within a template literal that shows which one is the issue.

api/controllers/v2/github.ts
```diff
@@ -87,6 +141,138 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
+      }
+
+      const jobPrefix = repoInfo?.prefix ? repoInfo['prefix'][env] : '';
+
+      //add commit hash here: const newHead = body.workflow_run.head_sha;
```
mmeigs, line 209: Just wondering what this is? A future ticket or something that can be erased?

src/repositories/projectsRepository.ts
anabellabuckvar: I added in a proper return type, but it's worth noting that nearly all of the methods in the `repositories` directory have a return type of `any`

api/controllers/v2/github.ts
anabellabuckvar: Future ticket! [DOP-4401](https://github.com/mongodb/docs-worker-pool/pull/1003). I can remove the comment now but it didn't seem like an issue to leave in since I'm already working on that future ticket

src/repositories/projectsRepository.ts
anabellabuckvar: @branberry @mmeigs to confirm, you mean like with the example of the docsets collection where it's set in `worker-env-construct.ts`from aws parameter store as shown [here](https://github.com/mongodb/docs-worker-pool/blob/025a8722b5aa69da322b4fbbc6cdee05f3d510e7/cdk-infra/lib/constructs/worker/worker-env-construct.ts#L67)?

api/controllers/v2/github.ts
anabellabuckvar: I changed it so that they are all logged in a template literal for simplicity, lmk if it looks good to you!

api/controllers/v2/github.ts
anabellabuckvar: I implemented this, please let me know if it's what you had in mind!

api/controllers/v2/github.ts
```diff
@@ -1,71 +1,125 @@
...
+    jobType = 'productionDeploy';
+    branchName = 'master';
+  } else {
+    if (!githubEvent) {
+      throw 'Non SmokeTest Deploy jobs must have a github Event';
+    }
+    action = 'push';
+    jobType = 'githubPush';
+    branchName = githubEvent.ref.split('/')[2];
+    url = githubEvent?.repository.clone_url;
```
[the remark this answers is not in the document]
anabellabuckvar: Done

api/controllers/v2/github.ts
```diff
@@ -1,71 +1,125 @@
...
-      isFork: githubEvent.repository.fork,
-      repoOwner: githubEvent.repository.owner.login,
-      url: githubEvent.repository.clone_url,
-      newHead: githubEvent.after,
-      urlSlug: urlSlug,
-      prefix: prefix,
-      project: project,
-      directory: directory,
-    },
+    payload: payload,
```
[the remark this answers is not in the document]
anabellabuckvar: Done

src/repositories/projectsRepository.ts
anabellabuckvar: I added `projects` as a constant as done with the collections for the other repository classes except that I didn't configure it to pull the value from AWS parameter store. If that looks good to you both I'll do the same with `docs_metadata` as a constant

.github/workflows/deploy-stg-enhanced-webhooks.yml
```diff
@@ -1,9 +1,10 @@
 on:
   push:
-    paths: ["api/v2/**", "cdk-infra/lib/constructs/api/**", "cdk-infra/utils/**"]
+    # paths: ["api/v2/**", "cdk-infra/lib/constructs/api/**", "cdk-infra/utils/**"]
```
mmeigs, line 3: I believe put this back in?

.github/workflows/deploy-stg-enhanced-webhooks.yml
```diff
@@ -1,9 +1,10 @@
...
     branches:
       - "main"
       - "integration"
+      - "DOP-4414"
```
mmeigs, line 7: NOTE: reminder to remove before merge :)

api/controllers/v2/github.ts
```diff
@@ -79,6 +133,139 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
+  //   return {
+  //     statusCode: 202,
+  //     headers: { 'Content-Type': 'text/plain' },
+  //     body:
+  //       `Build on branch
+  //       ${body.workflow_run.head_branch}
+  //       will not trigger site deployments as it was not on main branch in upstream repo`,
+  //   };
+  // }
+
```
mmeigs, line 186: Should this be removed or uncommented?

mmeigs (commented):
I think the environment variable should work out, but you don't need to add it to any of the 4 config files you have because those are never touched if you're not accessing it through the config object!

.github/workflows/deploy-stg-enhanced-webhooks.yml
anabellabuckvar: Yup, this was just commented out for testing  purposes!

api/controllers/v2/github.ts
anabellabuckvar: This will be uncommented before merge, it just needs to be commented out so I can continue testing any changes

api/controllers/v2/github.ts
```diff
@@ -79,6 +133,133 @@ export const TriggerBuild = async (event: APIGatewayEvent): Promise<APIGatewayPr
...
+          return jobId;
+        } catch (err) {
+          consoleLogger.error('TriggerBuildError', `${err} Error inserting job for ${repoName}`);
+          return err;
+        }
+      })
+    );
+  }
+
+  let returnVal;
```
branberry, line 228: NIT: We don't need to use `let returnVal` here. We can return in the `try` statement with the `returnVal`

branberry (approved):
One small non-blocking change. Other than that, LGTM!

src/repositories/repoEntitlementsRepository.ts
```diff
@@ -73,7 +73,7 @@ export class RepoEntitlementsRepository extends BaseRepository {
       `Mongo Timeout Error: Timedout while retrieving entitlements for ${slackUserId}`
     );
     // if user has specific entitlements
-    if ((entitlementsObject?.repos?.length ?? 0) > 0) {
+    if (entitlementsObject?.repos && entitlementsObject.repos.length > 0) {
```
mmeigs, line 76: nit: we can shorten this to `entitlementsObject?.repos?.length > 0`

api/controllers/v2/github.ts
```diff
@@ -1,71 +1,125 @@
...
+  let branchName: string;
+  let jobType: string;
+  let action: string;
+  let url: string;
+
+  if (isSmokeTestDeploy) {
+    url = 'https://github.com/' + repoOwner + '/' + repoName;
+    action = 'automatedTest';
+    jobType = 'productionDeploy';
+    branchName = 'master';
```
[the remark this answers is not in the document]
anabellabuckvar: Good point, thank you. I did consider this, but I checked at least the specific content repos that Allison requested always be deployed for smoke tests, and those at least all use "master" as the master branch.

If, at a later date, someone, for some reason, adds a repo to the list that uses main, they will have to add additional logic to handle that then. However, I agree that I don't believe it should be blocking right now for that hypothetical

src/job/productionJobHandler.ts
```diff
@@ -141,24 +141,36 @@ export class ProductionJobHandler extends JobHandler {
...
+  //         return `${this.currJob.payload.prefix}/${this.currJob.payload.urlSlug}`;
+  //       }
+  //     }
+  //     return this.currJob.payload.prefix;
+  //   } catch (error) {
+  //     this.logger.save(this.currJob._id, error).then();
+  //     throw new InvalidJobError(error.message);
+  //   }
+  // }
+
```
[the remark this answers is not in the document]
anabellabuckvar: It was commented out so I could do some testing! It's added back in now :)

api/controllers/v2/github.ts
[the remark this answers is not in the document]
mmeigs: Agreed!

src/job/productionJobHandler.ts
```diff
@@ -141,36 +141,24 @@ export class ProductionJobHandler extends JobHandler {
...
-      ? `${this.currJob.payload.prefix}/${this.currJob.payload.urlSlug}`
-      : this.currJob.payload.prefix;
-    if (this.currJob.payload.action == 'automatedTest') {
-      const titleArray = this.currJob.title.split(' ');
-      const commitHash = titleArray[titleArray.length - 5];
-      return `${prefix}/smokeTests/${commitHash}`;
+    try {
+      if (this.currJob.payload.prefix && this.currJob.payload.prefix === '') {
+        return this.currJob.payload.urlSlug ?? '';
+      }
```
mmeigs, line 147: I'm only worried about this because it is in the productionJobHandler and it's a difference in pathing that does not solely rely on making sure this is a deploy_all situation. Could we also make sure that this has a conditional dependent on if this job is specifically a deploy all slack command?

anabellabuckvar: When I click on the link to the version of the productionJobHandler file you reviewed, it shows that it is only reflecting changes from the past 4 commits- the changes on these lines within those commits are actually just reverting previous changes made. If you look at  those lines in src/job/productionJobHandler.ts [here](https://github.com/mongodb/docs-worker-pool/pull/1011/files#diff-ca78240af09f6070db1fcf5584b68368d38686db292057b80ab9469e204c42fc) you'll actually see that there aren't any differences in comparison to the main branch there!

mmeigs: ohhhhhhh wonderful.
You're correct, I must have been looking at the "changes since I last looked"

mmeigs (approved):

PR #1022, opened 2024-03-28
DOP-4481: Test-deploy all docs site button

anabellabuckvar:
### Stories/Links:

[DOP-4481](https://jira.mongodb.org/browse/DOP-4481)

### Notes
This PR introduces **deploy-all** button in Slack's test-deploy dialog! An image of what this looks like is included below. The radio button deploy options selectively render such that they only display for users who are admins as set in pool_test.entitlements(all DOP team members).  If the option "deploy-all" is selected upon modal submission, all prod-deployable non-internal docs sites are deployed. 

Unfortunately, Slack's [block-kit builder](https://api.slack.com/block-kit) has some major limitations.  The types of buttons available are limited to regular buttons, radio buttons, and checkboxes, but more importantly, Slack does not support disabling components. Therefore, to disable selecting individual repos after selecting the 'deploy_all' button, the view or payload message would have to be updated and to remove the button entirely after click. This workflow would be unintuitive and adds complexity. 

As a result, I've made the "select repos" dropdown optional so that the only required component is the radio buttons. For the same reasons explained above, it is still possible to select an individual repo(s) AND select the deploy all radio option, in which case the individual repo(s) selected will be ignored and all repos will be deployed as normal.

Screenshot 2024-04-03 at 10 06 57 AM

** currently, all repos are deployed with mongodb as the owner. There are changes in a different PR currently in review that enables determination of a repos' proper repo owner. Once that PR and the current PR is merged, a small ticket will be made to fix how repo owner is determined in this workflow

### README updates


api/controllers/v1/slack.ts
```diff
@@ -96,12 +117,11 @@ export const getDeployableJobs = async (
...
     if (!branchObject?.aliasObject) continue;
 
-    // TODO: Create strong typing for these rather than comments
-    const publishOriginalBranchName = branchObject.aliasObject.publishOriginalBranchName; // bool
-    let aliases = branchObject.aliasObject.urlAliases; // array or null
-    let urlSlug = branchObject.aliasObject.urlSlug; // string or null, string must match value in urlAliases or gitBranchName
-    const isStableBranch = branchObject.aliasObject.isStableBranch; // bool or Falsey
-    aliases = aliases?.filter((a) => a);
+    const publishOriginalBranchName: boolean = branchObject.aliasObject.publishOriginalBranchName;
+    const aliases: [] | null = branchObject.aliasObject.urlAliases;
```
branberry, line 121: I think this should be `string[] | null` for the type.

src/job/stagingJobHandler.ts
```diff
@@ -49,7 +49,7 @@ export class StagingJobHandler extends JobHandler {
   prepDeployCommands(): void {
     this.currJob.deployCommands = [
       `cd repos/${getDirectory(this.currJob)}`,
-      `make next-gen-stage${this.currJob.payload.pathPrefix ? `MUT_PREFIX=${this.currJob.payload.mutPrefix}` : ''}`,
+      `make next-gen-stage${this.currJob.payload.pathPrefix ? ` MUT_PREFIX=${this.currJob.payload.mutPrefix}` : ''}`,
```
branberry, line 52: NIT: I think we can remove this white space

src/services/slack.ts
```diff
@@ -22,8 +23,8 @@ function timeSafeCompare(a: string, b: string) {
 
 export interface ISlackConnector {
   validateSlackRequest(payload: any): boolean;
-  displayRepoOptions(repos: Array<string>, triggerId: string): Promise<any>;
-  parseSelection(payload: any): any;
+  displayRepoOptions(repos: Array<string>, triggerId: string, admin: boolean): Promise<any>;
+  parseSelection(payload: any, entitlement: any, repoBranchesRepository: RepoBranchesRepository): any;
```
branberry, line 27: NIT: It looks like `entitlement` isn't used in the implementation. I think it should be `isAdmin: boolean` instead for the second parameter.

src/services/slack.ts
```diff
@@ -52,25 +53,49 @@ export class SlackConnector implements ISlackConnector {
...
     };
 
+    // if deploy all was selected:
+    if (stateValues['block_deploy_option']['deploy_option']?.selected_option?.value == 'deploy_all') {
+      if (isAdmin) {
+        values['deploy_option'] = 'deploy_all';
+        values['repo_option'] = await repoBranchesRepository.getProdDeployableRepoBranches();
+        return values;
+      } else
+        throw {
```
branberry, line 76: We should `return` instead of `throw` here. If we throw, the API response will not be this because API gateway expects an object to be returned from the Lambda function.

src/services/slack.ts
```diff
@@ -52,25 +53,49 @@ export class SlackConnector implements ISlackConnector {
...
-      }
+
       // multi select is an array
-      else if (stateValuesObj?.selected_options && stateValuesObj.selected_options.length > 0) {
+      if (stateValuesObj?.selected_options?.length > 0) {
         values[blockInputKey] = stateValuesObj.selected_options;
       }
+      //retrun an error if radio button choice 'deploy individual repos' was selected but no repo was actually chosen
+      else if (blockInputKey == 'repo_option') {
+        throw {
```
branberry, line 94: We should return instead of throw here

api/controllers/v1/slack.ts
```diff
@@ -13,13 +14,23 @@ import {
   prepResponse,
 } from '../../handlers/slack';
 import { DocsetsRepository } from '../../../src/repositories/docsetsRepository';
+import { Payload } from '../../../src/entities/job';
 
-export const DisplayRepoOptions = async (event: any = {}, context: any = {}): Promise<any> => {
+export const DisplayRepoOptions = async (event: APIGatewayEvent): Promise<APIGatewayProxyResult> => {
```
branberry, line 19: Thank you for updating these types.

branberry (commented):
Hey Anabella, this looks great! Thank you for updating the types as well. Left a few comments, but nothing major

src/job/stagingJobHandler.ts
anabellabuckvar: I'd prefer to keep it, if that's ok! I'll explain:

There is a test for make next-gen-stage that expecets the whitespace there for when `this.currJob.payload.pathPrefix` evaluates to true here. The reason it wasn't failing before is that this.currJob.payload.pathPrefix, AFAIK, always evaluates to a falsy value at this point in the code. 
This is because getPathPrefix() is called in jobHandler.ts and sets the payload's pathPrefix, but for staging jobs, it always returns an empty string. I realized the test would fail if there was no extra whitespace because when I was testing how the prefix works with some staging jobs, I overrode how pathPrefix is currently set so that it would always evaluate to a truthy value.

src/services/slack.ts
anabellabuckvar: good catch, thank you!

api/controllers/v1/slack.ts
```diff
@@ -183,14 +202,26 @@ export const DeployRepo = async (event: any = {}, context: any = {}): Promise<an
   const parsed = JSON.parse(decoded);
   const stateValues = parsed.view.state.values;
 
+  //TODO: create an interface for slack view_submission payloads
+  if (!(parsed.type == 'view_submission')) {
```
mmeigs, line 206: nit: could we make this the more readable: `if (parsed.type !== 'view_submission')`?

src/services/slack.ts
```diff
@@ -22,8 +23,8 @@ function timeSafeCompare(a: string, b: string) {
 
 export interface ISlackConnector {
   validateSlackRequest(payload: any): boolean;
-  displayRepoOptions(repos: Array<string>, triggerId: string): Promise<any>;
-  parseSelection(payload: any): any;
+  displayRepoOptions(repos: Array<string>, triggerId: string, admin: boolean): Promise<any>;
+  parseSelection(payload: any, isAdmin: boolean, repoBranchesRepository: RepoBranchesRepository): any;
```
mmeigs, line 27: nit: could we name these parameters the same? I like the `isAdmin` to make it obviously a boolean

src/services/slack.ts
```diff
@@ -52,25 +53,50 @@ export class SlackConnector implements ISlackConnector {
...
-      if (stateValuesObj?.selected_option?.value) {
-        values[blockInputKey] = stateValuesObj.selected_option.value;
-      }
+
       // multi select is an array
-      else if (stateValuesObj?.selected_options && stateValuesObj.selected_options.length > 0) {
+      if (stateValuesObj?.selected_options?.length > 0) {
         values[blockInputKey] = stateValuesObj.selected_options;
       }
+      //retrun an error if radio button choice 'deploy individual repos' was selected but no repo was actually chosen
```
mmeigs, line 93: nit: typo (return)

src/services/slack.ts
```diff
@@ -110,8 +137,51 @@ export class SlackConnector implements ISlackConnector {
       },
     });
   }
+  private _getDropDownView(triggerId: string, repos: Array<any>, admin: boolean) {
+    const deployAll = admin
+      ? {
```
mmeigs, line 142: Would love that renaming: isAdmin :)

src/services/slack.ts
branberry: Following up on this, if we do want to throw here, let's create a new error object and throw that e.g. `throw new Error('error occurred');`. In the handler function where we catch the error, we can then return the object that you were originally throwing. If you need to differentiate between error types, you can either use the message to determine that, or maybe add an error type property. The latter approach would be a bit more work, so don't worry about it too much.

The benefit of this is that it's clear what the handler function will be returning.

anabellabuckvar: Thank you for the explanation! I've made changes so that an error object is thrown. How does that look to you?

branberry (approved):
LGTM, thank you!!

src/services/slack.ts
```diff
@@ -52,25 +53,44 @@ export class SlackConnector implements ISlackConnector {
     }
     return {};
   }
-  parseSelection(stateValues: any): any {
+
+  async parseSelection(
+    stateValues: any,
+    isAdmin: boolean,
+    repoBranchesRepository: RepoBranchesRepository
+  ): Promise<any> {
```
mmeigs, line 61: nit: if possible, can we make the return value more specific? But if that's complicated, totally not-blocking!

mmeigs (approved):
One non-blocking comment! Thank you for all the hard work on this!! Will be so useful.

src/services/slack.ts
anabellabuckvar: I'm open to that! However, I'm not sure what approach to take in this situation. `parseSelection` always returns an object called `values` as a Promise, but the attributes of `values` are inconsistent so it's not strongly typed itself. Ex: `values` always has a `repo_option` property but only sometimes a `hash_option` or `deploy_option`.
  I believe(?) specifying the return type further than Promise<object> would require a refactor of the function and some checks added elsewhere, but please correct me if you have something else in mind! If not, happy to do this in the follow up ticket to this

mmeigs: Yeah, non-blocking! We can always do this later! Seems like  a larger time commitment :)

PR #1026, opened 2024-04-03
DOP-4452: Add bucket routing rule to rehome Realm docs from /docs/realm to /docs/atlas/device-sdks

mayaraman19:
### Stories/Links:

DOP-4452

Rerouting all `/docs/realm` sites to `/docs/atlas/device-sdks`.
Adding this so that the bucket routing rules I created in AWS don't get overwritten by the next release. 

### Notes:

Not 100% sure how to test this?

### README updates


mmeigs (commented):
I believe pushing this up to certain branches (main and integration, etc.) will write it to AWS for those environments. I could be wrong, but I've seen it happen recently.

To test, I belieeeeve you could takeover preprd, push this up to it, and it should be altered within dotcomstg!

schmalliso (approved):

PR #1035, opened 2024-05-02
:package: DOP-4497 updtes the mongodb node driver

caesarbell:
### Stories/Links:

DOP-4497

### Notes

This PR was focused on updating MongoDB Node Driver from`v5.1.0` to `v6.5`. With upgrading the version, an issue occurred where the build would hang. 

Upon running the AB locally, I was able to discover that  `value` was not a property of the response, which lead me to remove the `value` proeprty and return just the response.

After the above change, I was able to successfully trigger the AB build using the webhook, resulting in the following [master](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=66354409b67dbf804abf9049) and [DOP-4497](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=queue&jobId=66354909b67dbf804ac10a87) build results.

The page can be viewed [here](https://docs-mongodbcom-staging.corp.mongodb.com/docs/docsworker-xlarge/master/).

**Update**
After some additional research, the reason for this change was because in 6.0+ sets `includeResultMetadata` to false by default. Adding `includeResultMetadata` true as an option would allow us to return `value` which would contain our returned document.

### Resource Links

[Behavioral Changes to Find One Family APIs](https://www.mongodb.com/blog/post/behavioral-changes-find-one-family-apis-node-js-driver-6-0-0)
[Server Release Compatibility Changes](https://www.mongodb.com/docs/drivers/node/current/upgrade/#server-release-compatibility-changes)

### README updates


mmeigs (commented):
Hmm, this makes me a teeny bit nervous because this change could/is more wide-ranging.

1. Do we know where in breaking changes for the Node driver this occurs? lol @ searching our own documentation
2. This (the `value` property) definitely is accessed other places.
- one example: in the same file (`src/repositories/jobRepository`) in the function `updateWithStatus`.

caesarbell:
For test sake, I applied the following option `{includeResultMetadata: true},` to 

```js
async updateWithStatus(
    id: string | mongodb.ObjectId,
    result: any,
    status: JobStatus,
    shouldNotifySqs = true
  ):
  ```
  
  to allow the `value` to be returned, since removing `value` from  `updateWithStatus` method causes the test to fail with some odd behavior.

File: src/repositories/jobRepository.ts [hunk 1]
```diff
@@ -122,8 +122,8 @@ export class JobRepository extends BaseRepository {
       options,
       `Mongo Timeout Error: Timed out while retrieving job`
     );
-    if (response.value) {
-      const job: Job = response.value;
+    if (response) {
+      const job: Job = response;
```
mmeigs, line 126: We should change this back!

tests/unit/repositories/jobRepository.test.ts
```diff
@@ -106,11 +106,11 @@ describe('Job Repository Tests', () => {
 
     test('getOneQueuedJobAndUpdate succeeds', async () => {
       const testData = TestDataProvider.getFindOneAndUpdateCallInfo();
-      const mockVal = { value: getBuildJobPlain() };
+      const mockVal = getBuildJobPlain();
       jest.spyOn(jobRepo, 'notify').mockResolvedValueOnce(true);
       dbRepoHelper.collection.findOneAndUpdate.mockResolvedValueOnce(mockVal);
 
-      await expect(jobRepo.getOneQueuedJobAndUpdate()).resolves.toEqual(mockVal.value);
+      await expect(jobRepo.getOneQueuedJobAndUpdate()).resolves.toEqual(mockVal);
```
mmeigs, line 113: And change this back, I believe!

src/repositories/jobRepository.ts
caesarbell: @mmeigs even when all the tests are working? I guess you are right, this will keep things consistent.

mmeigs (approved):
LGTM

src/repositories/jobRepository.ts
```diff
@@ -115,7 +115,7 @@ export class JobRepository extends BaseRepository {
 
   async findOneAndUpdateJob(query): Promise<Job | null> {
     const update = { $set: { startTime: new Date(), status: 'inProgress' } };
-    const options = { sort: { priority: -1, createdTime: 1 }, returnNewDocument: true };
+    const options = { sort: { priority: -1, createdTime: 1 }, returnNewDocument: true, includeResultMetadata: true };
```
seungpark, line 118: good catch! think this can be a default option for BaseRepository so we don't have to repeat it on each call

seungpark (approved):
minor nitpick below but overall LGTM 👍 thanks for testing and providing changelog documents!

PR #1042, opened 2024-05-08
DOP-4599: Deploy-all slack button deploys all active versions of repos

anabellabuckvar:
### Stories/Links:

[DOP-4599](https://jira.mongodb.org/browse/DOP-4599)

### Notes

This PR accomplishes two goals:

1. Anyone marked as an "admin" in their personal Atlas pool.entitlements entry can deploy all repos in repos branches without having to manually add _any_ entitlements to their entitlements array
2. The recently released deploy-all button will now deploy all prod-deployable repos and all of thei their active versions, as opposed to just the master branch of all prod-deployable repos

The PR is currently in PrePrd for reviewing purposes so reviewers can test themselves.
### README updates


src/services/slack.ts
```diff
@@ -72,7 +67,16 @@ export class SlackConnector implements ISlackConnector {
...
-      values['repo_option'] = await repoBranchesRepository.getProdDeployableRepoBranches();
+      //go through all options in dropdown by option group
+      //append version to repo_option if active
+      values['repo_option'] = [];
+      for (const group of optionGroups) {
+        values['repo_option'].push(
+          ...group.options.map((option) => {
+            if (!option.text.text.startsWith('(!inactive)')) return option;
+          })
+        );
```
mmeigs, line 78: We should use the `filter` method for this rather than `map` so that we don't end up with `undefined` values in the array

api/controllers/v1/slack.ts
```diff
@@ -66,7 +81,7 @@ async function deployRepo(deployable: Array<any>, logger: ILogger, jobRepository
   try {
     await jobRepository.insertBulkJobs(deployable, jobQueueUrl);
   } catch (err) {
-    logger.error('deployRepo', err);
+    console.error('Deploy repo error');
```
mmeigs, line 84: Could we still log the error? It can sometimes be such a blessing.

api/controllers/v1/slack.ts
```diff
@@ -142,7 +150,7 @@ export const getDeployableJobs = async (
       directory
     );
 
-    if (!aliases || aliases.length === 0) {
+    if (!aliases || aliases.length) {
```
rayangler, line 153: Do we want to keep the check for 0 aliases? Please correct me if I'm wrong, but I believe this change will prevent aliases from building

anabellabuckvar: yup this was an accident, thanks!

api/controllers/v1/slack.ts
anabellabuckvar: Happy to, but it won't make a difference in this case! The only place we can view the logs for the Slack Deploy Repo lambda is in its corresponding CloudWatch group, in which the output for console.error looks like the following screenshot:
Screenshot 2024-05-30 at 2 52 48 PM

src/services/slack.ts
```diff
@@ -72,7 +67,16 @@ export class SlackConnector implements ISlackConnector {
...
 
       values['deploy_option'] = 'deploy_all';
-      values['repo_option'] = await repoBranchesRepository.getProdDeployableRepoBranches();
+      //go through all options in dropdown by option group
+      //append version to repo_option if active
+      values['repo_option'] = [];
+      for (const group of optionGroups) {
+        values['repo_option'].push(
+          ...group.options.filter((option) => {
+            !option.text.text.startsWith('(!inactive)');
```
rayangler, line 76: We should return this to ensure things are being filtered as expected. Right now, I believe the array would be empty

api/controllers/v1/slack.ts
```diff
@@ -65,8 +80,9 @@ export const DisplayRepoOptions = async (event: APIGatewayEvent): Promise<APIGat
 async function deployRepo(deployable: Array<any>, logger: ILogger, jobRepository: JobRepository, jobQueueUrl) {
   try {
     await jobRepository.insertBulkJobs(deployable, jobQueueUrl);
+    console.error('testing console error logging');
```
rayangler, line 83: Should we remove this?

anabellabuckvar: Yup, it will be- left it in for Matt to reference regarding his comment on logging !

rayangler:
Hey @anabellabuckvar, I think there may be a bug with the test deployment process in that some jobs seem to fail due to an authorization error. I was looking into it and I wonder if we need to update the `throwIfUserNotEntitled` such that it bypasses the check for admin users. For example, I don't have `mongodb/docs-kotlin` as an entitlement, but attempting to test deploy all resulted in the error ([build log](https://workerpoolstaging-qgeyp.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=665a38e2cea8353fa1817820)).

api/controllers/v1/slack.ts
mmeigs: Fascinating! I wonder if that's always the case... I lean towards keeping the error logged just in case it can sometimes be helpful. But good catch!

PR #1043, opened 2024-05-09
DOP-4701: Preserve stdout in ShellCommandExecutor

i80and:
### Stories/Links:

[DOP-4701](https://jira.mongodb.org/browse/DOP-4701)

### Notes

The actual fix is the one line change in `commandExecutor.ts`.

The annoying part was making the test work:
* The existing test incorrectly mocked the `child_process.exec()` interface.
* According to the node docs, the exception raised by `util.promisify(child_process.exec())` will always have a `stdout` property. But just in case the exception get raised by something *else* in the try block, it'll fall back to returning `''` as in the current behavior.
* `util.promisify(child_process.exec)` has a custom promisifier (`child_process.exec[util.promisify.custom]`) that considerably changes the interface, but that jest loses when you call `jest.mock('child_process')`, and because of jest magic tomfoolery, I couldn't find any tidy way to *restore* that custom hook, which is needed to actually make the interface behave correctly.

  So I thought, you know what? Why mock? What value do we get by mocking `child_process.exec` besides introducing surface areas for bugs?
  
  So this PR just executes `commandExecutor.execute(["echo 'stdout output'", "echo 'stderr output' >&2", 'exit 1'])`, which gives us everything we need to make sure ShellCommandExecutor actually works, using only shell builtins that we already require.

### README updates

### DIFF
--- a/src/services/commandExecutor.ts
+++ b/src/services/commandExecutor.ts
@@ -65,7 +65,7 @@ export class ShellCommandExecutor implements ICommandExecutor {
       resp.status = CommandExecutorResponseStatus.success;
       return resp;
     } catch (error) {
-      resp.output = '';
+      resp.output = (error.stdout || '').trim();
       resp.error = error;
       resp.status = CommandExecutorResponseStatus.failed;
     }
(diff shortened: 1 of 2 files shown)


mmeigs (approved):
Thank you for the explanations!!

PR #1054, opened 2024-06-10
DOP-4704: Ignore errors when rsyncing mongo-ruby-driver

i80and:
This is an intermediate step to allow building docs-ruby without that submodule.

This line should go away entirely once the transition to not using a submodule is complete.

### Stories/Links:

DOP-4704

### README updates

### DIFF
--- a/makefiles/Makefile.docs-ruby
+++ b/makefiles/Makefile.docs-ruby
@@ -59,8 +59,8 @@ get-build-dependencies: fetch-submodule
 
 fetch-submodule:
 	git submodule update --remote --init
-	rsync -a --delete ${REPO_DIR}/mongo-ruby-driver/docs/ ${REPO_DIR}/source
-	rsync -a --delete ${REPO_DIR}/bson-ruby/docs/tutorials/bson.txt ${REPO_DIR}/source/tutorials/bson.txt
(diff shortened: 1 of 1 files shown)


makefiles/Makefile.docs-ruby
```diff
@@ -59,7 +59,7 @@ get-build-dependencies: fetch-submodule
 
 fetch-submodule:
 	git submodule update --remote --init
-	rsync -a --delete ${REPO_DIR}/mongo-ruby-driver/docs/ ${REPO_DIR}/source
+	-rsync -a --delete ${REPO_DIR}/mongo-ruby-driver/docs/ ${REPO_DIR}/source
 	rsync -a --delete ${REPO_DIR}/bson-ruby/docs/tutorials/bson.txt ${REPO_DIR}/source/tutorials/bson.txt
```
rayangler, line 63: (Non-blocking wonder) Looked at the ticket, but I wonder why we're interested in removing only the `mongo-ruby-driver` submodule instead of both 🤔

i80and: It's a good question! I'll ask!

i80and: We are removing both!

rayangler: Sounds good! Thanks for double-checking! LGTM!

PR #1057, opened 2024-06-24
DOP-4758: pull rstspec from tagged version 

seungpark:
### Stories/Links:

DOP-4758

### Notes
- [Prereq PR](https://github.com/mongodb/docs-worker-pool/pull/1058) for makefile changes
- This PR aims to get rid of `latest` references to rstspec and points these to the latest tag

### README updates

- - [ ] This PR introduces changes that should be reflected in the README, and I have made those updates.
- - [x] This PR does not introduce changes that should be reflected in the README

PR #1058, opened 2024-06-25
DOP-4758: remove flag for rstspec

seungpark:
### Stories/Links:

DOP-4758

### Notes
- Makefiles should point at the env var ([specified in Dockerfile](https://github.com/mongodb/docs-worker-pool/pull/1057/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R25)) or fallback to the main branch for latest rstspec

### README updates

### DIFF
--- a/makefiles/shared.mk
+++ b/makefiles/shared.mk
@@ -15,11 +15,11 @@ PATCH_CLAUSE=$(shell if [ ! -z "${PATCH_ID}" ]; then echo --patch "${PATCH_ID}";
 
 BUNDLE_PATH=${REPO_DIR}/bundle.zip
 ifdef SNOOTY_PARSER_VERSION
-PARSER_VERSION := $(SNOOTY_PARSER_VERSION)
+PARSER_VERSION := v$(SNOOTY_PARSER_VERSION)
 else
 PARSER_VERSION := main
 endif
-RSTSPEC_FLAG=--rstspec=https://raw.githubusercontent.com/mongodb/snooty-parser/v${PARSER_VERSION}/snooty/rstspec.toml
+RSTSPEC_FLAG=--rstspec=https://raw.githubusercontent.com/mongodb/snooty-parser/${PARSER_VERSION}/snooty/rstspec.toml
 
 ifeq ($(SNOOTY_INTEGRATION),true)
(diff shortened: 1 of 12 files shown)


seungpark:
Removed the DO NOT MERGE label as it will safely fall back to `main` branch without the env flags - the rstspec consolidation has [been merged to main](https://github.com/mongodb/snooty-parser/pull/607)

i80and:
Is there a reason we don't want to simply use the built-in rstspec? Under what circumstances would we want to build with a different parser version's rstspec?

seungpark:
> Is there a reason we don't want to simply use the built-in rstspec? Under what circumstances would we want to build with a different parser version's rstspec?

discussed offline! updated to remove the flag completely as it will be loaded from the same version if not specified

makefiles/shared.mk
```diff
@@ -14,8 +14,6 @@ PATCH_ID=$(shell if test -f "${PATCH_FILE}"; then git patch-id < ${PATCH_FILE} |
 PATCH_CLAUSE=$(shell if [ ! -z "${PATCH_ID}" ]; then echo --patch "${PATCH_ID}"; fi)
 
 BUNDLE_PATH=${REPO_DIR}/bundle.zip
-RSTSPEC_FLAG=--rstspec=https://raw.githubusercontent.com/mongodb/snooty-parser/latest/snooty/rstspec.toml
```
i80and, line 17: We should also remove the `${RSTSPEC_FLAG}` references in makefiles that override the `next-gen-parse` rule. I count 12 of them :(

* `Makefile.docs-atlas-cli`
* `Makefile.docs-atlas-operator`
* `Makefile.docs-landing`
* `Makefile.docs-laravel`
* `Makefile.mongocli`
* `Makefile.mongodb-internal-base`
* `Makefile.docs-mongodb-internal`
* `Makefile.mongoid`
* `Makefile.docs-php-library`
* `Makefile.docs-ruby`
* `Makefile.docs`
* `Makefile.docs-stage`

seungpark: great catch! removed from these makefiles

i80and (approved):
Good riddance -- this was always such a pain!

PR #1064, opened 2024-07-25
Release v1.5.23

mmeigs:
### Notes:

Update Dockerfiles for Release v1.5.23


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "workerpool",
-  "version": "1.0.0",
```
i80and, line 3: Why are we downgrading the version number?

mmeigs: I was upgrading it to the current version, idk thought it'd be nice. But is this not what I should do?

i80and: OH. I misread the diff somehow. Facepalm sorry!!

i80and (approved):

PR #1072, opened 2024-09-19
Set desired number of workers for the enhanced autobuilder to 0

branberry:
### Notes

Since we've removed the autobuilder for staging, we no longer need the long running tasks. This is an easy way to cut down our AWS costs by a significant amount. 

### README updates

- - [ ] This PR does not introduce changes that should be reflected in the README


mmeigs (approved):
These PRs are so exciting

PR #1088, opened 2025-03-10
DOP-5450: Change atlas operator routing `/atlas-operator` -> `/operator`

mmeigs:
### Stories/Links:

DOP-5450

### Notes

The bucket level redirect needs to change. I will merge this on approval and redeploy the autobuilder to propagate the changes to S3 as we spoke about in standies.

### README updates


seungpark (approved):
just making sure that we don't need the previous redirects! otherwise looks good

rayangler (approved):

rayangler:
> just making sure that we don't need the previous redirects!

I think we should be good to replace, [according to JW](https://mongodb.slack.com/archives/C018UTGP23T/p1740499202724269?thread_ts=1740438380.392939&cid=C018UTGP23T)

PR #1093, opened 2025-06-25
DOP-5854: only build one version for associated products at a time

seungpark:
### Stories/Links:

DOP-5854

This is a temporary update to the persistence module (since we are moving to netlify extension) for associated products.
Previously, we were only able to update one version at a time, so we updated all the sibling versions when calling the persistence module.

### README updates


seungpark:
concluded this is not the source of the problem. debugged as a group and found that netlify config for the persistence module (env var DB_NAME) was not set per env. this has been since updated

